PR #27, opened 2023-11-01
RTGen - Fixed enum value parsing in RTGenListener

Daniel-BB:
## Type of change

# Checklist:

# Description:

Previously, an enumeration value got "0x" removed when in Hex and then converted to Integer. The new string was presented to the language where it was hard to determine if it was Hex or Decimal.
In addition, a Decimal value (or Binary value converted to Decimal) would have been converted to Integer using the Hex-Converter resulting in a wrong Decimal value (eg. "12" -> 18).
The fix handles string-conversion correctly and will always return a Decimal value, since the Hex decorator may differ from language to language.


JakaMohorkoDS:
Shouldn't the RTGen binaries also be committed if the source was changed?

Daniel-BB:
> Shouldn't the RTGen binaries also be committed if the source was changed?

I totally forgot, added now.

shared/tools/RTGen/src/project/RTGen.Cpp/Parser/RTGenListener.cs
```diff
@@ -425,13 +428,16 @@ private bool TryGetEnumValue(RTGen3.ExpressionContext value, Enumeration enumDef
...
 
-            if (!int.TryParse(opt.Value, NumberStyles.HexNumber, CultureInfo.InvariantCulture, out int parsedCounter))
+            if (!int.TryParse(opt.Value, numberStyle, CultureInfo.InvariantCulture, out int parsedCounter))
             {
                 counterValue = 0;
                 return false;
             }
 
+            //store the integer value due to language differences for hex (0x1 or $1)
+            opt.Value = parsedCounter.ToString(CultureInfo.InvariantCulture);
```
Unsttopabull, line 438: This overrides the text for binary values (starting with `0b`)

Unsttopabull (changes requested):

Unsttopabull (approved):

PR #33, opened 2023-11-09
Websocket streaming alignment part 2

NikolaiShipilov:
## Type of change

# Checklist:

# Description:

openDAQ Websocket streaming client changes:
* received time/value signal meta info can be correctly handled independently from order of receiving
* `absoluteReference` streaming signal meta is converted into origin value of openDAQ signal - it fixes renderer time axis values
* `tableId` of time/value signals allowed to be not-equal to `signalId` of value signal
* name from descriptor is used as signal `"Name"` property instead of using signalId for signals of websocket pseudo device

### DIFF
--- a/shared/libraries/opcuatms/tests/opcuatms_integration/test_streaming_integration.cpp
+++ b/shared/libraries/opcuatms/tests/opcuatms_integration/test_streaming_integration.cpp
@@ -250,7 +250,12 @@ TEST_F(StreamingIntegrationTest, ByteStep)
     ASSERT_TRUE(packetsEqual(serverReceivedPackets, clientReceivedPackets));
 }
 
-TEST_F(StreamingIntegrationTest, ChangingSignal)
+// TODO websocket streaming does not recreate half assigned data descriptor changed event packet on client side
+// both: value and domain descriptors are always assigned in event packet
+// while on server side one descriptor can be assigned only
+// client side always generates 2 event packets for each server side event packet:
+// one for value descriptor changed and another for domain descriptor changed
+TEST_F(StreamingIntegrationTest, DISABLED_ChangingSignal)
 {
     const size_t packetsToGenerate = 5;
     const size_t initialEventPackets = 1;
--- a/shared/libraries/websocket_streaming/src/streaming_client.cpp
+++ b/shared/libraries/websocket_streaming/src/streaming_client.cpp
@@ -80,7 +80,7 @@ bool StreamingClient::connect()
         auto status = promiseFuturePair.second.wait_until(timeoutExpired);
         if (status != std::future_status::ready)
         {
-            LOG_W("singal {} has incomplete descriptors", id);
+            LOG_W("signal {} has incomplete descriptors", id);
         }
     }
 
@@ -102,9 +102,9 @@ void StreamingClient::onPacket(const OnPacketCallback& callack)
     onPacketCallback = callack;
 }
 
-void StreamingClient::onNewSignal(const OnSignalCallback& callback)
+void StreamingClient::onDataDescriptor(const OnSignalCallback& callback)
 {
-    onNewSignalCallback = callback;
+    onDataDescriptorCallback = callback;
 }
 
 void StreamingClient::onSignalUpdated(const OnSignalCallback& callback)
@@ -213,7 +213,17 @@ void StreamingClient::onProtocolMeta(daq::streaming_protocol::ProtocolHandler& p
 
                 std::promise<void> signalInitPromise;
                 std::future<void> signalInitFuture = signalInitPromise.get_future();
-                signalInitializedStatus.insert({signalId, std::make_pair(std::move(signalInitPromise), std::move(signalInitFuture))});
+                signalInitializedStatus.insert_or_assign(signalId, std::make_pair(std::move(signalInitPromise), std::move(signalInitFuture)));
+
+                if (auto signalIt = signals.find(signalId); signalIt == signals.end())
+                {
+                    auto inputSignal = std::make_shared<InputSignal>();
+                    signals.insert({signalId, inputSignal});
+                }
+                else
+                {
+                    LOG_E("Received duplicate of available signal. ID is {}.", signalId);
+                }
             }
 
             protocolHandler.subscribe(signalIds);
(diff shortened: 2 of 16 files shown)


shared/libraries/websocket_streaming/src/streaming_client.cpp
```diff
@@ -213,7 +213,17 @@ void StreamingClient::onProtocolMeta(daq::streaming_protocol::ProtocolHandler& p
...
+                signalInitializedStatus.insert_or_assign(signalId, std::make_pair(std::move(signalInitPromise), std::move(signalInitFuture)));
+
+                if (auto signalIt = signals.find(signalId); signalIt == signals.end())
+                {
+                    auto inputSignal = std::make_shared<InputSignal>();
+                    signals.insert({signalId, inputSignal});
+                }
+                else
+                {
+                    LOG_W("Signal id {} is already known as available", signalId);
```
JakaMohorkoDS, line 225: I'd try to log more descriptive messages in such cases. Something similar to "Received duplicate signal ID. Only one the first received signal will be streamed.".

shared/libraries/websocket_streaming/src/streaming_client.cpp
```diff
@@ -235,92 +245,149 @@ void StreamingClient::onMessage(const daq::streaming_protocol::SubscribedSignal&
...
-        {
-            iterator->second.first.set_value();
-        }
+        // value signal with tableId is known, set domain descriptor for it
+        setDomainDescriptor(signalIt->first, signalIt->second, sInfo.dataDescriptor);
+    }
+    else
+    {
+        // value signal with tableId is unknown, save domain descriptor
+        cachedDomainDescriptors.insert_or_assign(tableId, sInfo.dataDescriptor);
```
JakaMohorkoDS, line 301: Does this assume we only ever have 2 signals with the same table ID, and the signal is actually our data signal? Though I think we'll have issues with such cases in general, so this should be a good enough solution for now.

It might be better to look for a signal that has a time relation to the domain signal we received here.

shared/libraries/websocket_streaming/src/streaming_client.cpp
```diff
@@ -235,92 +245,149 @@ void StreamingClient::onMessage(const daq::streaming_protocol::SubscribedSignal&
...
     try
     {
+        {
+            LOG_I("Signal #{}; signalId {}; tableId {}; name {}; value type {}; Json parameters: \n\n{}\n",
+                  subscribedSignal.signalNumber(),
+                  subscribedSignal.signalId(),
+                  subscribedSignal.tableId(),
+                  subscribedSignal.memberName(),
+                  subscribedSignal.dataValueType(),
+                  params.dump());
```
JakaMohorkoDS, line 343: Should these be info or debug level?

shared/libraries/websocket_streaming/src/websocket_client_device_impl.cpp
```diff
@@ -77,13 +77,18 @@ void WebsocketClientDeviceImpl::createWebsocketStreaming()
...
     if (!sInfo.dataDescriptor.assigned())
         return;
 
     if (auto signalIt = deviceSignals.find(signalId); signalIt != deviceSignals.end())
     {
+        // sets signal name as it appeared in descriptor
+        auto protectedObject = signalIt->second.asPtr<IPropertyObjectProtected>();
+        if (sInfo.dataDescriptor.getName().assigned())
+            protectedObject.setProtectedPropertyValue("Name", sInfo.dataDescriptor.getName());
+
```
JakaMohorkoDS, line 91: `dataDescriptor.getName()` does not receive the name of the signal. The name of the signal is received within openDAQ as `signal.getName()`. At least the initial name should probably be transferred as metadata - and it's different from the signal ID.

shared/libraries/opcuatms/tests/opcuatms_integration/test_streaming_integration.cpp
```diff
@@ -250,7 +250,12 @@ TEST_F(StreamingIntegrationTest, ByteStep)
     ASSERT_TRUE(packetsEqual(serverReceivedPackets, clientReceivedPackets));
 }
 
-TEST_F(StreamingIntegrationTest, ChangingSignal)
+// TODO websocket streaming does not recreate half assigned data descriptor changed event packet on client side
+// both: value and domain descriptors are always assigned in event packet
+// while on server side one descriptor can be assigned only
+// client side always generates 2 event packets for each server side event packet:
+// one for value descriptor changed and another for domain descriptor changed
```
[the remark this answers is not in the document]
NikolaiShipilov: Sure, I'll create a ticket once these changes have been merged.

shared/libraries/websocket_streaming/src/streaming_client.cpp
NikolaiShipilov: It assumes that at least all data signals are published as `available`. There might be more than 2 signals sharing the same `tableId`: the data signal itself, related domain signal, related status signal etc. Signals other than the domain or the data signal itself will simply be ignored at the moment (in case of not being published as `available` ). 
An alternative to utilizing the `tableId` is to search for a signal with the type 'domain' within the `relatedSignals` list to identify which signal serves as the domain for the data signal. However, I don't particularly see any advantages to this approach.

shared/libraries/websocket_streaming/src/streaming_client.cpp
NikolaiShipilov: I'd prefer to keep this at the 'info' level to have this logs in release versions

shared/libraries/websocket_streaming/src/websocket_client_device_impl.cpp
NikolaiShipilov: I aimed to address Nils' request "_It would also nice, if for a pure caps=WS device, we see also here the signal name not the id_" and make gui_demo app display signal name instead of weird signal id containing `#` symbol. gui demo displays the string obtained from `signal.getName()` which is the value of `"Name"` property.  In turn, this property is initialized with `localId` value.

JakaMohorkoDS: Sure, but I don't think the name on the data descriptor is correct here. Is only an ID available on the streaming client, and no signal "display name" ?

NikolaiShipilov: signal name is available in `definition` json object on streaming client. For example it is `"name":"Connector1.Channel1.Signal2"` for fusion device and `"name":"DBX_CH1.value"` for mbedOS device. Signal generator tool keeps it default `"name":"value"`. openDAQ streaming server sets this field with name value from descriptor.

NikolaiShipilov: Changed by [commit](https://github.com/openDAQ/openDAQ/pull/33/commits/685c9884dc1d3f799f6d784edcc4fedbd402859a). Now signal name is used as `"name"` field in streaming `"definition"`meta-data. The descriptor name is encoded into `"desc_name"` field of `"interpretation"` object for openDAQ->openDAQ communication.

JakaMohorkoDS (approved):

PR #35, opened 2023-11-13
[TBBAS-597] support vector by renderer

denise-opendaq:
## Type of change

# Checklist:

# Description:


modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,333 @@
...
+        lastPacketSamples += wasBreaked;
+
+        *outputDomainData = inputDomainData[lastPacketSamples - 1];
+        packetStarted += 1000 * epochMs;
+    }
+
+    Int offset = static_cast<Int>(outputDataDescriptor.getDimensions()[0].getRule().getParameters().get("start"));
+    double valCnt = 0;
+
+    for (const auto & list_packet: packets) {
```
dejancrn, line 282: One input packet can contain huge number of samples, i.e. 1 million. That means that from one input packet multiple vector samples should be produced. I don't think this case is covered here.

Actually this would be a perfect use case for "block reader", but unfortunately we don't support readers in function blocks.

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,333 @@
...
+    : FunctionBlock(CreateType(), ctx, parent, localId)
+{
+    createInputPorts();
+    createSignals();
+    initProperties();
+}
+
+void ClassifierFbImpl::initProperties()
+{
+    const auto epochMsProp = IntProperty("EpochMs", 1);
```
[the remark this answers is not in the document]
denise-opendaq: BlockSize sounds like we are going to split data in a blocks of BlockSize elements. Plus its not representing in which time this values is configurated

dejancrn (approved):
As discussed, an option to define explicit axis in classification and use getLabels in renderer to get axis labels

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,360 @@
...
+        auto inputDomainData = static_cast<UInt*>(packet.getDomainPacket().getData());
+        UInt lastTime = inputDomainData[packet.getSampleCount() - 1];
+
+        // initialize members
+        if (packetStarted == UInt{}) {
+            packetStarted = inputDomainData[0];
+            sampleStarted = 0; 
+        }
+
+        outputPackages = (lastTime - packetStarted) / timeMs(blockSize);
```
JakaMohorkoDS, line 253: The time is usually in ticks, not yet scaled to the unit of the signal. This works only if the unit of the signal is "seconds", and the resolution is 1/1000.

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,360 @@
...
+
+    auto outputDomainPacket = DataPacket(outputDomainDataDescriptor, 1);
+    auto outputDomainData = static_cast<UInt*>(outputDomainPacket.getData());
+
+    DataPacketPtr outputPacket = DataPacketWithDomain(outputDomainPacket, outputDataDescriptor, 1);
+    OutputType* outputData = static_cast<OutputType*>(outputPacket.getData());
+
+    auto rangeSize = outputDataDescriptor.getDimensions()[0].getSize();
+    auto labels = outputDataDescriptor.getDimensions()[0].getLabels();
+    if (labels.getCount() == 0) return;
```
JakaMohorkoDS, line 267: New line before return.

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,360 @@
...
+        auto sampleCnt = listPacket.getSampleCount();
+        auto inputData = static_cast<InputType*>(listPacket.getData());
+        auto inputDomainData = static_cast<UInt*>(listPacket.getDomainPacket().getData());
+       
+        // reset array for new package
+        if (packageVals == 0) {
+            memset(outputData, 0, rangeSize * sizeof(OutputType));
+        }
+
+        bool packetInEpoch = timeInInterval(packetStarted, inputDomainData[sampleCnt - 1]);
```
JakaMohorkoDS, line 282: Same as above, we're assuming that the delta of "1" in the input signal domain data corresponds to 1ms.

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,360 @@
...
+            packets.pop_front();
+        } else {
+            if (packageVals) {
+                for (size_t i = 0; i < rangeSize; i++) {
+                    outputData[i] /= packageVals;
+                }
+            }
+
+            outputSignal.sendPacket(outputPacket);
+            outputDomainSignal.sendPacket(outputDomainPacket);
```
JakaMohorkoDS, line 336: If `outputPackages > 1` this will create issues, as you're sending the same packet object more than once, and re-writing the data buffer of the same packet. A new packet with its own buffer needs to be created for every `sendPacket` call.

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,360 @@
...
+
+        bool packetInEpoch = timeInInterval(packetStarted, inputDomainData[sampleCnt - 1]);
+
+        size_t sampleIdx = sampleStarted;
+        for (; sampleIdx < sampleCnt; sampleIdx++) {
+            // if there are values from another interval - check each value
+            if (!packetInEpoch) {
+                if (!timeInInterval(packetStarted, inputDomainData[sampleIdx]))
+                    break;
+            }
```
JakaMohorkoDS, line 290: Not sure we need to consider this here, but this loop will fail if the domain delta between two samples is greater than our block size (in ms).

modules/ref_fb_module/src/renderer_fb_impl.cpp
```diff
@@ -875,47 +996,56 @@ void RendererFbImpl::renderAxis(sf::RenderTarget& renderTarget, SignalContext& s
...
+
+                domainText.setString(domainStr.str());
+                const auto domainBounds = domainText.getGlobalBounds();
+                float xOffset;
+                if (i == 0)
+                xOffset = 0;
+                else if (i == xTickCount - 1)
+                xOffset = domainBounds.width;
+                else
+                xOffset = domainBounds.width / 2.0f;
```
JakaMohorkoDS, line 1043: Indentation

modules/ref_fb_module/src/renderer_fb_impl.cpp
```diff
@@ -875,47 +996,56 @@ void RendererFbImpl::renderAxis(sf::RenderTarget& renderTarget, SignalContext& s
...
-
-        if (!drawXAxisLabels)
-            continue;
-
-        // for absolute time show only 3 domain values, otherwise 5
-        if (i % 2 == 0 || !signalContext.hasTimeOrigin)
+    }
+    
+    // create labeles for horizontal axi
+    if (drawXAxisLabels) {
```
JakaMohorkoDS, line 1009: Seemed cleaner with the
``` 
if (!drawXAxisLabels)
            continue;
```

check in between.

modules/ref_fb_module/src/renderer_fb_impl.cpp
```diff
@@ -823,33 +924,55 @@ void RendererFbImpl::getYMinMax(const SignalContext& signalContext, double& yMax
 
 void RendererFbImpl::renderAxis(sf::RenderTarget& renderTarget, SignalContext& signalContext, const sf::Font& font, bool drawXAxisLabels, bool drawTitle)
```
JakaMohorkoDS, line 925: Was the axis rendering checked with both array and scalar signals on the renderer at the same time? Do all the properties work (eg. SingleXAxis, SingleYAxis) in that case?

modules/ref_fb_module/tests/test_ref_fb_module.cpp
```diff
@@ -150,3 +153,11 @@ TEST_F(RefFbModuleTest, StatisticsNumOfSignals)
     auto fb = module.createFunctionBlock("ref_fb_module_statistics", nullptr, "id");
     ASSERT_EQ(fb.getSignalsRecursive().getCount(), 3u);
 }
+
+TEST_F(RefFbModuleTest, CreateFunctionBlockClassifier)
```
JakaMohorkoDS, line 157: Missing integration tests that generate standard inputs and validate that the classification outputs are correct for both synchronous and asynchronous signals. Test cases with invalid input signals should also be added.

I'd try to think about when the function block will fail, and add said cases to the test suite as well - marked as "DISABLED" until the functionality to handle them is added to the classifier.

JakaMohorkoDS (changes requested):
Logic wise the function block implementation is okay, and the renderer lgtm. Stylistically, curly brackets, as well as `else` statements should be put into a new line.

modules/ref_fb_module/src/classifier_fb_impl.cpp
denise-opendaq: we will just get out from loop and will send empty package

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,374 @@
...
+#include <opendaq/input_port_factory.h>
+#include <opendaq/data_descriptor_ptr.h>
+
+#include <opendaq/event_packet_ptr.h>
+#include <opendaq/signal_factory.h>
+
+#include <opendaq/custom_log.h>
+#include <opendaq/event_packet_params.h>
+
+#include <coreobjects/unit_factory.h>
```
JakaMohorkoDS, line 12: Empty lines between includes.

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,374 @@
...
+    objPtr.addProperty(BoolProperty("UseExplicitDomain", false));
+    objPtr.getOnPropertyValueWrite("UseExplicitDomain") +=
+        [this](PropertyObjectPtr& obj, PropertyValueEventArgsPtr& args) { propertyChanged(true); };
+
+    auto explcitDimensionProp = ListPropertyBuilder("ExplcitDimension", List<Float>()).setVisible(EvalValue("$UseExplicitDomain")).build();
+    objPtr.addProperty(explcitDimensionProp);
+    objPtr.getOnPropertyValueWrite("ExplcitDimension") +=
+        [this](PropertyObjectPtr& obj, PropertyValueEventArgsPtr& args) { propertyChanged(true); };
+
+    auto blockSizeMsProp = IntPropertyBuilder("BlockSizeMs", 10).setVisible(EvalValue("!$UseExplicitDomain")).build();
```
JakaMohorkoDS, line 46: This could have the unit configured via `IPropertyBuilder::setUnit()`

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,374 @@
...
+    : FunctionBlock(CreateType(), ctx, parent, localId)
+{
+    createInputPorts();
+    createSignals();
+    initProperties();
+}
+
+void ClassifierFbImpl::initProperties()
+{
+    objPtr.addProperty(BoolProperty("UseExplicitDomain", false));
```
JakaMohorkoDS, line 37: ExplicitDimension, not domain.

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,374 @@
...
+    objPtr.getOnPropertyValueWrite("UseExplicitDomain") +=
+        [this](PropertyObjectPtr& obj, PropertyValueEventArgsPtr& args) { propertyChanged(true); };
+
+    auto explcitDimensionProp = ListPropertyBuilder("ExplcitDimension", List<Float>()).setVisible(EvalValue("$UseExplicitDomain")).build();
```
JakaMohorkoDS, line 41: Typo in "Explcit"

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,374 @@
...
+{
+    useExplicitDomain = objPtr.getPropertyValue("UseExplicitDomain");
+    explicitDimension =  objPtr.getPropertyValue("ExplcitDimension");
+    blockSizeMs = objPtr.getPropertyValue("BlockSizeMs");
+    classCount = objPtr.getPropertyValue("ClassCount");
+    outputName = static_cast<std::string>(objPtr.getPropertyValue("OutputName"));
+
+    assert(blockSizeMs > 0);
+    assert(classCount > 0);
+
```
JakaMohorkoDS, line 82: classCount does not need to be >0 if explicit dimensions are used.

modules/ref_fb_module/src/classifier_fb_impl.cpp
JakaMohorkoDS: ExplicitDimension does maybe not exactly convey what is meant by this. A description should be added to clarify via `IPropertyBuilder::setDescription()`

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,360 @@
...
+        this->inputDomainDataDescriptor = inputDomainDataDescriptor;
+
+    configure();
+}
+
+void ClassifierFbImpl::configure()
+{
+    if (!inputDataDescriptor.assigned() || !inputDomainDataDescriptor.assigned())
+    {
+        LOG_D("Incomplete signal descriptors")
```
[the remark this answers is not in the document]
JakaMohorkoDS: Logging message was not updated.

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,374 @@
...
+            inputSampleType != SampleType::Int16 &&
+            inputSampleType != SampleType::Int32 &&
+            inputSampleType != SampleType::Int64 &&
+            inputSampleType != SampleType::UInt8 &&
+            inputSampleType != SampleType::UInt16 &&
+            inputSampleType != SampleType::UInt32 &&
+            inputSampleType != SampleType::UInt64)
+            throw std::runtime_error("Invalid sample type");
+
+        if (inputDomainDataDescriptor.getSampleType() != SampleType::Int64 && inputDomainDataDescriptor.getSampleType() != SampleType::UInt64)
```
JakaMohorkoDS, line 127: Domain unit should be checked for whether the unit is "s"/"seconds".

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,374 @@
...
+        {
+            throw std::runtime_error("Incompatible domain data sample type");
+        }
+
+        const auto domainRule = inputDomainDataDescriptor.getRule();
+        if (domainRule.getType() != DataRuleType::Linear)
+        {
+            throw std::runtime_error("Domain rule type is not Linear");
+        }
```
JakaMohorkoDS, line 136: The function block is then limited to accept only synchronous signals (signals with equidistant samples)? If so, the data processing could be modified a bit, since this means that the block size will always be static, and can be calculated from the linear rule.

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,374 @@
...
+    {
+        DataDescriptorPtr inputDataDescriptor = packet.getParameters().get(event_packet_param::DATA_DESCRIPTOR);
+        DataDescriptorPtr inputDomainDataDescriptor = packet.getParameters().get(event_packet_param::DOMAIN_DATA_DESCRIPTOR);
+        processSignalDescriptorChanged(inputDataDescriptor, inputDomainDataDescriptor);
+    }
+}
+
+inline UInt ClassifierFbImpl::timeMs(UInt time) 
+{
+    return time * inputDeltaTicks;
```
JakaMohorkoDS, line 224: Assuming that the input is in seconds, to convert to ms, the "ticks" (raw data of the domain signal) should be multiplied as follows:
```
time * inputDomainDescriptor.getTickResolution() * (1 / 1000)
```

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,374 @@
...
+    using InputType = typename SampleTypeToType<InputSampleType>::Type;
+    using OutputType = Float;
+
+    // if packet does not have samples - ignore
+    if (packet.getSampleCount() == 0)
+        return;
+
+    packets.push_back(packet);
+
+    size_t outputPackages = 0;
```
JakaMohorkoDS, line 244: outputPackets seems more suitable of a name.

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,374 @@
...
+        }
+
+        outputPackages = (lastTime - packetStarted) / timeMs(blockSizeMs);
+    }
+
+    if (!outputPackages)
+        return;
+
+    auto outputDomainPacket = DataPacket(outputDomainDataDescriptor, 1);
+    auto outputDomainData = static_cast<UInt*>(outputDomainPacket.getData());
```
JakaMohorkoDS, line 263: Similar issue as before, domain packet data will get overridden when more than 1 packet is sent in the loop below, as the same packet will be sent multiple times.

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,374 @@
...
+
+    DataPacketPtr outputPacket {};
+    OutputType* outputData {};
+
+    auto rangeSize = outputDataDescriptor.getDimensions()[0].getSize();
+    auto labels = outputDataDescriptor.getDimensions()[0].getLabels();
+    if (labels.getCount() == 0) 
+        return;
+
+    size_t packageVals = 0;
```
JakaMohorkoDS, line 273: packetValueCount

modules/ref_fb_module/include/ref_fb_module/classifier_fb_impl.h
```diff
@@ -0,0 +1,92 @@
...
+    size_t classCount;
+    Float inputDeltaTicks;
+
+    std::string outputName;
+
+    bool useExplicitDomain;
+    ListPtr<Float> explicitDimension;
+
+    UInt packetStarted {};
+    size_t sampleStarted {};
```
JakaMohorkoDS, line 65: Names don't really convey what this is intended for. This should convey that this is the index where the FB ended reading in the last block.

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,374 @@
...
+        auto sampleCnt = listPacket.getSampleCount();
+        auto inputData = static_cast<InputType*>(listPacket.getData());
+        auto inputDomainData = static_cast<UInt*>(listPacket.getDomainPacket().getData());
+       
+        // reset array for new package
+        if (packageVals == 0)
+        {
+            outputPacket = DataPacketWithDomain(outputDomainPacket, outputDataDescriptor, 1);
+            outputData = static_cast<OutputType*>(outputPacket.getData());
+            memset(outputData, 0, rangeSize * sizeof(OutputType));
```
JakaMohorkoDS, line 287: This is invalid if we are using explicit dimensions, as rangeSize should be equal to the dimension size.

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,374 @@
...
+
+        auto outputDataDescriptorBuilder = DataDescriptorBuilder().setSampleType(SampleType::Float64);
+        
+        auto dimensions = List<IDimension>();
+        if (useExplicitDomain) 
+            dimensions.pushBack(Dimension(ListDimensionRule(explicitDimension)));
+        else 
+        {
+            auto inputLowValue = static_cast<Float>(inputDataDescriptor.getValueRange().getLowValue());
+            auto inputHighValue = static_cast<Float>(inputDataDescriptor.getValueRange().getHighValue());
```
JakaMohorkoDS, line 150: Not all descriptors might have this value configured. In that case, the signal should be dropped, or a default low/high should be present.

JakaMohorkoDS (changes requested):
Test cases with mock signals are still missing for the classification function block.

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,529 @@
...
+{
+    std::scoped_lock lock(sync);
+    readProperties();
+    if (configure)
+        this->configure();
+}
+
+void ClassifierFbImpl::readProperties()
+{
+    useExplicitDomain = objPtr.getPropertyValue("UseExplicitDimension");
```
JakaMohorkoDS, line 88: Typo, variable name should be "useExplicitDimension"

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,529 @@
...
+    {
+        switch (packet.getType())
+        {
+            case PacketType::Event:
+                processEventPacket(packet);
+                break;
+
+            case PacketType::Data:
+                SAMPLE_TYPE_DISPATCH(inputSampleType, processDataPacket, packet);
+                break;
```
JakaMohorkoDS, line 248: We could already call different functions here, depending on the sample type.

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,529 @@
...
+        UInt lastTime = inputDomainData[packet.getSampleCount() - 1];
+
+        // initialize members
+        if (packetStarted == UInt{})
+        {
+            packetStarted = inputDomainData[0];
+            lastReadSampleInBlock = 0; 
+        }
+
+        outputPackets = (lastTime - packetStarted) / timeMs(blockSize);
```
JakaMohorkoDS, line 431: Should first scale ticks to "s" using resolution.

modules/ref_fb_module/src/classifier_fb_impl.cpp
```diff
@@ -0,0 +1,529 @@
...
+    objPtr.addProperty(BoolPropertyBuilder("UseExplicitDimension", false)
+        .setDescription("Choose classification based on input signal min max with step of ClassCount or custom marks").build());
+    objPtr.getOnPropertyValueWrite("UseExplicitDimension") +=
+        [this](PropertyObjectPtr& obj, PropertyValueEventArgsPtr& args) { propertyChanged(true); };
+
+    auto explcitDimensionProp = ListPropertyBuilder("ExplicitDimension", List<Float>()).setVisible(EvalValue("$UseExplicitDimension"))
+        .setDescription("Set custom list for classification rule").build();
+    objPtr.addProperty(explcitDimensionProp);
+    objPtr.getOnPropertyValueWrite("ExplicitDimension") +=
+        [this](PropertyObjectPtr& obj, PropertyValueEventArgsPtr& args) { propertyChanged(true); };
```
JakaMohorkoDS, line 43: The final touch here would be to rename the properties to "UseCustomClasses", and "CustomClassList"

JakaMohorkoDS (commented):
Lgtm, only minor issues left. Tick scaling with resolution should be addressed when doing timestamp arithmetic, and additional tests for the classification function block features must be written before merging.

modules/tests/test_ref_modules/test_ref_modules.cpp
```diff
@@ -575,5 +575,493 @@ TEST_F(RefModulesTest, UpdateDevicePower)
...
+    const auto classifierDomainSignal = classifierSignal.getDomainSignal();
+    const auto classifierDomainSignalDescription = classifierDomainSignal.getDescriptor();
+
+    const auto deviceDomainDescriptor = deviceSignal.getDomainSignal().getDescriptor();
+
+    const auto deviceDomainRule = deviceDomainDescriptor.getRule();
+    const auto classifierDomainRule = classifierDomainSignalDescription.getRule();
+
+    // Check Linear Rule
+    if (deviceDomainRule.getType() == DataRuleType::Linear)
```
JakaMohorkoDS, line 675: Is this `if` statement needed? (similar question for the checks in the other tests)

modules/tests/test_ref_modules/test_ref_modules.cpp
```diff
@@ -575,5 +575,493 @@ TEST_F(RefModulesTest, UpdateDevicePower)
...
+    classifierFb.setPropertyValue("UseCustomClasses", false);
+    classifierFb.setPropertyValue("CustomClassList", customClassList);
+    classifierFb.setPropertyValue("BlockSize", blockSize);
+    classifierFb.setPropertyValue("ClassCount", classCount);
+
+    const auto deviceChannel = device.getChannels()[0];
+    deviceChannel.setPropertyValue("CustomRange", inputRange);
+    const auto deviceSignal = deviceChannel.getSignals()[0];
+
+    const auto classifierInputPort = classifierFb.getInputPorts()[0];
```
JakaMohorkoDS, line 635: A lot of these checks seem to be exact duplicates from the previous test.

modules/tests/test_ref_modules/test_ref_modules.cpp
```diff
@@ -575,5 +575,493 @@ TEST_F(RefModulesTest, UpdateDevicePower)
...
+    classifierInputPort.connect(deviceSignal);
+
+    const auto classifierSignal = classifierFb.getSignals()[0];
+    const auto classifierSignalDescription = classifierSignal.getDescriptor();
+
+    // Classifier returning values with float float
+    ASSERT_EQ(classifierSignalDescription.getSampleType(), SampleType::Float64);
```
JakaMohorkoDS, line 726: Same as above, these tests are repeated and somewhat obfuscate the purpose of the test.

The goal of each test should be clear in what it's testing. If the test is titled "CheckBuiltInClassCustomRange" it should test for that, not for all other unrelated properties.

In general, these tests should aim to be as short as possible. Most of the tests available here should be split into 2-4 smaller, more targeted test cases.

modules/tests/test_ref_modules/test_ref_modules.cpp
```diff
@@ -575,5 +575,493 @@ TEST_F(RefModulesTest, UpdateDevicePower)
...
+    const auto classifierSignalDescription = classifierSignal.getDescriptor();
+    const auto classifierSignalLabelsSize = classifierSignal.getDescriptor().getDimensions()[0].getSize();
+
+    auto reader = BlockReader<outputSignalType>(classifierSignal, outputPacketCount);
+
+    auto dataPacket = input.createDataPacket(inputPacketCount);
+    auto dataPtr = static_cast<inputSignalType*>(dataPacket.getData());
+    for (size_t i = 0; i < inputPacketCount; i++)
+    {
+        auto deliter = inputPacketCount > 1 ? inputPacketCount - 1 : 1;
```
JakaMohorkoDS, line 814: "deliter" ?

modules/tests/test_ref_modules/test_ref_modules.cpp
```diff
@@ -575,5 +575,493 @@ TEST_F(RefModulesTest, UpdateDevicePower)
...
+
+    InputStubSignal input;
+    input.setUp(inputSignalSampleType, inputSignalRange, inputSignalSync);
+    auto instance = input.getInstance();
+    auto inputSignal = input.getSignal();
+
+    SizeT blockSizeMs = 5;
+    SizeT classCount = 1;
+    SizeT inputPacketCount = blockSizeMs;
+    SizeT outputPacketCount = inputPacketCount / blockSizeMs;
```
JakaMohorkoDS, line 794: This is basically always dividing "1 / 1".

modules/tests/test_ref_modules/test_ref_modules.cpp
```diff
@@ -575,5 +575,493 @@ TEST_F(RefModulesTest, UpdateDevicePower)
...
+    classifierFb.setPropertyValue("UseCustomClasses", false);
+    classifierFb.setPropertyValue("BlockSize", blockSizeMs);
+    classifierFb.setPropertyValue("ClassCount", classCount);
+
+    const auto classifierPort = classifierFb.getInputPorts()[0];
+    classifierPort.connect(inputSignal);
+    const auto classifierSignal = classifierFb.getSignals()[0];
+
+    const auto classifierSignalDescription = classifierSignal.getDescriptor();
+    const auto classifierSignalLabelsSize = classifierSignal.getDescriptor().getDimensions()[0].getSize();
```
JakaMohorkoDS, line 806: There's a lot of duplicate variables here that use different names to describe the same thing. In this case

```
classifierSignalLabelsSize == classCount
```

This makes the test cases hard to read and to grasp what each variable represents.

modules/tests/test_ref_modules/test_ref_modules.cpp
JakaMohorkoDS, line 814: The test cases are statically defined - they shouldn't need to have cases for checking what the inputs are, and what, for example, the `inputPacketCount` is. These are known at the start of the test.

modules/tests/test_ref_modules/test_ref_modules.cpp
```diff
@@ -575,5 +575,493 @@ TEST_F(RefModulesTest, UpdateDevicePower)
...
+    {
+        auto deliter = inputPacketCount > 1 ? inputPacketCount - 1 : 1;
+        auto delta = (static_cast<Float>(inputSignalRange.getHighValue()) - static_cast<Float>(inputSignalRange.getLowValue())) / deliter;
+        dataPtr[i] = inputSignalRange.getLowValue() + delta * i;
+    }
+    input.sendPacket(dataPacket);
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+
+    SizeT blockCnt = outputPacketCount;
```
JakaMohorkoDS, line 822: Similar as above, we're renaming a variable for no real effect.

modules/tests/test_ref_modules/input_stub_signal_helper.h
```diff
@@ -0,0 +1,121 @@
...
+#include <opendaq/reader_utils.h>
+#include <opendaq/sample_type_traits.h>
+#include <opendaq/scheduler_factory.h>
+#include <opendaq/signal_factory.h>
+#include <testutils/testutils.h>
+#include <random>
+
+using namespace daq;
+
+struct InputStubSignal
```
JakaMohorkoDS, line 32: This isn't really a signal, but an entire test app setup. Considering that, I'd also add the "ref_fb_classifier" function block in the setup method, and rename the struct appropriately to "ClassifierTestHelper" or something similar.

modules/tests/test_ref_modules/test_ref_modules.cpp
```diff
@@ -546,7 +546,7 @@ TEST_F(RefModulesTest, UpdateDevicePower)
     device1Channel.release();
     device0Signal.release();
     device1Signal.release();
-//    rendererFb.release();
+    //    rendererFb.release();
```
JakaMohorkoDS, line 511: These indentation changes seem somewhat odd. If changed in this manner, there should not be a whole tab following the "//", but a single space.

JakaMohorkoDS (changes requested):
A few general comments:
- A large portion of each test is the setup, which is duplicated between the tests. It makes the tests really hard to read. Such setup should be extracted to a separate method.
- Tests have baseline checks that are repeated, but don't actually test what the title of the unit test implies -> they actually hide the true purpose of the test.
- Tests with asynchronous inputs are missing (iirc, the classifier supports those now)

JakaMohorkoDS (approved):

PR #38, opened 2023-11-15
Add begin and end update system to property object

dejancrn:


core/coreobjects/include/coreobjects/property_object.h
```diff
@@ -288,6 +288,12 @@ DECLARE_OPENDAQ_INTERFACE(IPropertyObject, IBaseObject)
      * kept in insertion order at the end of the Property object's list of properties.
      */
     virtual ErrCode INTERFACE_FUNC setPropertyOrder(IList* orderedPropertyNames) = 0;
+
+    virtual ErrCode INTERFACE_FUNC beginUpdate() = 0;
+    virtual ErrCode INTERFACE_FUNC endUpdate() = 0;
```
JakaMohorkoDS, line 293: Missing doxygen comments.

core/opendaq/opendaq/src/instance_impl.cpp
```diff
@@ -659,6 +659,21 @@ ErrCode InstanceImpl::getOnPropertyValueRead(IString* propertyName, IEvent** eve
...
+
+ErrCode InstanceImpl::endUpdate()
+{
+    return rootDevice->endUpdate();
+}
+
+ErrCode InstanceImpl::getOnEndUpdate(IEvent** event)
+{
+    return rootDevice->endUpdate();
+}
```
JakaMohorkoDS, line 675: I feel like this might be misunderstood to mean that the entire instance and its subtree will enter the "isUpdating" state, while this only applies to a single property object.

On that note - would it make sense for nested property objects to also be put into the same state, and its changes applied when the owner stops updating?

JakaMohorkoDS (commented):
OPC UA has no changes here. Would it make sense to at least override the begin/end update calls to return NOT_IMPLEMENTED? 

If not, should I already add the OPC UA logic to this branch or make a new PR?

core/coreobjects/include/coreobjects/property_value_event_args.h
```diff
@@ -37,13 +37,16 @@ DECLARE_OPENDAQ_INTERFACE(IPropertyValueEventArgs, IEventArgs)
...
 };
 
 OPENDAQ_DECLARE_CLASS_FACTORY(
     LIBRARY_FACTORY, PropertyValueEventArgs,
     IProperty*, prop,
     IBaseObject*, value,
-    PropertyEventType, type
+    PropertyEventType, type,
+    Bool,
+    isBatched
```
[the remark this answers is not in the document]
dejancrn: done

core/opendaq/opendaq/src/instance_impl.cpp
dejancrn: I would do this at next iteration.

JakaMohorkoDS (approved):

PR #49, opened 2023-11-27
[TBBAS-1148][TBBAS-1151] Typed TailReader and BlockReader, Time wrappers for readers

Yarikk26:
## Type of change

# Checklist:

# Description:

- TailReader and BlockReader have typed read methods now.
- TimeStreamReader, TimeTailReader, TimeBlockReader wrappers are added.
- Tests and docs are updated.


bindings/python/tests/test_reader_datetime.py
```diff
@@ -74,13 +66,149 @@ def test_read_with_timestamps(self):
...
 
         time.sleep(0.1)
         (values, domain) = reader.read_with_timestamps(100)
         for t in domain:
             self.assertIsInstance(t, numpy.datetime64)
 
+    def test_tail_read(self):
+        import time
+
+        instance = opendaq.Instance()
```
dejancrn, line 79: Calling opendaq.Instance can be expensive, because it loads modules. I would suggest the following approach:

Expose "CreateMockSignal" function to python from the bindings. This returns a signal wrapper which has the following functions:

- getSignal,
- getDomainSignal,
- addData

When calling init of a wrapper this:

- creates a value signal and a domain signal
- sets domainSignal property of value signal to domainSignal
- configures both signals with signal descriptor (data type can be parameter to init)

AddData method:

- takes py:array parameter
- constructs value and domain packet
- copies data into packet
- calls ISignal.addPacket on both value and domain packets.

You can then pass mockSignal.getSignal() to reader factories. Mock signal should be used in tests for readers, function blocks etc...

dejancrn (changes requested):

JakaMohorkoDS (changes requested):
Lgtm, with the exception of documentation and Python test updates.

The quick start guide also uses readers in the examples, and I think the syntax in those is not 100% correct anymore. Also, some tests for the guide probably have invalid syntax now as well.

JakaMohorkoDS (approved):

bindings/python/tests/test_reader_datetime.py
```diff
@@ -5,204 +5,173 @@
...
-        signal = device.signals_recursive[0]
-        reader = opendaq.StreamReader(signal)
+        mock.add_data(numpy.arange(100))
+        self.assertEqual(reader.available_count, 100)
 
-        time.sleep(0.1)
-        values = reader.read(100)
+        values = reader.read(10)
         for v in values:
             self.assertIsInstance(v, numpy.float64)
```
dejancrn, line 21: Can you do something like `self.assert(values, numpy.arange(100))`

dejancrn (approved):

JakaMohorkoDS (approved):

dejancrn (approved):

PR #53, opened 2023-11-29
Add SearchParams for tree traversal

JakaMohorkoDS:
## Type of change

# Checklist:

# Description:


dejancrn (approved):
Looks nice, however, being a "generic" lover I have to say I would design the search parameters a little different. For example, how would you search for invisible components only?

ISearchParams=interface
{
    ErrCode acceptComponent(IComponent* component)
}

Then I would define several classes that implement the ISearchParams:

- LocalIdSearchParams(IString* id)
- VisibleSearchParams()
- Tags(IList* tags)

And also define composition ISearchParams:

- And(ISearchParams p1, p2) -> ISearchParams
- Or(ISearchParams p1, p2) -> ISearchParams
- Not(ISearchParams p) -> ISearchParams

To search for invisible signals, you would call:
getSignals(Not(Visible()))

You could still define a composite search params object like it is defined now.

The question remains how to pass recursive parameter? Perhaps with second default argument or have a wrapper around ISearchParams that returns some IRecursiveSearchParams.

JakaMohorkoDS:
As the PR is again huge, I'd split up at least part of it amongst reviewers:
- @NikolaiShipilov Streaming changes
- @janm-dw OPC UA changes
- @dejancrn Core changes

Of course everyone is welcome and encouraged to look at the whole PR :)

JakaMohorkoDS:
The PR is currently missing documentation and changelog updates.

shared/libraries/opcuatms/opcuatms_server/src/tms_server_context.cpp
```diff
@@ -0,0 +1,28 @@
...
+}
+
+TmsServerContext::~TmsServerContext()
+{
+    this->context.getOnCoreEvent() -= event(this, &TmsServerContext::coreEventCallback);
+}
+
+void TmsServerContext::registerComponent(const ComponentPtr& component, tms::TmsServerObject& obj)
+{
+    idToObjMap.insert(std::make_pair(component.getGlobalId(), obj.weak_from_this()));
```
janm-dw, line 18: What happens if server component is deleted? I think we should remove it from the map.

janm-dw (approved):

shared/libraries/opcuatms/opcuatms_server/src/tms_server_context.cpp
JakaMohorkoDS: True, but I think we have issues in OPC UA in general when anything changes on the server structure-wise. Don't think we can currently react to any addition/removal. Maintaining this list isn't really an issue, but I don't see a gain in it at the moment, considering the OPC UA doesn't get deleted in response anyhow.

shared/libraries/websocket_streaming/tests/test_websocket_client_device.cpp
```diff
@@ -106,14 +107,15 @@ TEST_F(WebsocketClientDeviceTest, SingleSignalWithDomain)
...
+    testSignal.asPtr<IComponentPrivate>().unlockAllAttributes();
+    testSignal.setName("NewName");
+    testSignal.setDescription("NewDescription");
     std::this_thread::sleep_for(std::chrono::milliseconds(250));
 
     ASSERT_EQ(clientDevice.getSignals()[0].getName(), "NewName");
     ASSERT_EQ(clientDevice.getSignals()[0].getDescription(), "NewDescription");
 
-    ASSERT_THROW(clientDevice.getSignals()[0].setName("ClientName"), AccessDeniedException);
+    ASSERT_NO_THROW(clientDevice.getSignals()[0].setName("ClientName"));
```
NikolaiShipilov, line 118: Shouldn't the use of `setName`/`setDescription` methods be restricted for mirrored signals, at least for pseudo device signals? 
On the flip side, the drawback of such approach would be the need for some alternative to the protected property interface, which was previously used to initialize/update `name` and `description` with values received via the streaming protocol.

NikolaiShipilov (approved):

shared/libraries/websocket_streaming/tests/test_websocket_client_device.cpp
JakaMohorkoDS: They are restricted in the sense that they do not do anything. The setters such as `setName / Description` do not throw an exception, but instead simply return `OPENDAQ_IGNORED`. By default the mirrored signals should have these fields locked.

shared/libraries/opcuatms/opcuatms_server/include/opcuatms_server/objects/tms_server_component.h
```diff
@@ -189,10 +188,53 @@ void TmsServerComponent<Ptr>::bindCallbacks()
...
+
     tmsPropertyObject->registerToExistingOpcUaNode(this->nodeId);
 }
 
+template <typename Ptr>
+void TmsServerComponent<Ptr>::onCoreEvent(const CoreEventArgsPtr& args)
+{
+    Super::onCoreEvent(args);
+
+    if (!selfChange && args.getEventId() == core_event_ids::AttributeChanged)
```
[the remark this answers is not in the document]
JakaMohorkoDS: I'll log the error. The setDisplayName and setDescription OPC UA calls throw an exception, so it should be caught in order to not crash the server app if it fails.

PR #54, opened 2023-11-30
<TBBAS-1200> Implement opcua attribute reader

janm-dw:
## Type of change

- [ ] New feature (non-breaking change which adds functionality)

# Checklist:

# Description:

### DIFF
--- a/shared/libraries/opcua/opcuaclient/include/opcuaclient/attribute_reader.h
+++ b/shared/libraries/opcua/opcuaclient/include/opcuaclient/attribute_reader.h
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2022-2023 Blueberry d.o.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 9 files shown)


JakaMohorkoDS (approved):
Lgtm. A couple concerns/questions:

- This looks like it would be used for a 1-time read operation of multiple nodes. I don't see a case where `read` would ever be called more than once. Is it worth having a `reset()` method? Maybe we could just "freeze" the object once `read` has been called.
- I'd use this internally in function calls on the client side that require you to read multiple attributes (I'm not 100% which those are at the moment, if any). At that point a new instance of it should be created.

janm-dw:
> Lgtm. A couple concerns/questions:
> 
> * This looks like it would be used for a 1-time read operation of multiple nodes. I don't see a case where `read` would ever be called more than once. Is it worth having a `reset()` method? Maybe we could just "freeze" the object once `read` has been called.
> * I'd use this internally in function calls on the client side that require you to read multiple attributes (I'm not 100% which those are at the moment, if any). At that point a new instance of it should be created.

Since we will have one global reader it would mean that we are able to do only one read request with it. I can very well se the use for reset() in the future, where we want to update part of the attributes.

PR #57, opened 2023-11-30
Add native config protocol framework

dejancrn:


shared/libraries/config_protocol/include/config_protocol/config_protocol.h
```diff
@@ -0,0 +1,110 @@
...
+
+#include <functional>
+#include <stdexcept>
+#include <string>
+#include <coretypes/string_ptr.h>
+
+namespace daq::config_protocol
+{
+
+enum PacketType: uint8_t { getProtocolInfo = 0x80, upgradeProtocol = 0x81, rpc = 0x82, serverNotification = 0x83, invalidRequest = 0x84 };
```
JakaMohorkoDS, line 27: Would it make sense to have these be a few bits apart? Eg. 0x80, 0x85, 0x90..?

shared/libraries/config_protocol/include/config_protocol/config_protocol_server.h
```diff
@@ -0,0 +1,49 @@
...
+    SerializerPtr serializer;
+    std::unordered_map<std::string, std::function<BaseObjectPtr(const DictPtr<IString, IBaseObject>&)>> rpcDispatch;
+
+    PacketBuffer processPacket(const PacketBuffer& packetBuffer);
+    StringPtr processRpc(const StringPtr& jsonStr);
+
+    BaseObjectPtr callRpc(const StringPtr& name, const DictPtr<IString, IBaseObject>& params);
+
+    BaseObjectPtr getComponent(const DictPtr<IString, IBaseObject>& params) const;
+    static ComponentPtr findComponent(const ComponentPtr& component, const std::string& id);
```
JakaMohorkoDS, line 46: There should probably be an internal hashmap of all components available either within the server, or somewhere that's accessible by all modules. Maybe in the context?

shared/libraries/config_protocol/include/config_protocol/config_protocol.h
dejancrn: I actually don't expect many more packet types. Almost all communication is either through rpc or notification packet type.

shared/libraries/config_protocol/include/config_protocol/config_protocol_server.h
dejancrn: Perhaps, but at the moment I do not expect performance issues and hashmap needs to be updated when new components are added and removed.

JakaMohorkoDS (approved):

PR #64, opened 2023-12-07
Cached reference browsing

janm-dw:


shared/libraries/opcuatms/opcuatms_client/src/objects/tms_client_context.cpp
```diff
@@ -48,4 +50,26 @@ opcua::OpcUaNodeId TmsClientContext::getNodeId(const BaseObjectPtr object) const
...
+    size_t maxNodesPerBrowse = 0;
+
+    try
+    {
+        const auto maxNodesPerBrowseId = OpcUaNodeId(UA_NS0ID_SERVER_SERVERCAPABILITIES_OPERATIONLIMITS_MAXNODESPERBROWSE);
+        maxNodesPerBrowse = client->readValue(maxNodesPerBrowseId).toInteger();
+    }
+    catch (const std::exception& e)
+    {
+        LOGW << "Failed to read maxNodesPerBrowse variable: " << e.what();
```
JakaMohorkoDS, line 69: Is it okay to always try to read everything at once if the server fails to communicate its capabilities?

JakaMohorkoDS (approved):

PR #79, opened 2023-12-21
<TBBAS-1199> Implement opcua attribute collector

janm-dw:


shared/libraries/opcuatms/opcuatms_client/src/objects/tms_client_object_impl.cpp
```diff
@@ -77,32 +77,17 @@ void TmsClientObjectImpl::subscriptionStatusChangeCallback(UA_StatusChangeNotifi
...
-        return VariantConverter<IInteger>::ToDaqObject(client->readValue(numberInListId));
-    }
-    catch(...)
-    {
-    }
-
-    return std::numeric_limits<uint32_t>::max();
+    const auto numberInListId = clientContext->getReferenceBrowser()->getChildNodeId(nodeId, "NumberInList");
+    const auto variant = clientContext->getAttributeReader()->getValue(numberInListId, UA_ATTRIBUTEID_VALUE);
+    return VariantConverter<IInteger>::ToDaqObject(variant);
```
JakaMohorkoDS, line 90: Will this now crash if the NumberInList node is either missing or null?

JakaMohorkoDS (approved):

shared/libraries/opcuatms/opcuatms_client/src/objects/tms_client_object_impl.cpp
JakaMohorkoDS: The main reason the "try" was there is for cases where someone does not implement NumberInList. I have a feeling this is not done in HBK currently. As such this was left there as a failsafe.

If we assume that the NumberInList is always configured, we also don't need to maintain the ordered/unordered lists.

PR #85, opened 2024-01-08
<TBBAS-1214> Add coretype wrapper implementation for enums

NikolaiShipilov:
## Type of change

# Checklist:

# Description:


bindings/python/core_types/generated/py_enumeration.cpp
```diff
@@ -0,0 +1,57 @@
+//------------------------------------------------------------------------------
```
dejancrn, line 1: Python wrapper for enum should inherit from python IntNum class and therefore it would behave as true python enum. But this is out of the scope of this PR and should be done as a separate task, probably by Yaroslav.

dejancrn (approved):

core/coretypes/include/coretypes/enumeration.h
```diff
@@ -0,0 +1,68 @@
...
+     * @brief Gets the Enumeration's type.
+     * @param[out] type The Enumeration type
+     */
+    virtual ErrCode INTERFACE_FUNC getEnumerationType(IEnumerationType** type) = 0;
+
+    /*!
+     * @brief Gets the Enumeration value as String containing the name of the enumerator constant.
+     * @param[out] value Emumeration value.
+     */
+    virtual ErrCode INTERFACE_FUNC getValue(IString** value) = 0;
```
JakaMohorkoDS, line 56: Are enumerations mutable, or immutable objects - as in, can we select another value after they are created? Looking at the definition of core types they should not be mutable, but I'm not sure whether they're useful in that case.

core/coretypes/include/coretypes/enumeration_type_impl.h
```diff
@@ -0,0 +1,56 @@
...
+
+class EnumerationTypeImpl : public GenericTypeImpl<IEnumerationType>
+{
+public:
+    explicit EnumerationTypeImpl(StringPtr typeName, ListPtr<IString> enumeratorNames, ListPtr<IInteger> enumeratorValues);
+    explicit EnumerationTypeImpl(StringPtr typeName, ListPtr<IString> enumeratorNames, Int firstEnumeratorValue);
+
+    // IEnumerationType
+    ErrCode INTERFACE_FUNC getEnumeratorNames(IList** names) override;
+    ErrCode INTERFACE_FUNC getEnumeratorIntValues(IList** values) override;
```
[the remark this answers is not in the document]
JakaMohorkoDS: This is modelled after the struct implementation. It came from a discussion with @anzeskerjanc where the request was that we would have an alternative to dictionaries when operating with the struct names/values.

JakaMohorkoDS (approved):
Does this allow enumeration-type properties to be created? We should probably prevent doing so as of now, or check at least in the OPC UA server that the property does not get published.

core/coretypes/include/coretypes/enumeration.h
NikolaiShipilov: As was discussed enumeration object kept immutable and there is an [addition](https://github.com/openDAQ/openDAQ/pull/85/commits/b47563cbee24adcda2656b4cbb17fe74d90768bd#diff-bf70b119ff04f0d923254e2dc9f807b4e36d0179f1cad158b9673e04f2987143R116) to smart pointer implementation for assigning a new enumeration object created with the given value.

PR #93, opened 2024-01-16
Freeze and clone nested object-type property default values

JakaMohorkoDS:
## Type of change

# Checklist:

# Description:

Object-type properties as of now behaved differently than any other property type. They were not frozen on add to property objects, but were frozen when added to property object classes. This was made more consistent by always freezing them, but also creating a local clone of the object that is returned to users on `getPropertyValue`. That clone can be edited.

Additionally, Property object core events now contain the path to the property if the property is contained within a nested property object.

### DIFF
--- a/bindings/python/opendaq/generated/component/py_component_private.cpp
+++ b/bindings/python/opendaq/generated/component/py_component_private.cpp
@@ -35,7 +35,7 @@ PyDaqIntf<daq::IComponentPrivate, daq::IBaseObject> declareIComponentPrivate(pyb
 
 void defineIComponentPrivate(pybind11::module_ m, PyDaqIntf<daq::IComponentPrivate, daq::IBaseObject> cls)
 {
-    cls.doc() = "";
+    cls.doc() = "Provides access to private methods of the component.";
 
     cls.def("lock_attributes",
         [](daq::IComponentPrivate *object, daq::IList* attributes)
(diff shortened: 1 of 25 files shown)


bindings/python/opendaq/generated/component/py_component_private.cpp
```diff
@@ -44,35 +44,35 @@ void defineIComponentPrivate(pybind11::module_ m, PyDaqIntf<daq::IComponentPriva
...
+        "Locks all attributes of the component.");
     cls.def("unlock_attributes",
         [](daq::IComponentPrivate *object, daq::IList* attributes)
         {
             const auto objectPtr = daq::ComponentPrivatePtr::Borrow(object);
             objectPtr.unlockAttributes(attributes);
         },
         py::arg("attributes"),
-        "");
+        "Removes the attributes contained in the list from the set of locked components.");
```
denise-opendaq, line 62: i see that another functions has description like lock/unlock attrubutes and this one has description which from the first look seems different. maybe rename it like "unlock set of attributes of the component"?

denise-opendaq (approved):

dejancrn (approved):
Looks good.

One thing that comes to my mind is how setting a property of object type will work over remote protocol. Should we even allow that? At the moment this is not supported in native configuration protocol. The property object that is set as a value must not be set directly on client side, but either deserialized from server reply or transformed from setPropertyValue parameter from plain property object to remote property object

JakaMohorkoDS:
> Looks good.
> 
> One thing that comes to my mind is how setting a property of object type will work over remote protocol. Should we even allow that? At the moment this is not supported in native configuration protocol. The property object that is set as a value must not be set directly on client side, but either deserialized from server reply or transformed from setPropertyValue parameter from plain property object to remote property object

We could just say that PropertyObjects need a default value, and only the properties of the default can be modified. It'd be a reasonable restriction in my opinion.

PR #95, opened 2024-01-16
<TBBAS-1138> Component statuses

NikolaiShipilov:
## Type of change

# Checklist:


core/opendaq/component/src/component_status_container_impl.cpp
```diff
@@ -0,0 +1,97 @@
...
+    if (valueObj.getEnumerationType() != oldValue.getEnumerationType())
+        return OPENDAQ_ERR_INVALIDTYPE;
+    if (valueObj == oldValue)
+        return OPENDAQ_IGNORED;
+
+    auto errCode = statuses->set(name, value);
+    if (OPENDAQ_FAILED(errCode))
+        return errCode;
+
+    if (onStatusChangedEvent.hasListeners())
```
dejancrn, line 86: I think we also need to trigger global event on context. Parameters are component and statuseventargs

dejancrn (changes requested):

core/opendaq/component/src/component_status_container_impl.cpp
JakaMohorkoDS: I'm implementing a similar mechanism for `ITags` at the moment. Changes should not be triggered in a local event, but use the global Context one instead (which in turn triggers the `IComponent` changed event).

I opened a PR for the current Tags rework - that can be used for reference. Note that it's still a draft PR.

https://github.com/openDAQ/openDAQ/pull/97

core/opendaq/component/include/opendaq/status_event_args.h
```diff
@@ -0,0 +1,68 @@
...
+     * @param[out] name The name of changed Component status.
+     */
+    virtual ErrCode INTERFACE_FUNC getStatusName(IString** name) = 0;
+
+    /*!
+     * @brief Gets the value of changed Component status.
+     * @param[out] value The value of changed Component status.
+     */
+    virtual ErrCode INTERFACE_FUNC getStatusValue(IEnumeration** value) = 0;
+};
```
JakaMohorkoDS, line 58: This should use the CoreEventArgs interface and mechanism. The StatusChanged on a component should be a core event.

core/opendaq/component/include/opendaq/component_status_container.h
```diff
@@ -0,0 +1,75 @@
...
+    virtual ErrCode INTERFACE_FUNC getStatuses(IDict** statuses) = 0;
+
+    // [templateType(event, IComponentStatusContainer, IStatusEventArgs)]
+    /*!
+     * @brief Gets the Status Event object that triggers whenever there is a change in the status of
+     * the openDAQ Component.
+     * @param[out] event The Status Event object. The event triggers with a Component Status
+     * reference and a StatusEventArgs object as arguments.
+     */
+    virtual ErrCode INTERFACE_FUNC getOnStatusChanged(IEvent** event) = 0;
```
JakaMohorkoDS, line 68: I don't think this is necessary, I'd just use the Component's core event and the global Context core event.

JakaMohorkoDS (changes requested):
Lgtm, but should use the Core event mechanism for "changed event" handling.

Notably, we now have "locked attributes" in the core. Do the Status container and Tags classify as Attributes, or separate objects? If they are attributes, should they be locked/unlocked like name, visible, active, description? 

I'm currently working on the `ITags` rework where I added the locking mechanism, and included the `add` and `remove` functions on `ITags` instead of adding `ITagsPrivate`. Status container and Tags seem like really similar objects, so we should handle them in the same manner.

Do we go with a private interface, or attribute locking? If we choose the former, should we create two new core events (status changed, tags changed) instead of using Attribute changed for those?

JakaMohorkoDS (approved):

JakaMohorkoDS:
I think this PR has been reviewed and is okay to be merged with one addition - would it be possible to parse the OPC UA signal status on the client-side and encode it as the signal's set of statuses? 

That should probably be part of a different PR though. In my view this is ready to be merged.

JakaMohorkoDS:
@dejancrn Can this be merged? Think that GitHub does not allow merging before a 2nd review has been done when changes were requested.

dejancrn (approved):
Looks ok, however, I think we need an example implementation in one of the reference function blocks. Details will follow.

PR #103, opened 2024-01-19
[TBBAS-939] Error recovery and logging in OPC UA client. Add debug logger sink

denise-opendaq:
Surround OpcUa client with try catches and add readable warnings.

# Checklist:

# Description:
- wrapped OpcUa client components with try catch
- add warnigs
- implement custom debug logger sink which stores last log messages
- update test which expected getting and exception which checking log message


shared/libraries/opcuatms/opcuatms_client/src/objects/tms_client_component_impl.cpp
```diff
@@ -56,12 +70,17 @@ ErrCode TmsClientComponentBaseImpl<Impl>::getName(IString** name)
...
     {
-        StringPtr nameObj =this->client->readDisplayName(this->nodeId);
-        *name = nameObj.detach();
-        return OPENDAQ_SUCCESS;
-    });
+        nameObj = this->client->readDisplayName(this->nodeId);
+    }
+    catch(...)
+    {
+        nameObj = "unknown name";
```
JakaMohorkoDS, line 80: I'd just return empty strings here, rather than "unknown name", or default to the core behaviour where the local ID is returned, if the name is empty.

shared/libraries/opcuatms/opcuatms_client/src/objects/tms_client_component_impl.cpp
```diff
@@ -89,12 +108,18 @@ ErrCode TmsClientComponentBaseImpl<Impl>::getDescription(IString** description)
 {
     OPENDAQ_PARAM_NOT_NULL(description);
 
-    return daqTry([&]
+    try
     {
         StringPtr descObj = this->client->readDescription(this->nodeId);
         *description = descObj.detach();
```
JakaMohorkoDS, line 114: Same comment as the `active` getter. The same applies for all other getters (I won't add a comment to each instance).

shared/libraries/opcuatms/opcuatms_client/src/objects/tms_client_component_impl.cpp
```diff
@@ -144,12 +169,17 @@ ErrCode TmsClientComponentBaseImpl<Impl>::setVisible(Bool visible)
     }
     catch (...)
     {
-        const auto loggerComponent = this->daqContext.getLogger().getOrAddComponent("OpcUaClient");
-        LOG_D("OPC UA Component {} failed to set \"Active\" state.", this->localId);
+        const auto loggerComponent = getLogger();
```
JakaMohorkoDS, line 172: I'd rename `getLogger` to `getLoggerComponent`, and maybe add it to some `details::` namespace.

shared/libraries/opcuatms/opcuatms_client/src/objects/tms_client_device_impl.cpp
```diff
@@ -360,7 +388,7 @@ void TmsClientDeviceImpl::findAndCreateStreamingOptions()
     }
     catch (const std::exception& e)
     {
-        LOG_E("Failed to find 'StreamingOptions' OpcUa node: {}", e.what());
+        LOG_W("Failed to find 'StreamingOptions' OpcUA node on OpcUA client device \"{}\": {}", this->localId, e.what());
```
JakaMohorkoDS, line 391: I'm not 100% on if all these logs should be warnings. I think some are okay to just be debug messages, as getting the default values is fine. 

Eg. any setter that in the core also just returns OPENDAQ_IGNORED should log it at the debug or info level. Similarly for example `getName` could just return the local ID and log at debug level, that the OPC UA call failed.

shared/libraries/opcuatms/opcuatms_client/src/objects/tms_client_property_object_impl.cpp
```diff
@@ -22,38 +22,55 @@ using namespace opcua;
...
             {
-                return this->makeErrorInfo(OPENDAQ_ERR_NOTIMPLEMENTED, "Object type properties cannot be set over OPC UA");
+                lastProccessDescription = "Object type properties cannot be set over OpcUA";
+                return OPENDAQ_ERR_NOTIMPLEMENTED;
             }
-
+            lastProccessDescription = "Property not found";
             return OPENDAQ_ERR_NOTFOUND;
         });
+    if (OPENDAQ_FAILED(errCode))
```
JakaMohorkoDS, line 69: I'd return an error for `OPENDAQ_NOT_FOUND` here, similarly as done in the core.

shared/tools/RTGen/src/project/RTGen/Examples/async_channel_buffer.h
```diff
@@ -108,18 +108,20 @@ OPENDAQ_DECLARE_CLASS_FACTORY_WITH_INTERFACE_AND_CREATEFUNC(LIBRARY_FACTORY,
...
+OPENDAQ_DECLARE_CLASS_FACTORY_WITH_INTERFACE_AND_CREATEFUNC(LIBRARY_FACTORY,
+    InputGroupTypeWithCallback,
+    IInputGroupType,
+    InputGroupTypeWithCallback_Create,
+    IString*,
+    type,
+    IFuncObject*,
+    onGetCount,
+    Int,
+    maxCount);
```
JakaMohorkoDS, line 124: Is this change supposed to be here?

shared/libraries/opcuatms/opcuatms_client/src/objects/tms_client_component_impl.cpp
```diff
@@ -12,20 +12,33 @@ using namespace daq::opcua;
...
         this->template writeValue<IBoolean>("Active", active);
-        return OPENDAQ_SUCCESS;
-    });
+    }
+    catch(...)
+    {
+        auto loggerComponent = getLoggerComponent();
+        LOG_W("Failed to set active of component \"{}\"", this->globalId);
+    }
+    return OPENDAQ_SUCCESS;
```
JakaMohorkoDS, line 41: I think setters should return OPENDAQ_IGNORED, and log the fact at debug or info level. The core implementation does it in a similar manner (at info). The previous comment about returning OPENDAQ_IGNORED was aimed at getters, not setters.

```
...
        if (lockedAttributes.count("Name"))
        {
            if (context.assigned() && context.getLogger().assigned())
            {
                const auto loggerComponent = context.getLogger().getOrAddComponent("Component");
                StringPtr descObj;
                this->getName(&descObj);
                LOG_I("Name of {} is locked", descObj);
            }

            return OPENDAQ_IGNORED;
        }

        this->name = name;
...
``` 

Getters could also be left at "warning level", but I have a feeling we might flood stdout with that. The log could be changed to:

`LOG_I("Failed to get active of component \"{}\". The default value was returned (\"true\")", this->globalId);`

shared/libraries/opcuatms/opcuatms_client/src/objects/tms_client_device_impl.cpp
```diff
@@ -88,20 +88,27 @@ void TmsClientDeviceImpl::findAndCreateSubdevices()
...
-        auto subdeviceNodeId = OpcUaNodeId(ref->nodeId.nodeId);
-        auto clientSubdevice = TmsClientDevice(context, devices, browseName, clientContext, subdeviceNodeId, createStreamingCallback);
-                    
-        auto numberInList = this->tryReadChildNumberInList(subdeviceNodeId);
-        if (numberInList != std::numeric_limits<uint32_t>::max() && !orderedDevices.count(numberInList))
-            orderedDevices.insert(std::pair<uint32_t, ComponentPtr>(numberInList, clientSubdevice));
-        else
-            unorderedDevices.emplace_back(clientSubdevice);
+        try
+        {
```
janm-dw, line 92: Why is this suddenly in a try catch?

shared/libraries/opcuatms/tests/opcuatms_integration/tms_object_integration_test.cpp
```diff
@@ -5,15 +5,35 @@ using namespace daq;
 using namespace daq::opcua;
 using namespace daq::opcua::tms;
 
+static LoggerPtr createLoggerWithDebugSink(const LoggerSinkPtr& sink)
```
janm-dw, line 8: PascalCase?

core/opendaq/logger/include/opendaq/logger_sink_last_message_impl.h
```diff
@@ -0,0 +1,104 @@
...
+#include <condition_variable>
+#include <chrono>
+
+namespace spdlog {
+namespace sinks {
+
+template<typename Mutex>
+class LoggerSinkLastMessage : public base_sink<Mutex>
+{
+protected:
```
janm-dw, line 30: Public on top, then protected, then private

janm-dw:
Three notes (second one worries me the most):
- When catching an exception, please try to be as specific ass possible with the exception type. Please try void catch(...). Consider catch(const OpcUaException& e) and then also maybe logging the exception message.
- You added try / catch inside every `for (const auto& [browseName, ref] : references.byBrowseName)` loop. Why is that? When catching an exception, you should have a very good reason for that. Adding try catch statements like that hides structure errors and makes application harder to debug.
- And when you really need to add a try catch statement around a larger block of code, you should consider extracting that code to a separate private method.

janm-dw (changes requested):

shared/libraries/opcuatms/opcuatms_client/src/objects/tms_client_component_impl.cpp
```diff
@@ -12,20 +12,34 @@ using namespace daq::opcua;
...
-{
-    return daqTry([&]() {
+{   
+    try
+    {
         *active = this->template readValue<IBoolean>("Active");
-        return OPENDAQ_SUCCESS;
-    });
+    }
+    catch(...)
```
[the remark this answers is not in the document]
JakaMohorkoDS: In general, it was a request to not fail on connect, unless it's a critical error. Invalid nodes should simply be skipped and the error logged.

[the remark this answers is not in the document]
janm-dw: Yes, but we need to be way more conservative about this. In this case we should catch only opcua communication / permission errors. 
- if node is not of correct type (as specified in TMS model), we should fail
- if a **mandatory** node is missing, we should fail
It is the only way of how to enforce correct device structure.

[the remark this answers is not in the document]
JakaMohorkoDS: I don't really agree.

Basically the request was to make the client-side of OPC UA more lenient towards smaller mistakes/errors on the serve-side model design. This to me means that on connect we should not fail when the structure of the device is slightly malformed - we should just ignore the malformed nodes.

As for function calls - those need to be evaluated on an individual basis. For me, `getName` for example should not fail, but just return the local ID, as that's good enough. Some other calls, on the other hand, should maybe throw an error.

The structure on our side will be enforced no matter what - if the server has a malformed part of the tree, that part will simply not appear on the client side.

I do agree though that we could go and review all the try catches here, and possibly sometimes log and rethrow.

JakaMohorkoDS (approved):

shared/libraries/opcuatms/opcuatms_client/src/objects/tms_client_component_impl.cpp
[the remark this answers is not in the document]
janm-dw: I do agree about the getName(). However we should still not be wrapping everything in try catch. 

In my opinion we should only skip objects at component level. Or maybe property property level at most. That means that if a component is missing a mandatory node, we should skip the whole component. Of course, we can try to assume the default value for that node if it makes sense, but in most cases, we should skip the whole component.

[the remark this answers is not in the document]
JakaMohorkoDS: I mean, I'd think we skip it at the bottom-most level. If a device is missing a component, we should just skip the component. If a property has missing fields, we can probably just revert to defaults - the server should handle things like "readOnly" anyhow. 

I think the intended philosophy was to be more forgiving on connect, to parse as much of the structure as possible, even if the server does not 100% adhere to all modelling rules (eg. some nodes are null). Function calls should probably still throw errors though. 

We decide to log + rethrow now, so that the rethrow can be removed at any point if deemed necessary.

Personally, I think this PR is an improvement over the previous state where throwing/logging was inconsistently handled, and should solve a potential issues on the HBK integration side.

janm-dw (approved):

PR #105, opened 2024-01-24
Add support for setting property values on object properties

dejancrn:


JakaMohorkoDS:
Handled in https://github.com/openDAQ/openDAQ/pull/191

PR #111, opened 2024-01-26
Project/native config core events

JakaMohorkoDS:
## Type of change

# Checklist:

# Description:

# TODO:

Issues found so far:

- [bug] Replay changes on `applyUpdate` instead of copying values
- [feature] begin/end update remote call
- [feature] Throw error on not-allowed RPC functions
- [feature] Toggling and usage of the "public" attribute
- [feature] Config client tags
- [feature] Client objects update mechanism
- [feature] Implement missing RPC functions
- [feature] Device flag that marks a device with server-side event support

### DIFF
--- a/core/coreobjects/include/coreobjects/core_event_args_ids.h
+++ b/core/coreobjects/include/coreobjects/core_event_args_ids.h
@@ -25,22 +25,22 @@ BEGIN_NAMESPACE_OPENDAQ
  * @{
  */
 
-namespace core_event_ids
+enum class CoreEventId : uint32_t
 {
-    constexpr Int PropertyValueChanged = 0;
-    constexpr Int PropertyObjectUpdateEnd = 10;
-    constexpr Int PropertyAdded = 20;
-    constexpr Int PropertyRemoved = 30;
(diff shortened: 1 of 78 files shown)


core/coreobjects/include/coreobjects/core_event_args_ids.h
```diff
@@ -41,6 +41,22 @@ namespace core_event_ids
     constexpr Int TagsChanged = 110;
 }
 
+enum class CoreEventId
```
dejancrn, line 44: Feels a little hacky that OpenDAQ events are defined in coreobjects

dejancrn (approved):

core/coreobjects/include/coreobjects/core_event_args_ids.h
JakaMohorkoDS: Technically we could place all of it into opendaq, but then the property object implementation would not be able to create its own arguments.

As a side-note: I've added the enum class for the event IDs because I wanted to use a switch statement. Should we use the `constexpr` above, the enum class, or both?

PR #115, opened 2024-01-28
Transfer type manager via native config protocol and register the types

dejancrn:
Added test for struct properties. Enum properties (when implemented) should also work.

### DIFF
--- a/shared/libraries/config_protocol/src/config_protocol_client.cpp
+++ b/shared/libraries/config_protocol/src/config_protocol_client.cpp
@@ -66,7 +66,10 @@ BaseObjectPtr ConfigProtocolClientComm::getPropertyValue(const std::string& glob
     auto getPropertyValueRpcRequestPacketBuffer = createRpcRequestPacketBuffer(generateId(), "GetPropertyValue", dict);
     const auto getPropertyValueRpcReplyPacketBuffer = sendRequestCallback(getPropertyValueRpcRequestPacketBuffer);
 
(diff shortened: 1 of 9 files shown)


shared/libraries/config_protocol/src/config_protocol_client.cpp
```diff
@@ -313,6 +324,24 @@ void ConfigProtocolClient::connect(const ComponentPtr& parent)
...
+    const auto types = typeManager.getTypes();
+
+    for (const auto& typeName : types)
+    {
+        const auto type = typeManager.getType(typeName);
+        if (localTypeManager.hasType(type.getName()))
+        {
+            const auto localType = localTypeManager.getType(type.getName());
+            if (localType != type)
+                throw InvalidValueException("Remote type different than local");
```
JakaMohorkoDS, line 338: I would probably just log this, and try to use the local types instead with this device. Then if they're incompatible, either throw an error later, or log that as well.

Throwing an exception while connecting seems like a bit overkill, as you can't handle the exception with an application-level wrapper.

JakaMohorkoDS (approved):

shared/libraries/config_protocol/src/config_protocol_client.cpp
dejancrn: This will actually throw an exception inside "Connect" call. There are other reasons why Connect can throw exceptions. The client code / native config module should in any case wrap the connect call in try/catch statement and then rethrow or handle the error in addDevice call.

PR #123, opened 2024-02-04
Support data packets with struct data type and implement CAN channel …

dejancrn:
Missing/TODO:
- update stream reader to support struct sample type
- support struct sample in ISignal.getLastValue


JakaMohorkoDS (approved):

JakaMohorkoDS:
Maybe an integration test with StreamingLT and NativeStreaming would be useful. In the 1st case, I think the streaming should not be able to parse the signal, while in the 2nd, we should get data.

PR #137, opened 2024-02-12
Added statistics function block tests and statistics bug fix

aljazfrancic:
## Type of change

# Checklist:

# Description:
Added a few statistics function block tests and fixed a bug where start was not used for explicit domain data calculation in statistics function block.


aljazfrancic:
Found some problems with the PR. I will fix them soon.

modules/ref_fb_module/src/statistics_fb_impl.cpp
```diff
@@ -19,16 +22,32 @@ StatisticsFbImpl::StatisticsFbImpl(const ContextPtr& ctx, const ComponentPtr& pa
...
+        else
+            packetReadyNotification = PacketReadyNotification::SameThread;
+    }
+    else
+    {
+        packetReadyNotification = PacketReadyNotification::Scheduler;
+    }
+
+    createAndAddInputPort("input", packetReadyNotification);
 }
```
JakaMohorkoDS, line 38: This can be simplified to 
```
    packetReadyNotification = PacketReadyNotification::Scheduler;
    if (config.assigned() && config.hasProperty("UseMultiThreadedScheduler"))
    {
        if (!config.getPropertyValue("UseMultiThreadedScheduler"))
            packetReadyNotification = PacketReadyNotification::SameThread;
    }
```

Or if you prefer 1 liners:
```
const bool useScheduler = config.assigned() && 
                          config.hasProperty("UseMultiThreadedScheduler") && 
                          config.getPropertyValue("UseMultiThreadedScheduler");
packetReadyNotification = useScheduler ? PacketReadyNotification::Scheduler : PacketReadyNotification::SameThread;
```

modules/ref_fb_module/tests/test_fb_statistics.cpp
```diff
@@ -0,0 +1,428 @@
...
+                              {1.0, 1.1, 1.2},
+                              {1.3, 1.4, 1.5},
+                              {1.6, 1.7, 1.8},
+                              {1.9, 2.0, 2.1},
+                              {2.2, 2.3, 2.4},
+                              {2.5, 2.6, 2.7},
+                              {2.8, 2.9, 3.0}};
+    vecvec<Float> expectedAvg{{}, {}, {}, {0.55}, {}, {}, {1.55}, {}, {}, {2.55}};
+    vecvec<Float> expectedRms{{}, {}, {}, {0.62048368229954287}, {}, {}, {1.5763882770434448}, {}, {}, {2.5661254840712679}};
+    vecvec<Int> expectedDomain{{}, {}, {}, {3}, {}, {}, {23}, {}, {}, {43}};
```
JakaMohorkoDS, line 304: Are we expecting the statistics fb to output empty packets here?

JakaMohorkoDS (approved):
Looks good. In general you might find that such a test suite with an increasing number of test conditions can become hard to extend/maintain. 

It's okay as is, but it could be worth, if we ever are required to write another similar test suite for a new reference function block, to refactor the helpers to use a common class template with overridable virtual methods. These virtual methods should contain some of the most common default behaviours that can be adjusted by more specific helper implementations.

modules/ref_fb_module/tests/test_fb_statistics.cpp
aljazfrancic: We are not expecting any packets back at that point. It is up to the function block implementation to store input packet values in the buffer unit the BlockSize is reached, which in this example is after the fourth, seventh and tenth input packets.

modules/ref_fb_module/src/statistics_fb_impl.cpp
aljazfrancic: The first one I agree with, but I think in the 1 liner example is wrong here, because we want to default to Scheduler in case the property is not set.

I think the correct 1 liner would be:
```
const bool useScheduler = config.assigned() && 
                          config.hasProperty("UseMultiThreadedScheduler") && 
                          config.getPropertyValue("UseMultiThreadedScheduler");
packetReadyNotification = !useScheduler ? PacketReadyNotification::SameThread : PacketReadyNotification::Scheduler;
```

However I prefer the version from your first example. I'll update the code accordingly.

aljazfrancic: Or even more concise:
```
packetReadyNotification = PacketReadyNotification::Scheduler;
if (config.assigned() && config.hasProperty("UseMultiThreadedScheduler") && !config.getPropertyValue("UseMultiThreadedScheduler"))
{
    packetReadyNotification = PacketReadyNotification::SameThread;
}
```
I'll update in both trigger and statistics function block.

PR #146, opened 2024-02-13
implement logic for signal.setPublic(...)

denise-opendaq:

# Checklist:

# Description:


JakaMohorkoDS (approved):
Lgtm for the OPC UA side of things. Should wait for @NikolaiShipilov for the streaming implementation.

Also as far as I can tell, we are still missing the StreamingLT (WebsocketStreamingServer) implementation?

NikolaiShipilov:
Is it possible that the data signal is public and its domain signal is not public? If so, then a non-public domain signal should still be published in streaming, at least in the native one, since domain packets are streamed explicitly within the native streaming.

NikolaiShipilov (approved):

JakaMohorkoDS:
> ~@JakaMohorkoDS Is it possible that the data signal is public and its domain signal is not public? If so, then a non-public domain signal should still be published in streaming, at least in the native one, since domain packets are streamed explicitly within the native streaming.~
> 
> > Also as far as I can tell, we are still missing the StreamingLT (WebsocketStreamingServer) implementation?
> 
> That one could be placed [here](https://github.com/openDAQ/openDAQ/blob/main/shared/libraries/websocket_streaming/src/streaming_server.cpp#L138)
> 
> @JakaMohorkoDS Since private signals are not published over OpcUa shouldn't these also be excluded in Native config protocol ?

Good point, though I'm not 100% certain at which point that should be implemented. Maybe on the deserialization side where the signal would not be attached to the tree if it's not public?

nilsRoettgerAtBB:
Can this be merged now? I am waiting for it, that it goes into main :-)

JakaMohorkoDS (approved):

PR #154, opened 2024-02-16
Implement recursive begin and end update

dejancrn:


core/coreobjects/include/coreobjects/end_update_event_args.h
```diff
@@ -28,11 +28,13 @@ DECLARE_OPENDAQ_INTERFACE(IEndUpdateEventArgs, IEventArgs)
 {
     // [elementType(properties, IString)]
     virtual ErrCode INTERFACE_FUNC getProperties(IList** properties) = 0;
+    virtual ErrCode INTERFACE_FUNC getIsParentUpdating(Bool* isParentUpdating) = 0;
```
JakaMohorkoDS, line 31: Do we also require this on CoreEvents?

JakaMohorkoDS, line 31: It'd be great to have some javadoc comments on this.

core/coreobjects/include/coreobjects/property_object_impl.h
```diff
@@ -1745,11 +1754,61 @@ ErrCode GenericPropertyObjectImpl<PropObjInterface, Interfaces...>::beginUpdate(
...
+        if (value.assigned())
+        {
+            const auto propObj = value.template asPtrOrNull<IPropertyObject>(true);
+            if (propObj.assigned())
+            {
+                propObj.endUpdate();
+            }
+        }
+    }
+}
```
JakaMohorkoDS, line 1792: This behaviour is right now hidden from users - there was no API or documentation change that came with this addition. We should maybe consider implementing this as an optional parameter - eg. `beginUpdate(bool recursive = true)`. At least the javadoc documentation must explain the new behaviour.

JakaMohorkoDS: As agreed, a change to the Doxygen documentation is sufficient.

PR #156, opened 2024-02-16
<TBBAS-1228> Add support for adding function blocks with config object

janm-dw:
- support for getAvailableFunctionBlockTypes() on client side
- support adding of function blocks with config object


core/opendaq/functionblock/include/opendaq/function_block_type_impl.h
```diff
@@ -38,6 +40,105 @@ class FunctionBlockTypeImpl : public GenericComponentTypeImpl<IFunctionBlockType
...
+            StringPtr description;
+            if (serializedObj.hasKey("description"))
+                description = serializedObj.readString("description");
+
+            FunctionPtr createDefaultConfig;
+            if (serializedObj.hasKey("defaultConfig"))
+            {
+                PropertyObjectPtr defaultConfig = serializedObj.readObject("defaultConfig", contextPtr, factoryCallbackPtr);
+                createDefaultConfig = Function([defaultConfig] { return defaultConfig; });
+            }
```
JakaMohorkoDS, line 136: Not sure if it's really correct to deserialize the object in this manner. Also, I think that `defaultConfig` returned by the lambda here will be nullptr, as the property object ptr will go out of scope.

I don't really think there's a good way of implementing this if we don't have a `clone` method for the property object.

In general, this is also missing a test case.

shared/libraries/opcuatms/opcuatms/tests/test_property_object_conversion_utils.cpp
```diff
@@ -53,3 +53,37 @@ TEST_F(PropertyObjectConversionUtilsTest, EmptyObject)
...
+    // Serializer fails to serialize default empty list property correctly.
+
+    auto obj = CreateTestPropertyObject();
+    obj.addProperty(ListProperty("list", List<IInteger>()));
+    obj.setPropertyValue("list", List<IInteger>(1, 2, 3));
+
+    auto clone = PropertyObjectConversionUtils::ClonePropertyObject(obj);
+
+    ASSERT_TRUE(TestComparators::PropertyObjectEquals(obj, clone));
+}
```
JakaMohorkoDS, line 89: Can you add this as a disabled coretypes test and open a Jira bug report?

shared/libraries/opcuatms/opcuatms_client/src/objects/tms_client_device_impl.cpp
```diff
@@ -450,29 +451,26 @@ void TmsClientDeviceImpl::findAndCreateCustomComponents()
 
 DictPtr<IString, IFunctionBlockType> TmsClientDeviceImpl::onGetAvailableFunctionBlockTypes()
 {
-    // todo: migrate to AvailableTypes folder node
+    auto browser = clientContext->getReferenceBrowser();
 
-    const auto fbFolderNodeId = getNodeId("FB");
-    const auto methodNodeId = clientContext->getReferenceBrowser()->getChildNodeId(fbFolderNodeId, "_TmpGetAvailableFunctionBlockTypes");
+    const auto fbFolderNodeId = browser->getChildNodeId(nodeId, "FB");
+    const auto availableTypesId = browser->getChildNodeId(fbFolderNodeId, "AvailableTypes");
```
JakaMohorkoDS, line 457: Note that all of these nodes are optional, as they're not in the model/only available in the SDK. As such when connecting to server without the `AvailableTypes` and `DefaultConfig` nodes, it should not crash.

In general, this PR should be backwards compatible with the old main branch. Was this tested?

JakaMohorkoDS (changes requested):

shared/libraries/opcuatms/opcuatms_client/src/objects/tms_client_device_impl.cpp
janm-dw: onGetAvailableFunctionBlockTypes() threw an exception before my changes (because it was not implemented). So if the AvailableTypes node is missing we can throw an exception and we are still backwards compatible. Or we can return an empty list. Which one do you want to do?

JakaMohorkoDS: I'd return an empty list.

core/opendaq/functionblock/include/opendaq/function_block_type_impl.h
janm-dw: `defaultConfig` is passed to lambda by value, so the ref count should be incremented. Since this code is just moved from cpp to h, I would spend much time on it (at least not under the scope of this ticket).

JakaMohorkoDS (approved):

PR #159, opened 2024-02-20
Statistics nested trigger

aljazfrancic:
## Type of change

# Checklist:

# Description:
- Adds a working trigger function block as a nested function block in statistics function block with tests
- Statistics uses the trigger outputs to determine when it should calculate, and when not to
- The trigger function block can be added based on a Boolean property of the statistics function block

### DIFF
--- a/modules/ref_fb_module/include/ref_fb_module/statistics_fb_impl.h
+++ b/modules/ref_fb_module/include/ref_fb_module/statistics_fb_impl.h
@@ -57,7 +57,7 @@ public:
     Bool getTriggerStateFromDomainValue(Int domainValue)
     {
         Int indexBefore = -1;
-        for (int i = 0; i < domainValues.size(); i++)
+        for (size_t i = 0; i < domainValues.size(); i++)
         {
             if (domainValues[i] <= domainValue)
                 indexBefore = i;
--- a/modules/ref_fb_module/include/ref_fb_module/statistics_fb_impl.h
+++ b/modules/ref_fb_module/include/ref_fb_module/statistics_fb_impl.h
@@ -183,6 +183,7 @@ private:
     void onPacketReceived(const InputPortPtr& port) override;
     void processTriggerPackets(const InputPortPtr& port);
     void processInputPackets(const InputPortPtr& port);
+    void calculateAndSendPackets(const DataPacketPtr& domainPacket, const DataPacketPtr& packet);
 };
 
 }
--- a/modules/ref_fb_module/include/ref_fb_module/statistics_fb_impl.h
+++ b/modules/ref_fb_module/include/ref_fb_module/statistics_fb_impl.h
@@ -26,7 +26,7 @@ BEGIN_NAMESPACE_REF_FB_MODULE
 namespace Statistics
 {
 
-// TODO
+// TODO Test all methods
 class TriggerHistory
 {
 public:
@@ -36,16 +36,32 @@ public:
         domainValues.push_back(domainValue);
     }
 
-    void dropHistory(Int toExcludingDomainValue)
+    // TODO USE THIS
+    void dropHistory(Int dropToExcludingDomainValue)
     {
-        auto whereTo = std::find(values.begin(), values.end(), toExcludingDomainValue);
-        if (whereTo != values.end())
+        auto whereTo = std::find(domainValues.begin(), domainValues.end(), dropToExcludingDomainValue);
+        if (whereTo != domainValues.end())
         {
-            // int index = it - v.begin();
-
-            // values.erase();
-            // domainValues.erase(domainValues.begin(), whereTo);
+            auto index = whereTo - domainValues.begin();
+            //  [first last)
+            values.erase(values.begin(), values.begin() + index);
+            domainValues.erase(domainValues.begin(), domainValues.begin() + index);
         }
+    }
+    Bool getTriggerStateFromDomainValue(Int domainValue)
+    {
+        Int indexBefore = -1;
+        for (int i = 0; i < domainValues.size(); i++)
+        {
+            if (domainValues[i] <= domainValue)
+                indexBefore = i;
+            else
+                break;
+        }
+        if (indexBefore == -1)
+            return false;
+        else
+            return values[indexBefore];
     }
 
 private:
@@ -97,9 +113,9 @@ private:
     };
 
     bool triggerMode;
-    bool doWork;
     FunctionBlockPtr nestedTriggerFunctionBlock;
     InputPortPtr triggerOutput;
+    TriggerHistory triggerHistory;
 
     size_t blockSize;
     DomainSignalType domainSignalType;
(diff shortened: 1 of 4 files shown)


modules/ref_fb_module/src/statistics_fb_impl.cpp
```diff
@@ -244,15 +281,19 @@ void StatisticsFbImpl::processSignalDescriptorChanged(const DataDescriptorPtr& v
...
 
-void StatisticsFbImpl::processDataPacket(const DataPacketPtr& packet)
+void StatisticsFbImpl::processDataPacketTrigger(const DataPacketPtr& packet)
 {
-    if (!valid)
-        return;
-
     const auto domainPacket = packet.getDomainPacket();
-    if (!domainPacket.assigned())
-        return;
```
JakaMohorkoDS, line 254: Is this check no longer required?

modules/ref_fb_module/src/statistics_fb_impl.cpp
```diff
@@ -244,15 +281,19 @@ void StatisticsFbImpl::processSignalDescriptorChanged(const DataDescriptorPtr& v
...
+    auto data = static_cast<Bool*>(packet.getData());
+    // Data packet from trigger only holds one value by design
+    auto triggerData = data[0];
+    // Domain packet from trigger only holds one value by design
+    auto domainStamp = static_cast<Int*>(domainPacket.getData())[0];
+    triggerHistory.addElement(triggerData, domainStamp);
```
JakaMohorkoDS, line 292: Maybe a check for packet sizes would be appropriate before accessing at index 0.

modules/ref_fb_module/src/statistics_fb_impl.cpp
```diff
@@ -310,6 +351,63 @@ void StatisticsFbImpl::processDataPacket(const DataPacketPtr& packet)
...
+        return;
+
+    const auto domainPacket = packet.getDomainPacket();
+    if (!domainPacket.assigned())
+        return;
+
+    if (triggerMode)
+    {
+        const auto packetBuf = static_cast<uint8_t*>(packet.getData());
+        auto domainBuf = static_cast<Int*>(domainPacket.getData());
```
JakaMohorkoDS, line 366: There's a lot of assumptions about the data types of the domain and trigger signal here. I'd still create a short function `validateTriggerDescriptors` that's called when a `DataDescriptorChanged` packet is received on the trigger input port.

modules/ref_fb_module/include/ref_fb_module/statistics_fb_impl.h
```diff
@@ -69,6 +112,11 @@ class StatisticsFbImpl final : public FunctionBlock
         }
     };
 
+    bool triggerMode;
+    FunctionBlockPtr nestedTriggerFunctionBlock;
+    InputPortPtr triggerOutput;
```
JakaMohorkoDS, line 117: `triggerInput` seems like a more fitting name.

modules/ref_fb_module/src/statistics_fb_impl.cpp
```diff
@@ -503,10 +601,34 @@ void StatisticsFbImpl::calculate(
 
 void StatisticsFbImpl::onPacketReceived(const InputPortPtr& port)
 {
-    processPackets(port);
+    if (port.getLocalId() == "trigger")
```
JakaMohorkoDS, line 604: You can do object comparison here instead of String comparsion - if you compare 2 ObjectPtr to each other, unless they have `equals()` overridden, they'll check whether they point to the same object. This can be replaced with: 

```
if (port == triggerOutput)
    ...
```

modules/ref_fb_module/src/statistics_fb_impl.cpp
```diff
@@ -310,6 +351,63 @@ void StatisticsFbImpl::processDataPacket(const DataPacketPtr& packet)
...
+    {
+        const auto packetBuf = static_cast<uint8_t*>(packet.getData());
+        auto domainBuf = static_cast<Int*>(domainPacket.getData());
+
+        size_t i = 0;
+        while (i < packet.getSampleCount())
+        {
+            // Go to where trigger state is true or end of packet
+            while (!triggerHistory.getTriggerStateFromDomainValue(domainBuf[i]) && i < packet.getSampleCount())
+                i++;
```
JakaMohorkoDS, line 373: I think there's still a TODO for dropping the triggerHistory, but it should probably be dropped within/after this while loop.

modules/ref_fb_module/tests/test_fb_statistics.cpp
```diff
@@ -426,3 +518,42 @@ TEST_F(StatisticsTest, StatisticsTestChangingOutputDomain)
...
+                                       {},
+                                       {},
+                                       triggerModeChangesBeforePackets,
+                                       newTriggerMode,
+                                       mockTriggerPackets,
+                                       mockTriggerDomainPackets);
+    helper.run();
+}
+
+// TODO add more trigger as nested function block tests, test multiple trigger condition changes in same packet
```
JakaMohorkoDS, line 559: I don't think too many more tests are needed, mostly I'd add a test where the trigger goes from 0 to 1 several times.

JakaMohorkoDS (commented):
Lgtm, I'll do another review once it's not in draft anymore.

modules/ref_fb_module/include/ref_fb_module/statistics_fb_impl.h
aljazfrancic: `triggerInput` instead of `triggerOutput`? But it is the output of the trigger here... It is however the input of the statistics function block at the moment.

modules/ref_fb_module/src/statistics_fb_impl.cpp
aljazfrancic: It was in the original statistics code, so I'm not sure. Can we assume that `domainPacket` is always assigned?

JakaMohorkoDS: Well, in a way we can, but it's just an extra guard against malformed inputs, so it doesn't hurt for it to be there.

modules/ref_fb_module/include/ref_fb_module/statistics_fb_impl.h
JakaMohorkoDS: It's the input port into which a signal containing on/off trigger values is connected. I understand the reasoning, to me `triggerInput` feels more natural, but both are fine.

JakaMohorkoDS (approved):

modules/ref_fb_module/src/statistics_fb_impl.cpp
```diff
@@ -52,22 +53,70 @@ void StatisticsFbImpl::initProperties()
...
 {
     std::scoped_lock lock(sync);
     readProperties();
-
     configure();
 }
 
+void StatisticsFbImpl::triggerModeChanged()
+{
+    sync.lock();
```
dejancrn, line 72: Use try/finally statements when locking manually and not using scoped_lock which uses RAII concept to make sure the resource is unlocked. This code has a potential to dead lock the system if exception occurs.

dejancrn (changes requested):

modules/ref_fb_module/src/statistics_fb_impl.cpp
aljazfrancic: I have made  what I believe are the appropriate changes.

PR #160, opened 2024-02-21
Enable additional CI jobs

JakaMohorkoDS:


JakaMohorkoDS:
VS 2017 is no longer available on GitHub runners. We need to figure out either how to install it, or add it to our self-hosted runners.

mojca:
> VS 2017 is no longer available on GitHub runners. We need to figure out either how to install it, or add it to our self-hosted runners.

... or let us all switch to VS 2022 and stop worrying about VS 2017. For how long do we need to keep supporting VS 2017?

Unsttopabull (approved):

PR #161, opened 2024-02-21
Add support for struct type in stream reader

dejancrn:


core/opendaq/reader/src/stream_reader_impl.cpp
```diff
@@ -230,6 +230,15 @@ ErrCode StreamReaderImpl::markAsInvalid()
 
 void StreamReaderImpl::inferReaderReadType(const DataDescriptorPtr& newDescriptor, std::unique_ptr<Reader>& reader) const
 {
+    if (newDescriptor.isStructDescriptor())
+    {
+        if (newDescriptor.getRawSampleSize() != newDescriptor.getSampleSize())
+            throw InvalidParameterException("Cannot use the reader on implicit or scaled structs");
```
JakaMohorkoDS, line 236: Is this the correct exception to throw here?

core/opendaq/reader/src/typed_reader.cpp
```diff
@@ -185,7 +193,10 @@ ErrCode TypedReader<ReadType>::readData(void* inputBuffer, SizeT offset, void**
         case SampleType::String:
             return readValues<SampleTypeToType<SampleType::String>::Type>(inputBuffer, offset, outputBuffer, count);
         case SampleType::Invalid:
```
JakaMohorkoDS, line 195: Do we want a Struct sample type?

JakaMohorkoDS (approved):

core/opendaq/reader/src/typed_reader.cpp
```diff
@@ -389,27 +411,40 @@ bool TypedReader<ReadType>::handleDescriptorChanged(DataDescriptorPtr& descripto
...
-    if (descriptor.assigned() && !descriptor.isStructDescriptor())
+
+    if constexpr (std::is_same_v<void*, ReadType>)
     {
-        auto postScaling = descriptor.getPostScaling();
-        if (!postScaling.assigned() || readMode == ReadMode::Scaled)
+        if (descriptor.assigned())
         {
-            dataSampleType = descriptor.getSampleType();
+            sampleSize = descriptor.getSampleSize();
```
Unsttopabull, line 419: shouldn't this be `getRawSampleSize()`?

core/opendaq/reader/src/stream_reader_impl.cpp
Unsttopabull: Maybe `InvalidState` would be better or just invalidate the reader

Unsttopabull (approved):

core/opendaq/reader/src/stream_reader_impl.cpp
dejancrn: Yes, probably invalidating the reader is the way to go. Thanks

core/opendaq/reader/src/typed_reader.cpp
dejancrn: I was thinking of that too, because I don't like that the struct descriptor is defined with InvalidSampleType and StructFields not being empty. However, I wanted to avoid updating the OPC UA model. If we can do that, I am willing to update the PR with the struct sample type.

PR #163, opened 2024-02-23
Implement native configuration remote/dummy connection and connect/di…

dejancrn:


shared/libraries/config_protocol/include/config_protocol/config_client_input_port_impl.h
```diff
@@ -53,21 +61,74 @@ inline ErrCode ConfigClientInputPortImpl::connect(ISignal* signal)
...
-            throw InvalidParameterException("Not a remote signal");
+                if (isConnected(signalPtr))
+                    return OPENDAQ_IGNORED;
+            }
 
-        StringPtr signalRemoteGlobalId;
-        checkErrorInfo(configObject->getRemoteGlobalId(&signalRemoteGlobalId));
+            const auto configObject = signalPtr.asPtrOrNull<IConfigClientObject>(true);
+            if (!configObject.assigned())
+                throw InvalidParameterException("Not a remote signal");
```
JakaMohorkoDS, line 80: We are missing a check here that the signal is actually owned by the correct remote client device. We should at least note it down as a TODO so it's not missed.

JakaMohorkoDS (approved):

shared/libraries/config_protocol/include/config_protocol/config_client_input_port_impl.h
dejancrn: added TODO comment, but such signal would anyway be rejected by the device, assuming that global ID is globally unique.

JakaMohorkoDS (approved):

PR #188, opened 2024-03-04
ToDaqObject returns nullptr instead of throwing an exception.

nilsRoettgerAtBB:
## Type of change

# Checklist:

# Description:

### DIFF
--- a/shared/libraries/opcuatms/opcuatms_client/src/objects/tms_client_property_impl.cpp
+++ b/shared/libraries/opcuatms/opcuatms_client/src/objects/tms_client_property_impl.cpp
@@ -161,19 +161,18 @@ void TmsClientPropertyImpl::configurePropertyFields()
                             // devices working which are deliviering not a default value via the opc-ua interface.
                             // Afterwards, the workaround needs to be rolled back.
 
-                            try
-                            {
-                                const auto value = reader->getValue(childNodeId, UA_ATTRIBUTEID_VALUE);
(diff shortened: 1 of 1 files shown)


shared/libraries/opcuatms/opcuatms_client/src/objects/tms_client_property_impl.cpp
```diff
@@ -161,20 +161,17 @@ void TmsClientPropertyImpl::configurePropertyFields()
...
                                 this->defaultValue = VariantConverter<IBaseObject>::ToDaqObject(value, daqContext);
                                 LOG_W(
-                                    "Failed to read default value of property {} on OpcUa client. Detault value is set to the value at connection time. {}",
-                                    this->name,
-                                    e.what());
+                                    "Failed to read default value of property {} on OpcUa client. Detault value is set to the value at connection time.",
+                                    this->name);
                             }
+                            this->defaultValue = VariantConverter<IBaseObject>::ToDaqObject(value, daqContext);
+
```
JakaMohorkoDS, line 174: This should still be wrapped in a try/catch, even if the case for isNull is handled.

nilsRoettgerAtBB: Disagree - because then also all other statements in this file need a try catch block. So, adding then here a try catch block is inconsistent in my point of view.

JakaMohorkoDS (approved):

PR #192, opened 2024-03-05
Other/optional renderer

dejancrn:


modules/ref_fb_module/src/CMakeLists.txt
```diff
@@ -23,62 +18,83 @@ set(SRC_Include common.h
...
 
-target_link_libraries(${LIB_NAME} PUBLIC daq::opendaq
-                                  PRIVATE sfml-graphics
-                                          sfml-window
-                                          date::date
-)
+target_link_libraries(${LIB_NAME} PUBLIC daq::opendaq)
+
+if (DAQMODULES_REF_FB_MODULE_ENABLE_RENDERER)
+    target_compile_definitions(${LIB_NAME} PRIVATE ENABLE_RENDERER)
```
Unsttopabull, line 93: I'd prefix definitions with `OPENDAQ_` so we're sure that it doesn't clash with anything

modules/ref_fb_module/src/ref_fb_module_impl.cpp
```diff
@@ -3,7 +3,9 @@
 #include <ref_fb_module/classifier_fb_impl.h>
 #include <ref_fb_module/power_fb_impl.h>
 #include <ref_fb_module/ref_fb_module_impl.h>
+#ifdef ENABLE_RENDERER
```
Unsttopabull, line 6: I'd prefer to use `#if defined(ENABLE_RENDERER)` as it can be extended and chained
e.g `#if defined(ENABLE_RENDERED) && defined(SFM_AVAILABLE)`

Unsttopabull (approved):

PR #194, opened 2024-03-07
Native config protocol has the wrong packet header size on the armv7l

anzeskerjancDS:
## Type of change

# Checklist:

- [X] Pull request title reflects its content
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules

# Description:

The packet header size should be the same on all platforms.

### DIFF
--- a/shared/libraries/config_protocol/include/config_protocol/config_protocol.h
+++ b/shared/libraries/config_protocol/include/config_protocol/config_protocol.h
@@ -41,14 +41,17 @@ inline auto ParamsDict(std::initializer_list<std::pair<const StringPtr, ObjectPt
 
 enum PacketType: uint8_t { getProtocolInfo = 0x80, upgradeProtocol = 0x81, rpc = 0x82, serverNotification = 0x83, invalidRequest = 0x84 };
 
+#pragma pack(push, 1)
 struct PacketHeader
 {
     uint8_t headerSize;
     PacketType type;
     char unused[2];
     uint32_t payloadSize;
-    size_t id;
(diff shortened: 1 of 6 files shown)


shared/libraries/config_protocol/include/config_protocol/config_protocol.h
```diff
@@ -41,14 +41,17 @@ inline auto ParamsDict(std::initializer_list<std::pair<const StringPtr, ObjectPt
 
 enum PacketType: uint8_t { getProtocolInfo = 0x80, upgradeProtocol = 0x81, rpc = 0x82, serverNotification = 0x83, invalidRequest = 0x84 };
 
+#pragma pack(push, 1)
```
dejancrn, line 44: force alignment on 1 byte is strictly not needed because the fields are laid in that way that they cannot be unaligned except if some system would align on non standard sizes (other than 1, 2, 4, 8)

dejancrn (approved):

shared/libraries/config_protocol/include/config_protocol/config_protocol.h
anzeskerjancDS: I added this just in case.

PR #199, opened 2024-03-10
Other/signal path optimizations

dejancrn:
Test case: RefModulesTest.DISABLED_RunDeviceScalingPerformanceTest

200 AI signals each connected to four cascaded scaler function blocks. Total 1000 signals. Sampling rate is
100 kHz. Packet size is fixed to 100 samples. In total that creates 1M (1.000.0000) packets per second. That also
means 1M scheduling tasks/second.

With the optimizations described below the CPU hit count in the profiler reached 20% of the original code.

If we come to a consensus that the approach is correct I am willing to write documentation (how to) guide on adding and
receiving packets, adding doxygen comments, and further developing the optimization concepts. 

QueryInterface optimization with static_cast
----------------------------------------------------
The templated base class produces a generic implementation of QueryInterface that uses dynamic_cast. 
dynamic_cast is slow. static_cast is fast. Would love to have a generic implementation that produces
static_cast only. Some of the classes that are on a hot signal path now have manual implementation of
QueryInterface with static_cast.

Scheduler IWork callback
----------------------------------------------------
The scheduler uses IFunction to schedule callbacks. These are costly because of unpacking the parameters.
A new type of scheduler-specific callback has been added which is more effective.

Packet reusing
----------------------------------------------------
Packets that come to function block can be reused. Note that this is different than reusing and returning packets 
at the end of the signal path. This requires different techniques not supported yet.

The prerequisite to reuse the packet when dequeued from a connection is that it has a reference count of 1. That means
the function block has the last reference to the packet. To achieve that, the producer has to "forget" the
reference to the packet when it sends it. Therefore `sendAndStealPacket` method was added to the ISignalConfig. This
method steals (adopts?) the packet. After the function is returned, the producer must not touch the packet again. This 
will make sure that the last connection (the only one if one input port is connected only) will receive the packet with
a reference count of 1. It can call reuse on the packet and do some inline algorithms on the packet buffer.

With C++ the "steal" methods should be converted to methods with rvalue packet parameters and possibly have the same name as the original method. It would be called like so:

`sendPacket(std::move(packet));`

However, this requires updating RTGen.

Our current implementation of sendPackets holds a reference to the last packet, which prevents the optimization described.
I have switched off keeping the last packet for non-public and invisible signals. 

NotifyMethod::SchedulerQueueWasEmpty
----------------------------------------------------
A new option to input port callback triggering has been added. This triggers a notification on `IConnection.enqueue` 
only if the connection queue was empty before the call. In practice, that means that the function block or listener has to
call `IConnection.dequeue` in the onPacketRecieved callback until it returns nullptr (empty queue), otherwise it will get
no more notifications until it empties the queue in some other event. Most of our function blocks behave like that. 
This can significantly reduce the number of scheduled tasks on heavily loaded systems.

Sending multiple packets
----------------------------------------------------
If producer has to send multiple packets (on a signal), it can do so more efficiently by calling `sendPackets` or
`sendAndStealPackets`. The notification will only be called for the last packet. Again this can significantly decrease the
number of scheduled jobs.

Other
----------------------------------------------------
Some unnecessary copies have been converted to moves.


JakaMohorkoDS (approved):

Unsttopabull (approved):

shared/tools/RTGen/src/project/RTGen.Python/Generators/PythonGenerator.cs
```diff
@@ -313,6 +313,17 @@ private bool isMethodTemporarilyDisabled(IMethod method)
             return false;
         }
 
+        private bool methodHasStealRef(IMethod method)
```
Daniel-BB, line 316: Please keep all method/function names CamelCase. Not so important, but that's the style used in RTGen (and by most C# coders).

PR #205, opened 2024-03-11
Prevent adding non-IPropertyObject default values

JakaMohorkoDS:
## Type of change


dejancrn (approved):
I would like to see a test case written for this :)

PR #207, opened 2024-03-12
Recursive active call + input port logic

JakaMohorkoDS:
## Type of change

# Description:

 - Connections no longer enqueue packets if input port is inactive.
 - `setActive` now recursively sets the active flag of all children

### DIFF
--- a/core/opendaq/signal/src/connection_impl.cpp
+++ b/core/opendaq/signal/src/connection_impl.cpp
@@ -18,8 +18,9 @@ ErrCode ConnectionImpl::enqueue(IPacket* packet)
 
     if (!port.getActive())
     {
-        void* intf;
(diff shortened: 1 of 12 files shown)


core/opendaq/signal/src/connection_impl.cpp
```diff
@@ -16,6 +16,13 @@ ErrCode ConnectionImpl::enqueue(IPacket* packet)
 {
     OPENDAQ_PARAM_NOT_NULL(packet);
 
+    if (!port.getActive())
+    {
+        void* intf;
+        if (OPENDAQ_FAILED(packet->borrowInterface(IEventPacket::Id, &intf)))
```
dejancrn, line 22: for packets we have IPacket.getPacketType() which should be more optimal than borrowing

dejancrn (approved):

PR #214, opened 2024-03-15
Make type manager thread safe

aljazfrancic:
## Type of change

# Checklist:

# Description:
Make type manager thread safe and some return logic changes.


core/coretypes/src/type_manager_impl.cpp
```diff
@@ -13,6 +13,8 @@ TypeManagerImpl::TypeManagerImpl()
 
 ErrCode TypeManagerImpl::addType(IType* type)
 {
+    std::scoped_lock lock(this->sync);
```
dejancrn, line 16: Do not lock what is not neccessary. The if statement does not have to be protected by mutex, scoped_lock can be moved down after checking type for nullptr.

dejancrn (commented):
Serialize method also needs to be protected by mutex

core/coretypes/src/type_manager_impl.cpp
aljazfrancic: I've moved the mutex in addType, removeType, getType, getTypes, and hasType.
I've added the mutex at the beginning of the serialize method.
Additionally I've also added a mutex in setCoreEventCallback, where I believe it also prevents a race condition.

dejancrn: Locking the core event "type added" and "type removed" can potentially lead to deadlock/mutex reentry if the code inside the event handler makes a call to the type manager. I suggest the core events are put outside of the lock.

dejancrn: Additionally, some places in the openDAQ code do this:

`if (!typeManager.hasType("typeName"))
{
    typeManager.addType(type);
}`

this is not safe/atomic as another same type can be added between `hasType` and `addType`. Code like this should be replaced just with new thread-safe implementation of `addType` can catch the exception if the element already exists and it's different.

dejancrn: Another thing not really related to this PR is that we should not allow to remove the type at any time, because multiple clients can add the same type and it's added only once. IMHO the only time it makes sense to remove the type is in the tests to avoid false positive memory leaks.

JakaMohorkoDS: > Additionally, some places in the openDAQ code do this:
> 
> `if (!typeManager.hasType("typeName")) { typeManager.addType(type); }`
> 
> this is not safe/atomic as another same type can be added between `hasType` and `addType`. Code like this should be replaced just with new thread-safe implementation of `addType` and catch the exception if the element already exists and it's different.

In general, we should not need this code after the struct `getLastValue` merge. I think that `addType` should check for equality of types and just ignore the call if the types match.



> Another thing not really related to this PR is that we should not allow to remove the type at any time, because multiple clients can add the same type and it's added only once. IMHO the only time it makes sense to remove the type is in the tests to avoid false positive memory leaks.
> 
> Not allowing to remove type also makes non-locked core triggers working correctly, because if we are allowed to remove the type and "add type" event is not protected with mutex, it could happen that event is triggered but before handler is called the type is already removed.

Not 100% sure here. We maybe want to update Types with the same name in the future, where they'd be extended with an additional field. In that case we'd need a remove (or an `update` method).

modules/ref_fb_module/src/scaling_fb_impl.cpp
```diff
@@ -251,11 +251,17 @@ void ScalingFbImpl::createSignals()
...
 {
-    if (!this->context.getTypeManager().hasType("InputStatusType"))
+    auto inputStatusType = EnumerationType("InputStatusType", List<IString>(InputDisconnected, InputConnected, InputInvalid));
+
+    try
     {
-        auto inputStatusType = EnumerationType("InputStatusType", List<IString>(InputDisconnected, InputConnected, InputInvalid));
         this->context.getTypeManager().addType(inputStatusType);
     }
+    catch (const std::exception& e)
```
JakaMohorkoDS, line 260: In a lot of these catch statements I'd format them as:
try
{}
catch(const std::exception& e)
{log with error info}
catch(...)
{log without error info}

JakaMohorkoDS (approved):
Minor comment about the catch statements, lgtm otherwise.

JakaMohorkoDS (approved):

PR #236, opened 2024-03-26
Fix nested property object begin and end update not working properly

dejancrn:


shared/libraries/config_protocol/include/config_protocol/config_server_component.h
```diff
@@ -145,4 +160,38 @@ inline BaseObjectPtr ConfigServerComponent::update(const ComponentPtr& component
...
+{
+    PropertyObjectPtr obj = component;
+    const auto pathStr = splitPath(static_cast<std::string>(path));
+    for (const auto& p : pathStr)
+    {
+        obj = obj.getPropertyValue(p);
+    }
+
+    return obj;
+}
```
JakaMohorkoDS, line 195: Does the PropertyObjectImpl dot access not work in this case? I'd think that just obj.getPropertyValue("obj1.obj2") should do exactly this.

JakaMohorkoDS (approved):

PR #237, opened 2024-03-27
Reset mirrored descriptor on unsubscribe

JakaMohorkoDS:
## Type of change

### DIFF
--- a/core/opendaq/opendaq/mocks/mock_physical_device.cpp
+++ b/core/opendaq/opendaq/mocks/mock_physical_device.cpp
@@ -26,6 +26,7 @@ inline MockPhysicalDeviceImpl::MockPhysicalDeviceImpl(const ContextPtr& ctx, con
 
     const size_t nanosecondsInSecond = 1000000000;
     auto delta = nanosecondsInSecond / 10000;
+    time = 0;
 
     auto valueDescriptor = DataDescriptorBuilder()
                                .setSampleType(SampleType::UInt64)
@@ -131,14 +132,13 @@ void MockPhysicalDeviceImpl::stopAcq()
 
 void MockPhysicalDeviceImpl::generatePackets(size_t packetCount)
 {
-    uint64_t time = 0;
     uint64_t tickDelta = 100;
 
     for (size_t i = 1; i <= packetCount; i++)
     {
         // we want tick values to be 100 % reproducable even if they do not reperesnt the exact real time
         std::this_thread::sleep_for(std::chrono::milliseconds(tickDelta));
-        time = i * tickDelta;
+        time = time + i * tickDelta;
 
         for (const auto& channel : ioFolder.getItems())
         {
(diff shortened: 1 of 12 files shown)


core/opendaq/opendaq/mocks/mock_physical_device.cpp
NikolaiShipilov, line 1: Since the meaning of generator parameters has been changed, the line:
[time = i * tickDelta;](https://github.com/openDAQ/openDAQ/blob/streaming/reset-descriptor-on-unsub/core/opendaq/opendaq/mocks/mock_physical_device.cpp#L141)
probably should be replaced by
```
time = tickDelta;
```
to prevent generating of large packets.
Similarly to how it was changed in [signal generator test](https://github.com/openDAQ/openDAQ/blob/streaming/reset-descriptor-on-unsub/shared/libraries/websocket_streaming/tests/test_signal_generator.cpp#L108)

modules/tests/test_opendaq_device_modules/test_streaming.cpp
```diff
@@ -576,36 +588,40 @@ TEST_F(NativeDeviceStreamingTest, ChangedDataDescriptorBeforeSubscribeNativeDevi
...
-    
-    clientInstance.getDevices()[0].setPropertyValue("ChangeDescriptors", 1);
-    ASSERT_EQ(callCount, sigCount);
 
-    const DataDescriptorPtr valueDataDesc = serverSignalPtr.getDescriptor();
-    const DataDescriptorPtr domainDataDesc = serverSignalPtr.getDomainSignal().getDescriptor();
+    for (int i = 1; i < 5; ++i)
+    {
+        callCount = 0;
+        clientInstance.getDevices()[0].setPropertyValue("ChangeDescriptors", i % 2);
```
NikolaiShipilov, line 595: Why is this property mechanism used rather than direct modifying of the server signal descriptor in the same manner as in the test fixture above ? Was it introduced to avoid some side effects of the NativeConfig protocol ?

NikolaiShipilov (approved):

modules/tests/test_opendaq_device_modules/test_streaming.cpp
JakaMohorkoDS: It's because the native protocol is push-based and we cannot configure the descriptors from the client side. When we change the descriptors on the server, the client is only after a while notified of the change. This simulates a change happening on the client that results in a descriptor change, with an immediate subscription afterwards.

If using the above method, we'd need to wait for the descriptor to change on the client side as well.

core/opendaq/opendaq/mocks/mock_physical_device.cpp
JakaMohorkoDS: I think it's a bit different, where `generateSamplesUntil` got its meaning changed to `generateSamplesForPeriod` or something akin to that. Basically, setting the `generateSamples` property value twice from the same test resulted in an error, as it the tick input did not increase from the previous invocation. 

I'll try and refactor a bit for the changes to make more sense.

PR #246, opened 2024-04-05
Fix spurious crashes when listeners are destroyed.

dejancrn:


JakaMohorkoDS (approved):
Looks good, but missing #include<thread> in tests.

PR #250, opened 2024-04-08
Add simple FFT function block + fix block reader

JakaMohorkoDS:
## Type of change

# Description:

- Adds a simple reference FFT implementation
- Fixes block reader `readPackets` call

### DIFF
--- a/core/opendaq/reader/src/block_reader_impl.cpp
+++ b/core/opendaq/reader/src/block_reader_impl.cpp
@@ -129,7 +129,7 @@ ErrCode BlockReaderImpl::packetReceived(IInputPort* inputPort)
 ErrCode BlockReaderImpl::readPacketData()
 {
     auto remainingSampleCount = info.dataPacket.getSampleCount() - info.prevSampleIndex;
-    SizeT toRead = std::min(remainingSampleCount, blockSize);
+    SizeT toRead = std::min(remainingSampleCount, info.remainingSamplesToRead);
 
     ErrCode errCode = valueReader->readData(getValuePacketData(info.dataPacket), info.prevSampleIndex, &info.values, toRead);
     if (OPENDAQ_FAILED(errCode))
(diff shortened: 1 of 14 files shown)


modules/ref_fb_module/include/ref_fb_module/_kiss_fft_guts.h
```diff
@@ -0,0 +1,167 @@
+/*
+ *  Copyright (c) 2003-2010, Mark Borgerding. All rights reserved.
+ *  This file is part of KISS FFT - https://github.com/mborgerding/kissfft
+ *
+ *  SPDX-License-Identifier: BSD-3-Clause
+ *  See COPYING file for more information.
+ */
```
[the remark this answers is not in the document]
Unsttopabull: Yeah this should be moved to external and pulled in via FetchContent otherwise we'll end-up with another `rapidjson` case.

modules/ref_fb_module/src/fft_fb_impl.cpp
```diff
@@ -0,0 +1,234 @@
...
+        const DataDescriptorPtr domainDesc = packet.getParameters().get(event_packet_param::DOMAIN_DATA_DESCRIPTOR);
+        processSignalDescriptorChanged(dataDesc, domainDesc);
+    }
+}
+
+void FFTFbImpl::calculate()
+{
+    std::scoped_lock lock(sync);
+    SizeT readAmount = linearReader.getAvailableCount();
+    const auto inputData = std::make_unique<float[]>(readAmount * blockSize);
```
adolenc, line 170: Please move _all_ of the memory allocations and frees out of calculate(), probably into configure(). All of these block sizes can be calculated in advance, and consequently, all the buffers can be allocated in advance as well.

I am also pretty sure you don't need memset anywhere here, since you just override the buffers with actual values.

modules/ref_fb_module/src/fft_fb_impl.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+        for (size_t blockIdx = 0; blockIdx < readAmount; blockIdx++)
+        {
+            kiss_fft_cfg cfg = kiss_fft_alloc(static_cast<int>(blockSize), 0, nullptr, nullptr);
+            const auto in = std::make_unique<kiss_fft_cpx[]>(blockSize);
+            const auto out = std::make_unique<kiss_fft_cpx[]>(blockSize);
+                        
+            for (size_t i = 0; i < blockSize; i++)
+            {
+                in[i].r = inputData[i];
```
adolenc, line 193: I am not sure where you move the inputData pointer, so you might always just be reading from the first block of data here?

modules/ref_fb_module/src/fft_fb_impl.cpp
```diff
@@ -0,0 +1,234 @@
...
+                                   .setSampleType(SampleType::Float64)
+                                   .setDimensions(dimensions)
+                                   .setUnit(inputDataDescriptor.getUnit())
+                                   .setValueRange(inputRange.assigned() ? Range(0, inputRange.getHighValue()) : Range(0, 10))
+                                   .build();
+
+        outputSignal.setDescriptor(outputDataDescriptor);
+
+        outputDomainDataDescriptor =
+            DataDescriptorBuilderCopy(inputDomainDataDescriptor).setRule(ExplicitDataRule()).setSampleType(SampleType::UInt64).build();
```
adolenc, line 145: I guess the rule could be implicit here if you set it up correctly...

modules/ref_fb_module/src/fft_fb_impl.cpp
JakaMohorkoDS: True, this is an error on my part. The block index is missing.

modules/ref_fb_module/src/fft_fb_impl.cpp
JakaMohorkoDS: True, both for the FFT input and the input data buffers we can pre-allocate. The FFT library is already a bit awkward as it has a struct type array, otherwise it could just get a memory ptr without any copies.

modules/ref_fb_module/src/fft_fb_impl.cpp
JakaMohorkoDS: Yes, but for now this is a bit simpler and the code is easier to understand. :) Can be updated in the future. In general this function block has a lot of possible optimizations.

core/opendaq/reader/src/block_reader_impl.cpp
```diff
@@ -178,43 +178,51 @@ ErrCode BlockReaderImpl::readPacketData()
...
+        if (notify.condition.wait_for(notifyLock, remainingTime, [this, &availableBlocks, maxBlocksToRead]
+        {
+             if (!notify.dataReady)
+                 return false;
+             availableBlocks = getAvailable();
+             notify.dataReady = false;
+             if (availableBlocks >= maxBlocksToRead)
+                return true;
+             return false;
+        }))
```
adolenc, line 203: I don't know if this is according to some guideline, but to me this is confusing as the {} block is a part of the lambda, not the if..true statement. I would indent the block to make it more obvious.

modules/ref_fb_module/src/fft_fb_impl.cpp
```diff
@@ -0,0 +1,262 @@
...
+    if (readAmount == 0)
+        return;
+
+    const auto outputDomainPacket = DataPacket(outputDomainDataDescriptor, readAmount);
+    const auto outputDomainData = static_cast<uint64_t*>(outputDomainPacket.getData());
+    memset(outputDomainData, 0, outputDomainPacket.getRawDataSize());
+
+    const auto outputPacket = DataPacketWithDomain(outputDomainPacket, outputDataDescriptor, readAmount);
+    const auto outputData = static_cast<double*>(outputPacket.getData());
+    memset(outputData, 0, outputPacket.getRawDataSize());
```
adolenc, line 224: Do you really need to memset?

modules/ref_fb_module/src/fft_fb_impl.cpp
```diff
@@ -0,0 +1,262 @@
...
+            DataDescriptorBuilderCopy(inputDomainDataDescriptor).setRule(ExplicitDataRule()).setSampleType(SampleType::UInt64).build();
+        outputDomainSignal.setDescriptor(outputDomainDataDescriptor);
+            
+        inputData = std::make_unique<float[]>(maxBlockReadCount * blockSize);
+        inputDomainData = std::make_unique<uint64_t[]>(maxBlockReadCount * blockSize);
+        
+        kiss_fft_free(cfg);
+        cfg = kiss_fft_alloc(static_cast<int>(blockSize), 0, nullptr, nullptr);
+        fftIn = std::make_unique<kiss_fft_cpx[]>(blockSize);
+        fftOut = std::make_unique<kiss_fft_cpx[]>(blockSize);
```
adolenc, line 163: I would just use std::vector with .resize()/.data() instead of unique_ptrs to naked arrays for the buffers.

PR #260, opened 2024-04-11
Register descriptor changes for all connected native streaming clients

NikolaiShipilov:
## Type of change

# Checklist:

# Description:
The signal descriptor gets registered/updated in the packet streaming server&client under the following circumstances:

* When an existing signal is registered in the streaming during the handling of a new client connection.
* When a newly added signal is registered in the streaming, such as when a server FB is added via a client request.
* If the initial event packet, generated during the creation of a server side reader upon signal subscription, differs from the descriptor registered in one of the two preceding cases
* When the signal descriptor changes during active signal reading by streaming server.

The bug originated from a single packet reader on the server side serving multiple subscribed clients, generating one event packet per signal change; and these event packets were only sent to subscribers, leading to the new descriptors not being registered in the packet streaming server and client of the native streaming clients that were not subscribed to the signal. 

For the FB's signal specifically, the first event packet contains `nullptr` descriptors since no signal is connected to the FB port initially. These `nullptr` descriptors are ignored by the packet streaming server&client, resulting in no descriptor being registered for the signal.

With the introduced fix, the event packets are now sent to all connected clients.

### DIFF
--- a/shared/libraries/native_streaming_protocol/src/native_streaming_server_handler.cpp
+++ b/shared/libraries/native_streaming_protocol/src/native_streaming_server_handler.cpp
@@ -167,12 +167,24 @@ bool NativeStreamingServerHandler::handleSignalSubscription(const SignalNumericI
 void NativeStreamingServerHandler::sendPacket(const SignalPtr& signal, const PacketPtr& packet)
 {
     auto signalNumericId = findSignalNumericId(signal);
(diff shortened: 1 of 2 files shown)


NikolaiShipilov:
Bug was fixed by https://github.com/openDAQ/openDAQ/pull/262

PR #292, opened 2024-04-26
Clone property object to create default config from type

NikolaiShipilov:
## Type of change

# Checklist:

# Description:


JakaMohorkoDS (approved):
With the discovery rework + streaming/config separation we need to figure out how to handle these default configs.

When we connect via "daq://", I don't think we would ever want users to provide a config object. It's supposed to be the simplest way of connecting, where you don't really care how you connect to a device.

As for when connecting using a specific protocol - we now have a new issue where we have separate config params when connecting streaming or configuration. Would we use nested objects to denote streaming connection parameters?

dejancrn (approved):

PR #299, opened 2024-05-02
NodeId as String not as numeric for Properties

nilsRoettgerAtBB:
In opendaq only objects had a nodeID as string. With this PR also Properties / Variables have a numeric node id.

## Type of change

# Checklist:

# Description:

### DIFF
--- a/shared/libraries/opcua/opcuaserver/src/opcuaaddnodeparams.cpp
+++ b/shared/libraries/opcua/opcuaserver/src/opcuaaddnodeparams.cpp
@@ -2,6 +2,19 @@
 
 BEGIN_NAMESPACE_OPENDAQ_OPCUA
 
+/*RequestedNodeIdBaseOnName*/
+
+static daq::opcua::OpcUaNodeId RequestedNodeIdBaseOnName(const std::string& name, const OpcUaNodeId& parentNodeId)
+{
+    if (parentNodeId.getValue().identifierType == UA_NODEIDTYPE_STRING)
+    {
+        std::string newNodeIdStr = utils::ToStdString(parentNodeId.getValue().identifier.string) + "/" + name;
+        return OpcUaNodeId(parentNodeId.getNamespaceIndex(), newNodeIdStr);
+    }
+    return UA_NODEID_NULL;
+}
+
+
 /*AddNodeParams*/
 
 AddNodeParams::AddNodeParams(const OpcUaNodeId& requestedNewNodeId, const OpcUaNodeId& parentNodeId, const OpcUaNodeId& referenceTypeId)
@@ -41,6 +54,12 @@ AddObjectNodeParams::AddObjectNodeParams(const OpcUaNodeId& requestedNewNodeId, 
 {
 }
 
+AddVariableNodeParams::AddVariableNodeParams(const std::string& name, const OpcUaNodeId& parentNodeId)
+    : GenericAddNodeParams<UA_VariableAttributes>(
+                  RequestedNodeIdBaseOnName(name, parentNodeId), parentNodeId, OpcUaNodeId(UA_NS0ID_HASPROPERTY), UA_VariableAttributes_default)
+{
+}
+
 /*AddVariableNodeParams*/
 
 AddVariableNodeParams::AddVariableNodeParams(const OpcUaNodeId& requestedNewNodeId)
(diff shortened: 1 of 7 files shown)


nilsRoettgerAtBB:
@JakaMohorkoDS : I assume this is not a breaking change because openDAQ do not care if this a string nodeIds or numeric. With having string node ids, a device ensures that it a specific node has always the same id. This is nice feature and was requested by HBK. The nodeId reflect the path. With this change objects and properties have also the same logic. The id reflects the hierarchical path. 
What I find a pity is that properties don't have a globalId. Because of this I need to implement it like this, and I could not do it as it is done for Objects.

JakaMohorkoDS:
> @JakaMohorkoDS : I assume this is not a breaking change because openDAQ do not care if this a string nodeIds or numeric. With having string node ids, a device ensures that it a specific node has always the same id. This is nice feature and was requested by HBK. The nodeId reflect the path. With this change objects and properties have also the same logic. The id reflects the hierarchical path. What I find a pity is that properties don't have a globalId. Because of this I need to implement it like this, and I could not do it as it is done for Objects.

Technically, we could have something similar to global IDs for properties and it's already available in the core.

The path to a property is denoted as follows: "Component1/Component2.Prop1.Prop2" Where you can have levels of nesting properties, if their parent is a property object, not a component. We'd need  to expose the path variable via an internal interface, but that's not an issue.

shared/libraries/opcua/opcuaserver/src/opcuaaddnodeparams.cpp
```diff
@@ -2,6 +2,19 @@
...
 BEGIN_NAMESPACE_OPENDAQ_OPCUA
 
+/*RequestedNodeIdBaseOnName*/
+
+static daq::opcua::OpcUaNodeId RequestedNodeIdBaseOnName(const std::string& name, const OpcUaNodeId& parentNodeId)
+{
+    if (parentNodeId.getValue().identifierType == UA_NODEIDTYPE_STRING)
+    {
+        std::string newNodeIdStr = utils::ToStdString(parentNodeId.getValue().identifier.string) + "/" + name;
+        return OpcUaNodeId(parentNodeId.getNamespaceIndex(), newNodeIdStr);
```
JakaMohorkoDS, line 12: I'm not entirely sure that the child node ID is always the same as the parent's here.

shared/libraries/opcua/opcuaserver/src/opcuaserver.cpp
```diff
@@ -598,6 +599,34 @@ void OpcUaServer::closeSession(UA_Server* server, UA_AccessControl* ac, const UA
...
+            UA_QualifiedName_delete(browseName);
+        }
+
+        parentNodeIdStr.append("/");
+        newNodeIdStr = parentNodeIdStr + objectTypeIdStr;
+        *targetNodeId = UA_NODEID_STRING_ALLOC(targetParentNodeId->namespaceIndex, const_cast<char*>(newNodeIdStr.c_str()));
+    }
+
+    return UA_STATUSCODE_GOOD;
+}
```
JakaMohorkoDS, line 627:
```suggestion
UA_StatusCode OpcUaServer::generateChildId(UA_Server* server,
                                           const UA_NodeId*/*sessionId*/,
                                           void*/*sessionContext*/,
                                           const UA_NodeId* sourceNodeId,
                                           const UA_NodeId* targetParentNodeId,
                                           const UA_NodeId*/*referenceTypeId*/,
                                           UA_NodeId* targetNodeId)
{
    if (targetParentNodeId->identifierType == UA_NODEIDTYPE_STRING) {
        const std::string parentNodeIdStr = utils::ToStdString(targetParentNodeId->identifier.string);
        std::string objectTypeIdStr;
		
        if (sourceNodeId->identifierType == UA_NODEIDTYPE_STRING)
        {
            objectTypeIdStr = utils::ToStdString(sourceNodeId->identifier.string);
        }
        else if (sourceNodeId->identifierType == UA_NODEIDTYPE_NUMERIC)
        {
            UA_QualifiedName* browseName = UA_QualifiedName_new();
            UA_Server_readBrowseName(server, *sourceNodeId, browseName);
            objectTypeIdStr = utils::ToStdString(browseName->name);
            UA_QualifiedName_delete(browseName);
        }
        else
        {
            return UA_STATUSCODE_GOOD;
        }

        const auto newNodeIdStr = parentNodeIdStr + "/" + objectTypeIdStr;
        *targetNodeId = UA_NODEID_STRING_ALLOC(targetParentNodeId->namespaceIndex, newNodeIdStr.c_str());
    }

    return UA_STATUSCODE_GOOD;
}
```

shared/libraries/opcua/opcuaserver/src/opcuaaddnodeparams.cpp
nilsRoettgerAtBB: I thought through it again, and I am 99,9% sure ;-). 
Because: the path is always built up as: <parent_id>/<childName>
Since openDAQ object have already every time a string nodeId, like IO, etc. we can use this string path + the child's name. 
If you have nested properties, also a property has a string id first, because it belongs to one of our objects. 

The only difference I see, based on your feedback from the other comment that a nodeID would look like this than:

Component1/Component2/Prop1/Prop2

and not like this:

Component1/Component2.Prop1.Prop2 as it is in the core.

JakaMohorkoDS: Ah, I might've worded my comment a bit oddly. I meant the ID due to the namespace index. We should probably use the namespace 0 index here?

JakaMohorkoDS (approved):

shared/libraries/opcua/opcuaserver/src/opcuaaddnodeparams.cpp
nilsRoettgerAtBB: 0 index is normaly reserved for the OPC-UA standard. See:

Namespace 0 is defined by the OPC Foundation and contains OPC Core nodes that you cannot modify.

Namespace 1 is specific to your server can contains server info like certificates and open connections.

Namespace 2 ... n is where your business objects are.

And all our objects are in Namespace 2 and greater so it is right...

PR #301, opened 2024-05-06
Enable support of any streaming protocol type for native-config device

NikolaiShipilov:
## Type of change

# Checklist:

# Description:

### DIFF
--- a/core/opendaq/modulemanager/src/module_manager_impl.cpp
+++ b/core/opendaq/modulemanager/src/module_manager_impl.cpp
@@ -341,9 +341,8 @@ ErrCode ModuleManagerImpl::createDevice(IDevice** device, IString* connectionStr
         return this->makeErrorInfo(OPENDAQ_ERR_ARGUMENT_NULL, "Connection string is not set or empty");
 
     bool useSmartConnection = (connectionStringPtr.toStdString().find("daq://") == 0);
-    auto configPtr = PropertyObjectPtr::Borrow(config);
 
-    DeviceInfoPtr deviceInfo;
+    DeviceInfoPtr discoveredDeviceInfo;
     
     if (useSmartConnection)
     {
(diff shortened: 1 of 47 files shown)


core/opendaq/modulemanager/src/module_manager_impl.cpp
```diff
@@ -687,53 +697,57 @@ void ModuleManagerImpl::attachStreamingsToDevice(const MirroredDeviceConfigPtr&
...
-                    MirroredSignalConfigPtr mirroredSignalConfigPtr = signal.template asPtr<IMirroredSignalConfig>();
-                    if (!mirroredSignalConfigPtr.getActiveStreamingSource().assigned() || overrideActiveSourceForSignal)
-                        mirroredSignalConfigPtr.setActiveStreamingSource(connectionString);
-                }
-            }
+            if (!signal.getPublic())
+                continue;
+            MirroredSignalConfigPtr mirroredSignalConfigPtr = signal.template asPtr<IMirroredSignalConfig>();
+            if (!mirroredSignalConfigPtr.getActiveStreamingSource().assigned() || overrideActiveSourceForSignal)
+                mirroredSignalConfigPtr.setActiveStreamingSource(connectionString);
```
JakaMohorkoDS, line 750: Does this mean another source might be active before the highest priority one? It might make more sense to first connect the highest priority source in general. If it's bundled with the configuration connection, I'd assume it's anyhow the highest priority one.

As for native, it should also be possible to connect config-only, without streaming. And if that's possible, I'd assume that all streamings are "detached" anyhow.

NikolaiShipilov: > Does this mean another source might be active before the highest priority one? 

When the `ModuleManagerImpl::attachStreamingsToDevice` is called for some device and either the "MIN_HOPS" or "FALLBACKS"  heuristic is selected, certain signals may already have an active streaming source assigned  - the highest priority source of the nested device. If the "FALLBACKS" heuristic is selected, this assignment gets overwritten by the highest priority source of the parent device. However, it doesn't really make sense to provide the "FALLBACKS" heuristic until we define and implement a procedure for automatically reconfiguring streaming in cases where the previously selected active streaming source disconnects or fails. Therefore, for now, I would limit the possible heuristics to these three: "MIN_HOPS," "MIN_CONNECTIONS," and "DISCONNECTED."

> As for native, it should also be possible to connect config-only, without streaming. And if that's possible, I'd assume that all streamings are "detached" anyhow.

Regarding the native streaming for native config devices - it's automatically connected by the client module during the config connection step to utilize the same pipe for both streaming and configuration. However, to ensure proper handling of streaming properties in the module manager, previously established streaming objects associated with the device are detached (removed) and re-added only if the corresponding protocol is enabled by the streaming properties. To streamline the implementation, we could remove the automatic creation of native streaming from the module implementation. Rather than being automatically created along with the config connection, the native streaming object for native config devices would be instantiated by the module manager, just like any other type of streaming. It would then operate in a separate pipe. This approach removes the necessity for custom handling of this source in the module manager, such as detaching it from the device and re-adding it.

dejancrn (approved):

JakaMohorkoDS (approved):

PR #312, opened 2024-05-16
Send invalid sample type descriptor instead of nullptr if signal has …

dejancrn:
When an input port listener receives the data descriptor change event is unable to detect if the descriptor is null or it did not change. This is an attempt to solve the issue by sending "null descriptor" when the signal has no descriptor. If the descriptor did not change, the param field is nullptr.

Null descriptor is defined as descriptor with sample type = invalid.

Would it make sense to introduce SampleType::Null?

### DIFF
--- a/shared/libraries/websocket_streaming/tests/streaming_test_helpers.h
+++ b/shared/libraries/websocket_streaming/tests/streaming_test_helpers.h
@@ -31,6 +31,15 @@
 #include <coreobjects/authentication_provider_factory.h>
 #include "streaming_protocol/Unit.hpp"
 
+// MAC CI issue
+#if !defined(SKIP_TEST_MAC_CI)
+#   if defined(__clang__)
+#       define SKIP_TEST_MAC_CI return
+#   else
+#       define SKIP_TEST_MAC_CI
+#   endif
+#endif
+
 namespace streaming_test_helpers
 {
     inline daq::InstancePtr createServerInstance()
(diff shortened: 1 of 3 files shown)


NikolaiShipilov:
From the streaming-LT perspective, the SampleType is used as a template parameter for output signal objects. Whether the descriptor for the signal is nullptr or has a SampleType of "Invalid", the output signal object will not be created in either case.

However, this means that when the SampleType for a signal changes, streaming-LT requires special handling for these changes. Implementation-wise, the old-type output signal object should be destroyed, and a new-type object should be created. Protocol-wise, the changed signal should probably be reported as unavailable first, and then as available again, to trigger a client-side update of the signal parameters.

As for native streaming, it seems that this NullType/InvalidType descriptor improvement should work as is.

shared/libraries/packet_streaming/src/packet_streaming_client.cpp
```diff
@@ -74,23 +58,37 @@ void PacketStreamingClient::addEventPacketBuffer(const PacketBufferPtr& packetBu
...
     {
-        // drop packet if signal descriptors have not been changed
-        if (BaseObjectPtr::Equals(packet, getDataDescriptorChangedEventPacket(signalId)))
-            forwardPacket = false;
+        const auto valueDescriptorParam = packet.getParameters().get(event_packet_param::DATA_DESCRIPTOR);
+        const auto domainDescriptorParam = packet.getParameters().get(event_packet_param::DOMAIN_DATA_DESCRIPTOR);
+        const bool valueDescriptorChanged = valueDescriptorParam.assigned();
+        const bool domainDescriptorChanged = domainDescriptorParam.assigned();
+        const DataDescriptorPtr newValueDescriptor =
+            valueDescriptorParam != NullDataDescriptor() ? valueDescriptorParam : nullptr;
```
dejancrn, line 66: This pattern is used very often so I would suggest having a standard function in openDAQ, perhaps called `validateDescriptor` or something like that.

PR #316, opened 2024-05-20
Fix native protocol inactivity timeout handling

NikolaiShipilov:
## Type of change

### DIFF
--- a/shared/libraries/native_streaming_protocol/src/base_session_handler.cpp
+++ b/shared/libraries/native_streaming_protocol/src/base_session_handler.cpp
@@ -34,12 +34,15 @@ void BaseSessionHandler::startConnectionActivityMonitoring(Int heartbeatPeriod, 
     {
         connectionInactivityTimer->cancel();
         connectionInactivityTimer->expires_from_now(inactivityTimeoutMs);
+
+        std::weak_ptr<Session> session_weak = session;
         connectionInactivityTimer->async_wait(
-            [this](const boost::system::error_code& ec)
+            [errorHandler = errorHandler, session_weak](const boost::system::error_code& ec)
             {
                 if (ec)
                     return;
-                this->errorHandler("Connection activity timeout error", session);
+                if (auto session = session_weak.lock())
+                    errorHandler("Connection activity timeout error", session);
             }
         );
     };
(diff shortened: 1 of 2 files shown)


JakaMohorkoDS (approved):

shared/libraries/native_streaming_protocol/src/base_session_handler.cpp
```diff
@@ -35,11 +35,11 @@ void BaseSessionHandler::startConnectionActivityMonitoring(Int heartbeatPeriod,
         connectionInactivityTimer->cancel();
         connectionInactivityTimer->expires_from_now(inactivityTimeoutMs);
         connectionInactivityTimer->async_wait(
-            [this](const boost::system::error_code& ec)
+            [errorHandler = errorHandler, session = session](const boost::system::error_code& ec)
             {
                 if (ec)
```
dejancrn, line 40: What does this mean? If operation failed we return and do nothing?

shared/libraries/native_streaming_protocol/src/native_streaming_client_handler.cpp
```diff
@@ -236,20 +236,9 @@ void NativeStreamingClientHandler::initClientSessionHandler(SessionPtr session)
...
-                    connectionStatusChanged(ClientConnectionStatus::Reconnecting);
-                    tryReconnect();
-                });
-        }
-        else
-        {
-            connectionStatusChanged(ClientConnectionStatus::Reconnecting);
-            tryReconnect();
-        }
+
```
dejancrn, line 239: No need to close the session?

dejancrn (commented):
Posted some questions just for my understaning.

shared/libraries/native_streaming_protocol/src/base_session_handler.cpp
NikolaiShipilov: We do smth (here triggering connection closing) when timer expires. If timer callback called with some error code (for example timer operation cancelled on shutdown)  - yes, we return do nothing.

shared/libraries/native_streaming_protocol/src/native_streaming_client_handler.cpp
NikolaiShipilov: No need to do this explicitly, as the connection will be closed automatically when the "Session" object is destroyed.

dejancrn: Ok

shared/libraries/native_streaming_protocol/src/base_session_handler.cpp
dejancrn: Ok

dejancrn (approved):

PR #324, opened 2024-05-24
Protocol level regression tests

aljazfrancic:
## Type of change

# Checklist:

# Description:

* Adds regression tests - protocol level for
    * OPC UA level compatibility
    * Native configuration compatibility
    * Native streaming compatibility
    * Streaming LT level compatibility


regression/tests/regression_test_input_port.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+        device = instance.addDevice(connectionString);
+
+        fb = device.getFunctionBlocks()[0];
+
+        port = fb.getInputPorts()[0];
+    }
+};
+
+TEST_F(RegressionTestInputPort, acceptsSignalConnectGetSignalGetConnectionDisconnect)
```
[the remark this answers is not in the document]
aljazfrancic: Will introduce minimum version for tests later (when we are testing against more than one version). I already know how to do it (via the environment variable `regression_tests_target_tag` - from `ci.yml` - which can later contain multiple targets) and then check against that in the tests.

regression/tests/regression_test_device.cpp
```diff
@@ -0,0 +1,225 @@
...
+{
+    ListPtr<IDevice> devices;
+    ASSERT_NO_THROW(devices = device.getDevices());
+    ASSERT_EQ(devices.getCount(), 0);
+}
+
+TEST_F(RegressionTestDevice, getAvailableDevices)
+{
+    ListPtr<IDeviceInfo> infos;
+    ASSERT_NO_THROW(infos = device.getAvailableDevices());
```
denise-opendaq, line 163: we can check discovering by enabling mdns in regression simultor generator. the example of enabling is in [quick_start_simulator](https://github.com/openDAQ/openDAQ/blob/main/examples/cpp/quick_start/quick_start_simulator.cpp)

aljazfrancic: Thank you for your comment, but I don't think this is possible in the version we are currently targeting with the regression tests (the version of the SDK that the regression simulator runs on), which is [tag v3.0.0](https://github.com/openDAQ/openDAQ/tree/v3.0.0). I think the best course of action is to test discovery on a later version. There are already plans to target different versions of the SDK with regression tests anyways, and this is on the top of the TODO list for further expansion of the regression test framework for sure.

euttsov-odaq (approved):

PR #325, opened 2024-05-24
<TBBAS-1468> Integrate streaming-LT range and post scaling metadata

NikolaiShipilov:
* encode/decode signal value range and linear scaling parameters within the LT "definition" metadata
* keep compatibility with older client/server versions by also using "interpretation" metadata to transmit range and scaling parameters

## Type of change

# Checklist:

### DIFF
--- a/shared/libraries/websocket_streaming/src/signal_descriptor_converter.cpp
+++ b/shared/libraries/websocket_streaming/src/signal_descriptor_converter.cpp
@@ -28,22 +28,23 @@ BEGIN_NAMESPACE_OPENDAQ_WEBSOCKET_STREAMING
 SubscribedSignalInfo SignalDescriptorConverter::ToDataDescriptor(const daq::streaming_protocol::SubscribedSignal& subscribedSignal)
 {
     SubscribedSignalInfo sInfo;
+    auto dataDescriptorBuilder = DataDescriptorBuilder();
 
     // *** meta "definition" start ***
     // get metainfo received via signal "definition" object and stored in SubscribedSignal struct
-    auto dataDescriptor = DataDescriptorBuilder().setRule(GetRule(subscribedSignal));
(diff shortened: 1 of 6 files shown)


shared/libraries/websocket_streaming/src/signal_descriptor_converter.cpp
```diff
@@ -54,30 +55,51 @@ SubscribedSignalInfo SignalDescriptorConverter::ToDataDescriptor(const daq::stre
...
+    if (streamingRange.isUnlimited())
+    {
+        // An unlimited range indicates that it is not set within the "definition" object.
+        // Since the range is used to configure the renderer, we need to set a default value for fusion
+        // (non-openDAQ) device signals.
+        // If the signal is owned by an openDAQ-enabled server, the value range, if present,
+        // is specified within the "definition" object or alternatively might be set within
+        // the "interpretation" object, if so that value will override the default one.
+        if (!subscribedSignal.isTimeSignal())
+            dataDescriptorBuilder.setValueRange(CreateDefaultRange(dataDescriptorBuilder.getSampleType()));
```
JakaMohorkoDS, line 73: Is "range" mandatory? I'd find it an issue of the renderer if it only accepts signals with a specified range.

JakaMohorkoDS (approved):

shared/libraries/websocket_streaming/src/signal_descriptor_converter.cpp
NikolaiShipilov: If a range is not specified for a signal, the renderer uses [default rendering limits of (0, 1)](https://github.com/openDAQ/openDAQ/blob/main/modules/ref_fb_module/src/renderer_fb_impl.cpp#L1355). Other FBs also apply their own default range if the input signal does not specify one. It might be beneficial to unify these default min/max values across various FBs.

PR #326, opened 2024-05-29
Other/config object rework

JakaMohorkoDS:
Description:
- Adds a default config that is used to configure both device (config) connection, streaming, and general settings at once
- Add StreamingType
- Add new general add-device-config settings


core/opendaq/modulemanager/src/module_manager_impl.cpp
```diff
@@ -754,37 +755,20 @@ uint16_t ModuleManagerImpl::getServerCapabilityPriority(const ServerCapabilityPt
...
+    if (!config.assigned())
+        return defConfig;
 
-    if (!resultConfig.hasProperty("AutomaticallyConnectStreaming"))
+    for (const auto& prop : defConfig.getAllProperties())
     {
-        resultConfig.addProperty(BoolProperty("AutomaticallyConnectStreaming", true));
+        const auto name = prop.getName();
+        if (config.hasProperty(name))
+            defConfig.setPropertyValue(name, config.getPropertyValue(name));
```
denise-opendaq, line 768: can we trigger the exception is the type is mismatched? or we will get just ignore err code?

modules/opcua_client_module/src/opcua_client_module_impl.cpp
```diff
@@ -163,14 +163,21 @@ StringPtr OpcUaClientModule::formConnectionString(const StringPtr& connectionStr
...
+        ipv6 = false;
         parsed = std::regex_search(urlString, match, regexIpv4Hostname);
+    }
 
     if (parsed)
     {
         prefix = match[1];
         host = match[2];
+        if (ipv6)
+            host = "[" + host + "]";
```
denise-opendaq, line 179: we just can replace ipv6 regexp with (so it will include brackets in host)
```(^(.*://)(\[[a-fA-F0-9:]+\])(?::(\d+))?(/.*)?$)```

File: modules/websocket_streaming_client_module/src/websocket_streaming_client_module_impl.cpp [hunk 3]
```diff
@@ -231,19 +231,25 @@ StringPtr WebsocketStreamingClientModule::formConnectionString(const StringPtr&
...
+        ipv6 = false;
         parsed = std::regex_search(urlString, match, regexIpv4Hostname);
+    }
 
     if (parsed)
     {
         prefix = match[1];
         host = match[2];
+        if (ipv6)
+            host = "[" + host + "]";
```
denise-opendaq, line 247: the same as in opcaua regexp

shared/libraries/config_protocol/include/config_protocol/config_client_device_impl.h
```diff
@@ -123,6 +129,35 @@ uint64_t GenericConfigClientDeviceImpl<TDeviceBase>::onGetTicksSinceOrigin()
...
+template <class TDeviceBase>
+DictPtr<IString, IDeviceType> GenericConfigClientDeviceImpl<TDeviceBase>::onGetAvailableDeviceTypes()
+{
+    return Dict<IString, IDeviceType>();
+}
+
+template <class TDeviceBase>
+DevicePtr GenericConfigClientDeviceImpl<TDeviceBase>::onAddDevice(const StringPtr& /*connectionString*/, const PropertyObjectPtr& /*config*/)
+{
+    throw NotImplementedException{};
```
NikolaiShipilov, line 147: Just to clarify: does this mean that adding a remote device in the same way as adding a function block via the native config protocol is completely disallowed, or is this just a temporary limitation due to the current lack of implementation?

core/opendaq/modulemanager/src/module_manager_impl.cpp
```diff
@@ -617,6 +665,75 @@ ErrCode ModuleManagerImpl::createStreaming(IStreaming** streaming, IString* conn
...
+        {
+            types = module.getAvailableStreamingTypes();
+        }
+        catch (const NotImplementedException&)
+        {
+            LOG_I("{}: GetAvailableFunctionBlockTypes not implemented", module.getName())
+        }
+        catch ([[maybe_unused]] const std::exception& e)
+        {
+            LOG_W("{}: GetAvailableFunctionBlockTypes failed: {}", module.getName(), e.what())
```
NikolaiShipilov, line 689: "GetAvailableFunctionBlockTypes" -> "GetAvailableStreamingTypes"

modules/tests/test_opendaq_device_modules/test_default_config.cpp
```diff
@@ -0,0 +1,240 @@
...
+}
+
+TEST_F(ModulesDefaultConfigTest, GeneralConfig)
+{
+    const auto instance = Instance();
+    const auto config = instance.createDefaultAddDeviceConfig();
+    const PropertyObjectPtr general = config.getPropertyValue("general");
+    ASSERT_TRUE(general.hasProperty("PrioritizedStreamingProtocols"));
+    ASSERT_TRUE(general.hasProperty("StreamingConnectionHeuristic"));
+}
```
NikolaiShipilov, line 23: The `"AllowedStreamingProtocols"` and `"AutomaticallyConnectStreaming"` properties should likely be tested here as well.

NikolaiShipilov (approved):
Could you please also update the code fragments in several device modules tests, identified by the comments `"// FIXME - use default config mega-object"`, to align with the new config mechanism.

shared/libraries/config_protocol/include/config_protocol/config_client_device_impl.h
JakaMohorkoDS: This is a temporary limitation. It should work and invoke an RPC, but NotImplemented better represents the current state than some arbitrary default behaviour.

PR #334, opened 2024-06-10
Maintain MSVC binary compatibility with older runtime environments

JakaMohorkoDS:
## Type of change

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,9 +86,7 @@ jobs:
 
       - name: Install ninja-build
         if: matrix.cmake_generator == 'Ninja'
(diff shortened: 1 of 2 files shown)


aljazfrancic (approved):

.github/workflows/ci.yml
```diff
@@ -250,10 +255,10 @@ jobs:
         run: |
           apt-get update
           apt-get install -y git openssh-client
-          
+
```
mojca, line 258: I didn't notice that I also committed these. Whitespace changes should ideally be separate out (into its own commit) from the rest of  the functional changes.

JakaMohorkoDS: Think it's okay as it is.

PR #343, opened 2024-06-19
Antora documentation: native protocols specifications

NikolaiShipilov:
## Type of change

# Description:

Adds an Antora versions of native protocols specs:
* [Packet streaming protocol](https://blueberrydaq.atlassian.net/wiki/spaces/~640b55549cba7ca0287bc6e0/pages/551288854/Packet+streaming+protocol+specification+draft)
* [Native communication protocol](https://blueberrydaq.atlassian.net/wiki/spaces/~640b55549cba7ca0287bc6e0/pages/544964610/openDAQ+native+communication+protocol+specification+draft)

The specifications are created separately from the user guide

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -606,14 +606,23 @@ jobs:
           npm i @asciidoctor/tabs
           npm i @springio/antora-extensions
 
-      - name: Run Antora
+      - name: Run Antora to build user guide
         run: >
           [ "refs/heads/main" != "${{ github.ref }}" ] && npx antora antora-playbook.yml
           || (export DAQ_DEV_WEBSITE=docs-dev; npx antora antora-playbook-dev.yml)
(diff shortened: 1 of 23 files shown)


JakaMohorkoDS (approved):
In general, this looks good to me. 

One thing that I think is missing, is examples of what packets are sent by the server and client when establishing a connection and in what order these are sent. This does not need to include the full payloads, just examples of what happens.

Here we have 3 cases:
- Establish a streaming-only connection
- Establish a configuration connection
- Establish a configuration connection and then a streaming connection

This should be formatted something like:

1. Client->Server: Initialization request (Packet ID#123)
2. Server->Client: Ack (Packet ID#234)
3. Client->Server: Transport layer properties (Packet ID#345)
4. ...

JakaMohorkoDS (approved):

PR #348, opened 2024-06-21
Modify discovery capability merge

JakaMohorkoDS:
- Take device caps as source of truth
- Add port to discovery caps
- Fall back to default port if not specified in cap

### DIFF
--- a/core/opendaq/modulemanager/src/module_manager_impl.cpp
+++ b/core/opendaq/modulemanager/src/module_manager_impl.cpp
@@ -532,7 +532,7 @@ ErrCode ModuleManagerImpl::createDevice(IDevice** device, IString* connectionStr
                 for (const auto& capability : devicePtr.getInfo().getServerCapabilities())
                 {
                     // assigns missing connection strings for server capabilities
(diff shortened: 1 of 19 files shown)


core/opendaq/modulemanager/src/module_manager_impl.cpp
```diff
@@ -488,27 +490,43 @@ ErrCode ModuleManagerImpl::createDevice(IDevice** device, IString* connectionStr
...
+                        connectedDeviceInfoInternal.addServerCapability(capPtr);
                     }
                 }
 
                 for (const auto& capability : devicePtr.getInfo().getServerCapabilities())
                 {
                     // assigns missing connection strings for server capabilities
-                    if (capability.getConnectionString() == "" &&
-                        capability.getConnectionStrings().empty())
+                    if (capability.getConnectionString() == "" && capability.getConnectionStrings().empty())
```
denis-erokhin, line 529: i think we can live just 
```
if (capability.getConnectionStrings().empty())
```

denis-erokhin (approved):

PR #355, opened 2024-06-27
Case standardization (PascalCase)

aljazfrancic:
## Type of change

# Checklist:

# Description:
Change many things to PascalCase


7AXEL:
Check out this project: https://github.com/7AXEL/Web2APK

aljazfrancic:
Relevant backwards compatibility commits are:
* ec0b22cd86cfddf8a6c095044e34b70ca0d5e9f8
* 687137bd9c3f3fd2ca7ab140806d30bc0dc79ea1

JakaMohorkoDS:
@Unsttopabull @dejancrn I added you two as reviewers to double check on the Integration side on whether all of these changes are okay or not.

core/coretypes/src/ratio_impl.cpp
```diff
@@ -10,10 +10,10 @@ BEGIN_NAMESPACE_OPENDAQ
 
 RatioImpl::RatioImpl(Int numerator, Int denominator)
     : GenericStructImpl<daq::IRatio, daq::IStruct, daq::IConvertible, daq::IComparable>(
-          RatioStructType(), Dict<IString, IInteger>({{"numerator", numerator}, {"denominator", denominator}}))
+          RatioStructType(), Dict<IString, IInteger>({{"Numerator", numerator}, {"Denominator", denominator}}))
```
JakaMohorkoDS, line 13: It'd be good to have a check for everything that has a struct type whether deserialization with old struct types works. Since here we're basically redefining all struct types to be PascalCase instead of camelCase.

I'd just include a test suite that has a set of serialized object strings of structs as they were before the change, then the test should try to deserialize them with the new IDs.

modules/ref_device_module/src/ref_device_impl.cpp
```diff
@@ -121,8 +121,8 @@ void RefDeviceImpl::initClock()
 
 void RefDeviceImpl::initIoFolder()
 {
-    aiFolder = this->addIoFolder("ai", ioFolder);
-    canFolder = this->addIoFolder("can", ioFolder);
```
JakaMohorkoDS, line 125: I suppose these are okay for now as these are reference modules. But in general we need to avoid any changes to the local IDs of our components, as we'll break every stored setup doing so.

I don't think I noticed any local Id changes throughout the PR except in the reference modules, but if there are any, they should be kept as-is.

modules/native_streaming_client_module/src/native_streaming_client_module_impl.cpp
```diff
@@ -575,7 +575,7 @@ StringPtr NativeStreamingClientModule::onCreateConnectionString(const ServerCapa
     }
 
     return NativeStreamingClientModule::createUrlConnectionString(
-        serverCapability.getProtocolId() == "opendaq_native_streaming" ? NativeStreamingPrefix : NativeConfigurationDevicePrefix,
```
JakaMohorkoDS, line 578: We need to be really careful on these since I'm pretty sure these break backwards compatibility. Anywhere on the client side where we check for protocol ID we need to make sure that we accept either the new or the old one.

So this concerns mainly:
- OPC UA client
- Native client
- Streaming LT client

Basically all of these should be checked for backwards compatibility. The easiest way to check is to start a simulator and connect to it *without* running `getAvailableDevices` before connecting, as this will use the capabilities of the device, not those created through discovery.

core/coretypes/src/ratio_impl.cpp
aljazfrancic: I believe this shouldn't be a problem. For example, a Ratio may serialize into `{"__type":"Ratio","num":1,"den":2}`, which is the same as before, because `SerializeId()` is the same (`Ratio`), and so are the keys (`num` and `den`). Am I wrong?

JakaMohorkoDS: It shouldn't be an issue as long as these StructTypes are not serialized. The only result will likely be that there will be a duplicate entry in the TypeManager of these struct types on the native client side - one entry will be camelCase the other PascalCase. It shouldn't cause issues, however.

JakaMohorkoDS (approved):

changelog/changelog_3.0.0-4.0.0.txt
```diff
@@ -1,11 +1,45 @@
...
+			- Reference modules
+			- Streaming/config clients
+			- Server modules
+		- Server capability protocol ID (backwards compatible)
+		- Server capability protocol name
+		- Struct Type field names
+		- Other
+
+Required integration changes:
+	- Generally none, except for where integration depends upon changed strings listed above (in the description) in some way
```
JakaMohorkoDS, line 33: I'd add some extra information:
- If relying on string comparison to hardcoded old IDs of things like FB, device, server types, or protocol  IDs, those comparisons will need to be updated to match the new IDs. eg. a check like `if (fbType.getId() == "ref_fb_module_renderer")` will never be true
- Old IDs can still be used when adding new objects to a device via `addDevice`/`addFunctionBlock` or similar calls.

JakaMohorkoDS (approved):

core/coreobjects/tests/serialization_test.cpp
```diff
@@ -94,8 +94,8 @@ TEST_F(SerializationTest, JsonSerObjectToJsonString)
     PropertyObjectPtr child1_2_1 = PropertyObject();
     PropertyObjectPtr child2_1 = PropertyObject();
 
-    child1_2_1.addProperty(StringProperty("String", "string"));
-    child1_2_1.addProperty(StringPropertyBuilder("ReadOnlyString", "string").setReadOnly(true).build());
+    child1_2_1.addProperty(StringProperty("String", "String"));
```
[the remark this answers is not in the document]
aljazfrancic: This file was restored to the previous version.

core/coreobjects/tests/test_property_object.cpp
```diff
@@ -1865,66 +1865,66 @@ TEST_F(PropertyObjectTest, BeginEndUpdateClonedClear)
...
+    ASSERT_NE(propObj.getPropertyValue("child"), classObj.getProperty("child").getDefaultValue());
+    ASSERT_TRUE(classObj.getProperty("child").getDefaultValue().isFrozen());
+    ASSERT_FALSE(propObj.getPropertyValue("child").isFrozen());
 }
 
 // TODO: Enable once nested object property deserialization works without classes
 TEST_F(PropertyObjectTest, DISABLED_ClonedObjectsSerialize)
 {
-    const std::string expectedJson = R"({"__type":"PropertyObject","propValues":{"Child1":{"__type":"PropertyObject","propValues":{"Child2":{"__type":"PropertyObject"}}}}})";
+    const std::string expectedJson = R"({"__type":"PropertyObject","PropValues":{"Child1":{"__type":"PropertyObject","PropValues":{"Child2":{"__type":"PropertyObject"}}}}})";
```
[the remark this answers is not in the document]
aljazfrancic: This change has been reverted.

core/coretypes/include/coretypes/type_name_detail.h
```diff
@@ -99,49 +99,49 @@ namespace daq::Detail
...
 
     template <>
     constexpr auto nameOf<long long>()
     {
-        return constexprString("int64");
+        return constexprString("Int64");
     }
 
     template <>
     constexpr auto nameOf<unsigned long long>()
```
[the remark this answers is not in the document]
aljazfrancic: This change has been reverted.

core/coretypes/tests/test_type_name.cpp
```diff
@@ -17,37 +17,37 @@ namespace daq
 TEST_F(TypeNameTest, Char)
 {
     constexpr auto intType = typeName<char>();
-    ASSERT_EQ(intType, "char");
+    ASSERT_EQ(intType, "Char");
```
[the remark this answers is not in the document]
aljazfrancic: This change has been reverted.

PR #357, opened 2024-06-30
Other/fix ci issues

dejancrn:
This PR fixes some of the CI issues:

1.) Logger crashing
---------------------------------
Logger is crashing because when LOG macro is called, it makes a copy of log message and puts into intermediary buffer which is then picked up by a worker thread. However, it also takes a pointer to function name in file name. These are pointers to strings which have global/static lifetime, but only until a module is unloaded. So the scenario for crash is something like something like this: module logs a message -> module unloads -> worker thread tries to log a message -> crash because it accesses a pointer to function name/file name from unloaded module.

The workaround for this is not to take a function name / file name when logging. We anyway do not print those in the logs, or do we?

2.) Flaky tests with thread::sleep
Here we can use our blocking readers. The solution executes faster because it unblocks as soon it has enough data available, on the other hand it waits for a lot more time.

3.) Streaming LT race condition
Fixed in streaming protocol LT repo.

4.) There are more issues :(

### DIFF
--- a/core/opendaq/logger/include/opendaq/log.h
+++ b/core/opendaq/logger/include/opendaq/log.h
@@ -33,7 +33,7 @@
 /// Plain
 
 #define DAQLOG_PLAIN(loggerComponent, message, level) loggerComponent.logMessage(\
-    daq::SourceLocation{__FILE__, __LINE__, OPENDAQ_CURRENT_FUNCTION}, message, level);
+    daq::SourceLocation{nullptr, 0, nullptr}, message, level);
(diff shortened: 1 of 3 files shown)


JakaMohorkoDS (approved):

Unsttopabull (approved):
Until a better solution is found this is ok for the moment. 
A proper solution would be to find some other way of storing this info e.g.: a custom spdlog thread pool or a modification of the current one seems the most feasible.

PR #365, opened 2024-07-04
<TBBAS-1523> Enable signal adding and removing within LT-streaming

NikolaiShipilov:
## Type of change

# Checklist:

# Description:


janm-dw (approved):

shared/libraries/websocket_streaming/src/streaming_client.cpp
```diff
@@ -38,11 +39,46 @@ StreamingClient::StreamingClient(const ContextPtr& context, const std::string& h
...
+            LOG_I("Websocket streaming client background thread joined");
+        }
+        else
+        {
+            LOG_W("Websocket streaming client background thread is not joinable");
+        }
+    }
+    else
+    {
+        LOG_C("Websocket streaming client background thread cannot join itself");
```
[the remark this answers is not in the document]
NikolaiShipilov: An additional exception could be thrown here, but it may not be necessary since the program will call std::terminate in thread destructor if the thread is not joined, as in this case. The critical level log message here just helps identify which thread caused the termination.

shared/libraries/websocket_streaming/src/streaming_server.cpp
```diff
@@ -395,4 +362,126 @@ void StreamingServer::unsubscribeHandler(const std::string& signalId, OutputSign
...
+            addToOutputSignals(daqSignal, outputSignals, writer);
+            outputSignal = outputSignals.at(signalId);
+
+            if (placeholderSignal->isSubscribed())
+                outputSignal->setSubscribed(true);
+        }
+        catch (const DaqException& e)
+        {
+            LOG_W("Failed to create an output LT streaming signal for {}: {}", daqSignal.getGlobalId(), e.what());
+        }
```
JakaMohorkoDS, line 428: This warning likely appears quite often, as only a select amount of signal types are supported. It'd be great if this could be a bit more descriptive and have a reason as to why the signal isn't added. Eg. "Failed to create LT streaming output signal. Signals with that have an explicit rule domain signal are not supported in LT streaming."

Similarly, warning logs below should be a bit more descriptive.

shared/libraries/websocket_streaming/src/websocket_streaming_server.cpp
```diff
@@ -86,4 +92,90 @@ void WebsocketStreamingServer::stopInternal()
...
+        return;
+
+    LOG_I("Component: {}; is updated", updatedComponentGlobalId);
+
+    // remove all registered signal of updated component since those might be modified or removed
+    streamingServer.removeComponentSignals(updatedComponentGlobalId);
+
+    // add updated versions of signals
+    addSignalsOfComponent(updatedComponent);
+}
```
JakaMohorkoDS, line 179: This is quite an expensive change - basically we will lose data on every single signal of the component. We also shouldn't really care too much about modifications here - those should be propagated through the data path as descriptor changes. 

We should, however, handle any removals/additions.

NikolaiShipilov: Fixed

JakaMohorkoDS (approved):

PR #373, opened 2024-07-08
Force Node 16 usage in Manylinux CI

JakaMohorkoDS:
# Description:

Due to [GitHub actions transitioning to Node20](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/), the manylinux CI jobs are failing. This fix enforces the usage of Node16 in those actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,7 @@ env:
   windows_x64_artifact: SdkWindows64
   linux_x64_artifact: SdkLinux64
   nuget_artifact: Packages (openDAQ.Net NuGet)
+  ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true # Workaround for: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/
 
 jobs:
   build_windows:
@@ -461,7 +462,7 @@ jobs:
         run : git config --global --add safe.directory '*'
 
       - name: Checkout
-        uses: actions/checkout@v3 # TODO https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/
+        uses: actions/checkout@v3
 
       - name: Install build dependencies
         run: |
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -406,13 +406,6 @@ jobs:
         CXX: ${{ matrix.cxx }}
 
     steps:
-      # Addresses CentOS 7 EoL issue https://serverfault.com/questions/1161816/mirrorlist-centos-org-no-longer-resolve
-      - name: Update yum .repo files
-        run: |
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -461,7 +462,7 @@ jobs:
         run : git config --global --add safe.directory '*'
 
       - name: Checkout
-        uses: actions/checkout@v3 # TODO https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/
+        uses: actions/checkout@v3
```
mojca, line 465: Doesn't the following fix the original problem? (Please note that I did not test.)
```suggestion
        uses: actions/checkout@v4
```

mojca: Ah, sorry. That's an ancient Linux distro that would probably require some tweaking to get Node 20 running on it in the first place.
Like https://tecadmin.net/install-latest-nodejs-and-npm-on-centos/ pointing to https://rpm.nodesource.com/

aljazfrancic:
Not only has Node 16 reached EoL, but [so has CentOS](https://blog.centos.org/2023/04/end-dates-are-coming-for-centos-stream-8-and-centos-linux-7/).
I would suggest foregoing CentOS in our CI.

mojca:
> Not only has Node 16 reached EoL, but [so has CentOS](https://blog.centos.org/2023/04/end-dates-are-coming-for-centos-stream-8-and-centos-linux-7/). I would suggest foregoing CentOS in our CI.

According to https://github.com/pypa/manylinux?tab=readme-ov-file#docker-images maybe the best way would be to select `manylinux_2_28` with GLIBC 2.28 (based on AlmaLinux 8 which is kind of "successor" to the discontinued CentOS).

aljazfrancic (approved):
`ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true` seems to do the trick for now (everything's green). [However, this will only work until they upgrade the runner, removing Node16 later in "the spring".](https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/) So I'd merge this so we can continue with other PRs, but would consider this as a temporary workaround.

JakaMohorkoDS:
> > Not only has Node 16 reached EoL, but [so has CentOS](https://blog.centos.org/2023/04/end-dates-are-coming-for-centos-stream-8-and-centos-linux-7/). I would suggest foregoing CentOS in our CI.
> 
> According to https://github.com/pypa/manylinux?tab=readme-ov-file#docker-images maybe the best way would be to select `manylinux_2_28` with GLIBC 2.28 (based on AlmaLinux 8 which is kind of "successor" to the discontinued CentOS).

Sure, but that means our binaries are compatible with less systems than before. I'll keep this workaround for now, and update to a newer manylinux version later.

PR #374, opened 2024-07-09
Change company name from `Blueberry d.o.o.` to `openDAQ d. o. o.`

aljazfrancic:
Change company name from `Blueberry d.o.o.` to `openDAQ d. o. o.` in license headers (spaces after dots are correct)

### DIFF
--- a/app/audio_application/include/audio_application/utils.h
+++ b/app/audio_application/include/audio_application/utils.h
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022-2024 openDAQ d. o. o.
+ * Copyright 2022-2024 openDAQ d.o.o.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/app/audio_application/include/audio_application/utils.h
+++ b/app/audio_application/include/audio_application/utils.h
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022-2024 Blueberry d.o.o.
+ * Copyright 2022-2024 openDAQ d. o. o.
  *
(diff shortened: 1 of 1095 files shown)


app/audio_application/include/audio_application/utils.h
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022-2024 Blueberry d.o.o.
+ * Copyright 2022-2024 openDAQ d. o. o.
```
mojca, line 2: I have never ever seen any spaces in this context.

```suggestion
 * Copyright 2022-2024 openDAQ d.o.o.
```

[Ajpes](https://www.ajpes.si/podjetje/OpenDAQ_d.o.o.?enota=812508&EnotaStatus=1) says `OpenDAQ d.o.o.` (also with capital letter "O").

mojca: There are some length discussions at https://fran.si/iskanje?Query=d.%20o.%20o.

(I'm aware that the language rules would call for spaces, but I've never seen that used in practice. If we want to follow the grammar rules, it should be different in Ajpes as well.)

aljazfrancic: I've spoken to Sonja about it, and she says the capital "O" was a clerical mistake, when registering the name. Also she said no spaces. So the correct/intended version is actually `openDAQ d.o.o.`. So I'll change it to that. I suppose I'll get changed in Ajpes later.

mojca: Well, strictly speaking ... our grammar doesn't allow names to start with lowercase either 🤣

JakaMohorkoDS (approved):

PR #378, opened 2024-07-10
Add python example

Cverlllly:
The example:

- selects device
- selects signal
- measures for 10 seconds
- plots the signal
- calculates some statistics (min, max, ave, rms).

The example shows how to process data collected with openDAQ in python using numpy and matlibplot.

### DIFF
--- a/examples/python/signal_graphed_overtime.py
+++ b/examples/python/signal_graphed_overtime.py
@@ -1,0 +1,69 @@
+import opendaq
+import time
+import numpy as np
+import matplotlib.pyplot as plt
+
+
+def select_device(instance):
+    available_devices_info = instance.available_devices
+    print("Available devices - Select one to connect to:")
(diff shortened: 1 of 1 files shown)


examples/python/signal_graphed_overtime.py
```diff
@@ -0,0 +1,70 @@
...
+min_val = np.min(signal_values)
+max_val = np.max(signal_values)
+rms = np.sqrt(np.mean(signal_values ** 2))
+print(f'Average value: {avg}')
+print(f'Maximum value: {max_val}')
+print(f'Minimum value: {min_val}')
+print(f'Mean root square value: {rms}')
+
+# close the device
+instance.remove_device(device)
```
Yarikk26, line 70: Isn't necessary IMO

Yarikk26 (approved):

PR #380, opened 2024-07-11
Add support for native streaming authentication

janm-dw:


NikolaiShipilov (approved):

core/opendaq/modulemanager/src/module_manager_impl.cpp
```diff
@@ -1038,6 +1060,22 @@ ServerCapabilityPtr ModuleManagerImpl::mergeDiscoveryAndDeviceCap(const ServerCa
...
+    for (const auto& prop : general.getAllProperties())
+    {
+        const auto name = prop.getName();
+        const auto value = general.getPropertyValue(name);
+        const auto defaultValue = prop.getDefaultValue();
+
+        if (tartgetObj.hasProperty(name) && tartgetObj.getPropertyValue(name) == defaultValue)
+            tartgetObj.setPropertyValue(name, value);
+    }
+}
```
JakaMohorkoDS, line 1077: Can't we just use clone here?

shared/libraries/native_streaming_protocol/src/native_streaming_client_handler.cpp
```diff
@@ -30,6 +33,22 @@ NativeStreamingClientHandler::~NativeStreamingClientHandler()
...
+    auto normalizedObject = PropertyObject();
+
+    normalizedObject.addProperty(StringProperty("Username", ""));
+    normalizedObject.addProperty(StringProperty("Password", ""));
+
+    if (authenticationObject.assigned())
+    {
+        normalizedObject.setPropertyValue("Username", authenticationObject.getPropertyValue("Username"));
+        normalizedObject.setPropertyValue("Password", authenticationObject.getPropertyValue("Password"));
+    }
```
JakaMohorkoDS, line 47: We should probably check that the Username and Password properties are present.

JakaMohorkoDS (approved):

core/opendaq/modulemanager/src/module_manager_impl.cpp
janm-dw: Unfortunately no. I want the property to be copied from general object to target object only if the value is not present in the target object. This means that if the user sets username and password on the general config object and on the device config object, the values from the more specific device config object will be used.

PR #390, opened 2024-07-18
[TBBAS-1716] Python GUI App: short_id generating bugfix

Yarikk26:
## Type of change:

# Checklist:

# Description:

Short ID generating algo reworked. Long ID could be used in case of collisions

### DIFF
--- a/examples/python/gui_demo/app_context.py
+++ b/examples/python/gui_demo/app_context.py
@@ -87,27 +87,63 @@ class AppContext(object):
             images[file.split('.')[0]] = image
         self.icons = images
 
-    def short_id(self, global_id: str):
+    def is_server(self, device_id):
+        if not device_id:
+            return False
+
+        component = self.instance.find_component(device_id)
+        if component is None or not daq.IDevice.can_cast_from(component):
+            return False
+
+        device = daq.IDevice.cast_from(component)
+        return len(device.info.server_capabilities) > 0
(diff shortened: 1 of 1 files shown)


examples/python/gui_demo/app_context.py
```diff
@@ -86,28 +86,36 @@ def load_icons(self, directory):
...
         for signal in device.signals_recursive:
-            self.signals[device.global_id][self.short_id(
-                signal.global_id)] = signal
-            
+            short_id = self.short_id(signal.global_id)
+
+            if short_id not in self.signals[device.global_id]:
+                self.signals[device.global_id][short_id] = signal
+            else: #insert global id in case of collision
+                self.signals[device.global_id][signal.global_id] = signal
```
JakaMohorkoDS, line 117: If we have a collision, we should also retroactively replace the one we clashed with.

JakaMohorkoDS (commented):
I know we're currently just parsing the ID - how hard would it be to check whether the device has server capabilities or not? I'd really shorten the ID to the first device with server capabilities, as I foresee having issues with local devices if we do not do so.

examples/python/gui_demo/app_context.py
Yarikk26: Done

JakaMohorkoDS (approved):

PR #397, opened 2024-07-23
Pass authenticated user reference to ConfigProtocolServer

janm-dw:


core/coretypes/include/coretypes/json_serializer_impl.h
```diff
@@ -56,9 +57,13 @@ class JsonSerializerImpl : public ImplementationOf<ISerializer>
     ErrCode INTERFACE_FUNC endObject() override;
     ErrCode INTERFACE_FUNC writeString(ConstCharPtr string, SizeT length) override;
 
+    ErrCode INTERFACE_FUNC getUser(IBaseObject** user) override;
+    ErrCode INTERFACE_FUNC setUser(IBaseObject* user) override;
```
JakaMohorkoDS, line 61: Missing changelog.

JakaMohorkoDS (approved):
Will we ever be able to change users once a session is established? If so we will probably need to change these callbacks a bit.

janm-dw:
> Will we ever be able to change users once a session is established? If so we will probably need to change these callbacks a bit.

If this is ever the case, I would argue that a new session has to be established.

PR #411, opened 2024-07-28
Reference Domain Info

aljazfrancic:
## Type of change

# Checklist:

# Description:
* Reference Domain Info was added as an interface that gives additional information about the reference domain
* Reference Domain Info has getters for:
    * Reference Domain ID (Signals with the same Reference Domain ID share a common synchronization source and can be read together)
    * Reference Domain Offset (which must be added to the domain values of the Signal for them to be equal to that of the sync source)
    * Reference Time Source (which is used to determine if two signals with different Domain IDs can be read together); possible values are: 
        * [Tai](https://en.wikipedia.org/wiki/International_Atomic_Time)
        * [Gps](https://en.wikipedia.org/wiki/Global_Positioning_System#Timekeeping)
        * [Utc](https://en.wikipedia.org/wiki/Coordinated_Universal_Time)
        * Unknown
    * Uses Offset
* There is also a builder available for creating Reference Domain Info
* Reference Domain Info is a part of two interfaces:
    * Device Domain
    * Data Descriptor
* Reference Domain Info is currently only supported over Native, not over OPC UA or LT Streaming protocols (this will cause two data descriptor changed events to be sent when combining supported and unsupported protocols for configuration/streaming - for example OPC UA and Native streaming)


JakaMohorkoDS (commented):
Looks good! I think we need a slight amount of discussion on naming. Also There should be an example in the reference device/simulator. Its signals should have a domain ID (likely the local ID of the device), and a 0 offset.

Are all of the added fields visible in the Python GUI app?

regression/simulator/regression_simulator.cpp
```diff
@@ -51,7 +54,7 @@ int main(int /*argc*/, const char* /*argv*/[])
     // Add custom Struct Property
     instance.addProperty(StructProperty("TestStruct", Struct("TestName", dict, TypeManager())));
     // Add custom Argument Info Property
-    auto argInfo = ArgumentInfo_Create(String("TestArgInfo"), CoreType::ctInt);
+    auto argInfo = ArgumentInfo(String("TestArgInfo"), CoreType::ctInt);
```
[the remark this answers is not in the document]
aljazfrancic: We switched away from them here...
The reason was actually because Device Domain factory changed, while this code should work with both the new and the old implementation. However your other comment suggested leaving the factory as is and adding a new one or moving to a builder pattern. So if we do that, these changes are unwarranted. However I think the code looks a bit cleaner this way (without `_Create`), so I'll leave these modifications in.

core/opendaq/reader/src/multi_reader_impl.cpp
```diff
@@ -176,11 +173,37 @@ ListPtr<ISignal> MultiReaderImpl::getSignals() const
     return list;
 }
 
-static void checkSameDomain(const ListPtr<IInputPortConfig>& list)
+struct MultiReaderImpl::DomainBin
+{
+    StringPtr id;
+    BoolPtr isAbs;
```
denise-opendaq, line 179: maybe to rename them as referenceID referenceIsAbsolute?

denise-opendaq (approved):

changelog/changelog_3.0.0-4.0.0.txt
```diff
@@ -1,21 +1,56 @@
+20.08.2024
+Description:
+    - Reference Domain Info was added
+Required integration changes:
+    - None, however, users are encouraged to use Reference Domain Info
```
JakaMohorkoDS, line 5: This should reflect to some extent what the domain info is. For integration changes, I'd write where this is needed (signals, device domain), and maybe some pointers for references.

docs/Antora/modules/glossary/pages/glossary.adoc
```diff
@@ -103,8 +103,7 @@ The Packet offset is contained in every packet with a _non-explicit_ rule.
 * If the Data rule is _explicit_, samples are found as part of the data buffer.
 * If the Data rule is _linear_, or _constant_, the buffer is `null`, and each sample is calculated with a given packet offset
 (the packet offset is available in every <<data_packet>> when required for the rule calculation):
-** *Linear rule*: `Sample value = Packet offset + Index * Delta + Start`; Where `Index` is the index of the sample in a
-packet. `Delta` and `Start` are given by the Data rule.
+** *Linear rule*: `ReferenceDomainOffset + PacketOffset + SampleIndex * delta + start`; Where `SampleIndex` is the index of the sample in a packet. `delta` and `start` are given by the Data rule. `ReferenceDomainOffset` is only applied if available in Reference Domain Info of the Domain Signal's Data Descriptor.
```
JakaMohorkoDS, line 106: This is not true for only the linear rule. This is also applied to explicit signals. The rule calculation itself does not include the reference domain offset.

docs/Antora/modules/knowledge_base/pages/signals.adoc
```diff
@@ -108,6 +108,9 @@ in later sections.
 | Metadata
 | A dictionary of key-value pairs. The metadata dictionary is used to add any additional information required to correctly interpret the data. This information is not interpreted by openDAQ core.
 
+| Reference Domain Info
+| An optional field for Domain Signals that gives additional information about the reference domain, such as the Reference Domain ID (Signals with the same Reference Domain ID share a common synchronization source and can be read together), Reference Domain Offset (which must be added to the domain values of the Signal for them to be equal to that of the sync source), Reference Time Source (which is used to determine if two signals with different Domain IDs can be read together; possible values are Tai, Gps, Utc, and Unknown), and Uses Offset.
```
JakaMohorkoDS, line 112: A link to an explanation of the different time sources might be good to have.

docs/Antora/modules/knowledge_base/pages/signals.adoc
```diff
@@ -119,7 +122,7 @@ xref:packets.adoc[Packets] will not contain any data.
 To obtain data from Packets of Signals with an `Implicit` Data Rule, the values are
 calculated depending on the Rule type. Currently, two rules are available in openDAQ:
 
-* **Linear Rule**: The Rule parameters include a `delta` and `start` numbers. The values are calculated according to the following equation: `PacketOffset + SampleIndex * delta + start`, where the `PacketOffset` is available on the Packet carrying the data and the `SampleIndex` represents the index of the data entry being calculated within a given packet.
+* **Linear Rule**: The Rule parameters include a `delta` and `start` numbers. The values are calculated according to the following equation: `ReferenceDomainOffset + PacketOffset + SampleIndex * delta + start`, where the `PacketOffset` is available on the Packet carrying the data and the `SampleIndex` represents the index of the data entry being calculated within a given packet. `ReferenceDomainOffset` is only applied if available in Reference Domain Info of the Domain Signal's Data Descriptor.
```
JakaMohorkoDS, line 125: Similarly as above, offset does not only apply to the linear rule (it's not part of it).

modules/ref_device_module/src/ref_device_impl.cpp
```diff
@@ -123,7 +123,10 @@ void RefDeviceImpl::initClock()
 
     microSecondsFromEpochToDeviceStart = std::chrono::duration_cast<std::chrono::microseconds>(startAbsTime.time_since_epoch());
 
-    this->setDeviceDomain(DeviceDomain(RefChannelImpl::getResolution(), RefChannelImpl::getEpoch(), UnitBuilder().setName("second").setSymbol("s").setQuantity("time").build()));
+    this->setDeviceDomain(DeviceDomain(RefChannelImpl::getResolution(),
+                                       RefChannelImpl::getEpoch(),
+                                       UnitBuilder().setName("second").setSymbol("s").setQuantity("time").build(),
+                                       ReferenceDomainInfoBuilder().setReferenceDomainId(localId).setReferenceDomainOffset(0).build()));
```
JakaMohorkoDS, line 129: I'd set the reference domain ID to the serial number. Maybe "manufacturer_serial".

core/opendaq/modulemanager/src/module_manager_impl.cpp
```diff
@@ -779,7 +779,7 @@ ErrCode ModuleManagerImpl::createFunctionBlock(IFunctionBlock** functionBlock, I
                 }
             }
 
-            localIdStr = fmt::format("{}_{}", typeId, maxNum + 1);
+            localIdStr = fmt::format("{}{}", typeId, maxNum + 1);
```
JakaMohorkoDS, line 782: This is likely not a good idea. This basically breaks every single setup that contains a function block by changing their local IDs (serialization keys)

core/opendaq/reader/src/multi_reader_impl.cpp
```diff
@@ -213,34 +235,93 @@ static void checkSameDomain(const ListPtr<IInputPortConfig>& list)
...
+            LOG_I(R"(Domain signal "{}" Reference Domain Info not assigned.)", domain.getLocalId());
+        }
+        else
+        {
+            auto referenceDomainID = referenceDomainInfo.getReferenceDomainId();
+
+            if (!referenceDomainID.assigned() || referenceDomainID.getLength() == 0)
+            {
+                // This will perhaps be bumped up to a higher severity later on (warning)
+                LOG_I(R"(Domain signal "{}" Reference Domain ID not assigned.)", domain.getLocalId());
```
JakaMohorkoDS, line 276: I'd leave this at debug, not info. Same for the other logging here.

JakaMohorkoDS (approved):
Looks good, it would be worth writing into the changelog that this is not supported over OPC UA and LT yet. Additionally, this should be added to the protocol matrix in the "protocol_features_characteristics.adoc" document with native having a checkmark for this. A comment is also in place about the double descriptor behaviour when combining native and another streaming/config protocol.

core/opendaq/reader/tests/test_multi_reader.cpp
```diff
@@ -2292,3 +2276,1312 @@ TEST_F(MultiReaderTest, ReconnectWhileReading)
...
+    addSignal(0, 113, createDomainSignal("1993"));
+    addSignal(0, 113, createDomainSignal("1993"));
+    addSignal(0, 133, createDomainSignal("1993"));
+
+    ASSERT_NO_THROW(MultiReader(signalsToList()));
+
+    auto received = privateSink.waitForMessage(9001);
+    ASSERT_EQ(received, true);
+    auto str = privateSink.getLastMessage();
+    ASSERT_EQ(str, R"(Domain signal "time" Reference Domain Info not assigned.)");
```
denise-opendaq, line 2294: Domain signal "time" Reference Domain Info is not assigned.

core/opendaq/reader/tests/test_multi_reader.cpp
```diff
@@ -2292,3 +2276,1312 @@ TEST_F(MultiReaderTest, ReconnectWhileReading)
...
+TEST_F(MultiReaderTest, ReferenceDomainIdInequality01)
+{
+    constexpr const auto NUM_SIGNALS = 3;
+    readSignals.reserve(NUM_SIGNALS);
+
+    addSignal(0, 113, createDomainSignal("1993", nullptr, nullptr, ReferenceDomainInfoBuilder().setReferenceDomainId("B").build()));
+    addSignal(0, 113, createDomainSignal("1993", nullptr, nullptr, ReferenceDomainInfoBuilder().setReferenceDomainId("A").build()));
+    addSignal(0, 113, createDomainSignal("1993", nullptr, nullptr, ReferenceDomainInfoBuilder().setReferenceDomainId("A").build()));
+
+    ASSERT_THROW_MSG(MultiReader(signalsToList()), InvalidStateException, "Reference domain incompatible.");
```
denise-opendaq, line 2353: The reference domain is incompatible

core/opendaq/reader/tests/test_multi_reader.cpp
```diff
@@ -2292,3 +2276,1312 @@ TEST_F(MultiReaderTest, ReconnectWhileReading)
...
+        0, 113, createDomainSignal("1993", nullptr, nullptr, ReferenceDomainInfoBuilder().setReferenceTimeSource(TimeSource::Tai).build()));
+    addSignal(
+        0, 113, createDomainSignal("1993", nullptr, nullptr, ReferenceDomainInfoBuilder().setReferenceTimeSource(TimeSource::Tai).build()));
+    addSignal(
+        0, 113, createDomainSignal("1993", nullptr, nullptr, ReferenceDomainInfoBuilder().setReferenceTimeSource(TimeSource::Tai).build()));
+
+    ASSERT_NO_THROW(MultiReader(signalsToList()));
+
+    auto received = privateSink.waitForMessage(9001);
+    ASSERT_EQ(received, true);
```
denise-opendaq, line 2431: you can use ASSERT_TRUE(received) :)

core/opendaq/reader/tests/test_multi_reader.cpp
```diff
@@ -2292,3 +2276,1312 @@ TEST_F(MultiReaderTest, ReconnectWhileReading)
...
+                                 nullptr,
+                                 ReferenceDomainInfoBuilder().setReferenceDomainId("A").setReferenceTimeSource(TimeSource::Tai).build()));
+    addSignal(0,
+              113,
+              createDomainSignal("1993",
+                                 nullptr,
+                                 nullptr,
+                                 ReferenceDomainInfoBuilder().setReferenceDomainId("A").setReferenceTimeSource(TimeSource::Tai).build()));
+
+    ASSERT_THROW_MSG(MultiReader(signalsToList()), InvalidStateException, "One known Reference Time Source is allowed per Multi Reader.");
```
denise-opendaq, line 2543: Only one known Reference Time Source is allowed per Multi Reader.

JakaMohorkoDS (approved):

Issue #335, opened 2024-06-10
Calling getLastValue on connected ISignal causes deadlock


Yarikk26:
## Description
Calling **getLastValue** on **ISignal** connected to **Render FB** causing deadlock on Linux and exception on Windows.  
*Reference device modules are not affected*
### Logs
Application trace shows blocked execution on Linux and thrown exception on Windows causing impossibility to acquire the last signal's value
#### Linux
```
 --- modulename: utils, funcname: get_last_value_for_signal
utils.py(137):     last_value = 'N/A'
utils.py(138):     if output_signal is not None and daq.ISignal.can_cast_from(output_signal):
utils.py(139):         try:
utils.py(140):             last_value = output_signal.last_value
```
#### Windows
```
output_signal_row.py(19):         last_value = get_last_value_for_signal(output_signal)
 --- modulename: utils, funcname: get_last_value_for_signal
utils.py(137):     last_value = 'N/A'
utils.py(138):     if output_signal is not None and daq.ISignal.can_cast_from(output_signal):
utils.py(139):         try:
utils.py(140):             last_value = output_signal.last_value
utils.py(141):         except RuntimeError as e:
utils.py(142):             print(f'Error reading last value: {e}')
Error reading last value: resource deadlock would occur: resource deadlock would occur
utils.py(143):     return last_value
```
#### Preconditions
1. You need **simulator VM** running
2. **Python OpenDAQ package** installed
3. **GUI App** running by calling `python -m opendaq` (or with trace running from repo root dir `python -m trace -t examples/python/gui_demo.py`)
#### Steps to reproduce
1. Connect **Reference device simulator** to **OpenDAQ Client** via **Add device dialog**
2. Add **Renderer function block** to **OpenDAQ Client** via **Add function block** dialog
3. Select **Renderer FB** and connect **ai0** signal to it
4. Select **refch0**

Application will stop responding on Linux or will display **N/A** in **Output signals** view on Windows

PR #419, opened 2024-07-31
Fix deadlock when reading last value over native

JakaMohorkoDS:
## Type of change

# Description:

Fixes https://github.com/openDAQ/openDAQ/issues/335


Yarikk26 (approved):
No more deadlocks on my machine

PR #423, opened 2024-07-31
Native configuration access control

janm-dw:
- [x] partial serialization is supported
- [x] user is only able to see component assigned to its role
- [x] user is only able to modify components assigned to its role


core/coreobjects/include/coreobjects/property_object_internal.h
```diff
@@ -37,6 +37,7 @@ DECLARE_OPENDAQ_INTERFACE(IPropertyObjectInternal, IBaseObject)
     virtual ErrCode INTERFACE_FUNC clone(IPropertyObject** cloned) = 0;
     virtual ErrCode INTERFACE_FUNC setPath(IString* path) = 0;
     virtual ErrCode INTERFACE_FUNC isUpdating(Bool* updating) = 0;
+    virtual ErrCode INTERFACE_FUNC hasUserReadAccess(IBaseObject* userContext, Bool * hasAccessOut) = 0;
```
JakaMohorkoDS, line 40: The context parameter can probably be IUser?

JakaMohorkoDS (approved):
Looks good, I would just add some examples to our reference device that make use of the permission manager.

core/coreobjects/include/coreobjects/property_object_internal.h
janm-dw: In most cases we are passing serializer.getUser() to this method. It returns IBaseObject, since we did not want ISerializer to know about IUser. Therefore I wanted to wrap this type checking inside hasUserReadAccess() method.

aljazfrancic (approved):
Reviewed and tested! Looks good to me!
I would, however, suggest noting down what is supported and what isn't in `docs/Antora/modules/knowledge_base/pages/protocol_features_characteristics.adoc` before merging.

PR #439, opened 2024-08-07
add generate_connection_view_image.py, connection_view.py, and modified gui_demo.py

Cverlllly:
added generrate_conencection_view_image:
-generates image of channels, function blocks and possible connection between them 

added connection_view:
-displays image that's generated 
-export button for exporting image as svg or png
-setting button for showing additional properties, attributes, input ports, showing disconnected  signals or showing only nodes that are connected
-setting for changing font sizes or colors

modified gui_demo.py:
-added new tab connection view
-checker if graphviz is installed

### DIFF
--- a/examples/python/connection_view.py
+++ b/examples/python/connection_view.py
@@ -1,0 +1,485 @@
+import math
+import warnings
+import tkinter as tk
+import platform
+
+from tkinter import ttk, filedialog
+from tkinter.colorchooser import askcolor
+
+import opendaq
+from PIL import Image, ImageTk
+from generate_connection_view_image import ConnectionMap
+
+OS = platform.system()
+
+
+# taken from https://github.com/foobar167/junkyard/blob/master/zoom_advanced3.py
+
+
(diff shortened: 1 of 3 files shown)


examples/python/connection_view.py
```diff
@@ -0,0 +1,485 @@
+import math
+import warnings
+import tkinter as tk
+import platform
+
+from tkinter import ttk, filedialog
+from tkinter.colorchooser import askcolor
+
+import opendaq
+from PIL import Image, ImageTk
```
Yarikk26, line 10: New dependency here. Wheels manifest should be updated

examples/python/generate_connection_view_image.py
```diff
@@ -0,0 +1,271 @@
+import opendaq
+import pydot
```
Yarikk26, line 2: same thing with dependency

JakaMohorkoDS:
@dejancrn What's the status of this PR?

dejancrn:
It's still on my task list. I need to take some time to review and test it thoroughly.

PR #444, opened 2024-08-09
Allow passing of device options via connection string as url options

dejancrn:
Allows to pass options as a part of a connection string.

This only works for device only config options for now.

### DIFF
--- a/core/opendaq/modulemanager/src/module_manager_impl.cpp
+++ b/core/opendaq/modulemanager/src/module_manager_impl.cpp
@@ -13,7 +13,7 @@
 #include <opendaq/device_private.h>
 #include <string>
 #include <future>
-#include <boost/algorithm/string/predicate.hpp>
+#include <boost/algorithm/string.hpp>
 #include <opendaq/search_filter_factory.h>
 #include <coretypes/validation.h>
 #include <opendaq/server_capability_config_ptr.h>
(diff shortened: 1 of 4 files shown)


core/opendaq/modulemanager/src/module_manager_impl.cpp
```diff
@@ -1173,6 +1195,36 @@ std::string ModuleManagerImpl::getPrefixFromConnectionString(std::string connect
...
+
+    tsl::ordered_map<std::string, BaseObjectPtr> optionsMap;
+    for (const auto& option: options)
+    {
+        std::vector<std::string> keyAndValue;
+        boost::split(keyAndValue, option, boost::is_any_of("="));
+        if (keyAndValue.size() != 2)
+            throw InvalidParameterException("Invalid connection string");
+
+        BaseObjectPtr value = keyAndValue[1];
```
JakaMohorkoDS, line 1221: Should we be using `EvalValue(keyAndValue[1])` here? Since now we're always calling `setPropertyValue()` with a string value, relying that the prop object implementation will properly convert the string.

JakaMohorkoDS (approved):

core/opendaq/modulemanager/src/module_manager_impl.cpp
dejancrn: Good point, I will update the PR

PR #445, opened 2024-08-09
added add config dialog it gives you an option to change config befor…

Cverlllly:
added add config dialog it gives you an option to change config before adding a device

### DIFF
--- a/examples/python/gui_demo/components/add_device_dialog.py
+++ b/examples/python/gui_demo/components/add_device_dialog.py
@@ -8,6 +8,120 @@ from ..event_port import *
 from .diaolog import Dialog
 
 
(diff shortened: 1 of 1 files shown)


examples/python/gui_demo/components/add_device_dialog.py
```diff
@@ -8,6 +8,120 @@
 from .diaolog import Dialog
 
 
+class AddConfigDialog(tk.Toplevel):
+    def __init__(self, parent, context, connection_string=None):
```
Yarikk26, line 12: Why not Dialog as base?

dejancrn:
PR closed since new PR with this changes is opened

PR #456, opened 2024-08-20
IPv6 link local address interface added to connection string; connection string reachability check

JakaMohorkoDS:
## Type of change

# Description:

- Primary connection string is changed if the address is unreachable
- IPv6 connection strings can include link local interface identifier

### DIFF
--- a/shared/libraries/websocket_streaming/src/streaming_client.cpp
+++ b/shared/libraries/websocket_streaming/src/streaming_client.cpp
@@ -243,7 +243,7 @@ void StreamingClient::parseConnectionString(const std::string& url)
     std::smatch match;
 
     // parsing connection string to four groups: prefix, host, port, path
-    auto regexIpv6Hostname = std::regex(R"(^(.*://)?(\[[a-fA-F0-9:]+(?:\%\d+)?\])(?::(\d+))?(/.*)?$)");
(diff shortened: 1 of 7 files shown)


denise-opendaq (approved):

shared/libraries/websocket_streaming/src/streaming_client.cpp
```diff
@@ -243,7 +243,7 @@ void StreamingClient::parseConnectionString(const std::string& url)
     std::smatch match;
 
     // parsing connection string to four groups: prefix, host, port, path
-    auto regexIpv6Hostname = std::regex(R"(^(.*://)?(?:\[([a-fA-F0-9:]+)\])(?::(\d+))?(/.*)?$)");
+    auto regexIpv6Hostname = std::regex(R"(^(.*://)?(?:\[([a-fA-F0-9:]+(?:\%\s+)?)\])(?::(\d+))?(/.*)?$)");
```
denise-opendaq, line 246: auto regexIpv6Hostname = std::regex(R"(^(.*://)?(?:\[([a-fA-F0-9:]+(?:\%\w+)?)\])(?::(\d+))?(/.*)?$)");

JakaMohorkoDS: Missed that one.. Good catch.

PR #467, opened 2024-08-26
[BUG]: sync component  backwards compatibility. Removing the device in gui demo

denise-opendaq:
## Type of change

# Checklist:

The new version of client could not connect to the device without sync component as it was not found. The exception was replaced with warning. In gui demo the device was registered in dictionary with the typed connection string. If connection string prefix is "daq://" the device automatically chooses the optimal way of connection and stores not the connection string with "daq://" but with the preferred protocol which makes impossible to remove device as the connection string is different

### DIFF
--- a/shared/libraries/config_protocol/include/config_protocol/config_client_device_impl.h
+++ b/shared/libraries/config_protocol/include/config_protocol/config_client_device_impl.h
@@ -219,7 +219,7 @@ void GenericConfigClientDeviceImpl<TDeviceBase>::onRemoteUpdate(const Serialized
         {
             if (this->defaultComponents.count(id))
             {
-                LOG_W("The server does not provide default device component: {}", id);
+                DAQLOGF_W(this->loggerComponent, "The server does not provide default device component: {}", id);
             }
             else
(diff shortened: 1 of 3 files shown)


shared/libraries/config_protocol/include/config_protocol/config_client_device_impl.h
```diff
@@ -218,9 +218,13 @@ void GenericConfigClientDeviceImpl<TDeviceBase>::onRemoteUpdate(const Serialized
         if (!serialized.hasKey(id))
         {
             if (this->defaultComponents.count(id))
-                throw InvalidOperationException{"Serialized remote object does not contain default device component: " + id};
-            
-            toRemove.push_back(id);
+            {
+                DAQLOGF_W(this->loggerComponent, "The server does not provide default device component: {}", id);
```
JakaMohorkoDS, line 222: I don't think we should be outputting a warning for every device that was created with an older version of openDAQ. This is likely a debug message.

JakaMohorkoDS (approved):

PR #472, opened 2024-08-28
<TBBAS-1490> Add native streaming access control

janm-dw:


JakaMohorkoDS (commented):
Lgtm, but @NikolaiShipilov should double check the streaming handler code.

shared/libraries/native_streaming_protocol/src/server_session_handler.cpp
```diff
@@ -138,8 +146,19 @@ ReadTask ServerSessionHandler::readSignalSubscribe(const void* data, size_t size
...
+        const auto hasAccess = hasUserAccessToSignal(signal);
+
+        if (hasAccess && signalSubscriptionHandler(signalNumericId, signal, true, clientId))
+            sendSubscribingDone(signalNumericId);
+    }
+    catch (const NativeStreamingProtocolException& e)
+    {
+        LOG_W("Protocol warning: {}", e.what());
+    }
+
```
NikolaiShipilov, line 161: For me it seems simpler to place the access verifications at a higher level in `NativeStreamingServerHandler::handleSignalSubscription`, without needing to modify the `OnSignalSubscriptionCallback` signature or introduce an additional `OnFindSignalCallback`.

shared/libraries/native_streaming_protocol/src/native_streaming_server_handler.cpp
```diff
@@ -116,11 +116,12 @@ void NativeStreamingServerHandler::removeComponentSignals(const StringPtr& compo
...
-                                                            const std::string& signalStringId,
+                                                            const SignalPtr& signal,
                                                             bool subscribe,
                                                             const std::string& clientId)
 {
     std::scoped_lock lock(sync);
+    const auto signalStringId = signal.getGlobalId();
 
     if (subscribe)
     {
```
NikolaiShipilov, line 127: I would place the signal permissions check here instead of in the session handler impl, like this:
```
...
   try
   {
      if (!sessionHandlers.at(clientId)->hasUserAccessToSignal(streamingManager.findRegisteredSignal(signalStringId)))
         throw NativeStreamingProtocolException("Access denied");
      ...
   }
```

NikolaiShipilov (commented):
Overall, looks good to me; I just suggested a small optimization.

However, there's a consideration regarding adding and removing signal access rights for a user after the client has already connected. In addition to merely ignoring subscribe/unsubscribe requests from a client that has lost read authorization, a better approach would be to also send signal available/unavailable messages to the client authenticated with the specific username, based on the permissions that have been added or removed for that user.
This could probably be handled through the core event mechanism. Is this something that's already planned, or?

modules/tests/test_opendaq_device_modules/test_native_streaming_modules.cpp
```diff
@@ -607,3 +628,119 @@ TEST_F(NativeStreamingModulesTest, GetConfigurationConnectionInfo)
...
+    serverSignal.getPermissionManager().setPermissions(permissionsDenyAll);
+
+    // try to unsibscibe
+    std::promise<StringPtr> signalUnsubscribePromise;
+    std::future<StringPtr> signalUnsubscribeFuture;
+    test_helpers::setupUnsubscribeAckHandler(signalUnsubscribePromise, signalUnsubscribeFuture, signal);
+
+    reader.release();
+    ASSERT_FALSE(test_helpers::waitForAcknowledgement(signalUnsubscribeFuture, std::chrono::seconds(1)));
+}
```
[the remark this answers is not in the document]
janm-dw: I don't think sending a signal available event if user access was revoked after connect is even needed. At least not for now. The only reason, I am revoking read access on the signal after connect is because I want to test if the subscribe method itself is protected. If I were to deny a read access before connect, the client would not even see the signal, so testing the subscribe through standard interface would be impossible.

shared/libraries/native_streaming_protocol/src/native_streaming_server_handler.cpp
janm-dw: I was thinking about this approach, however it seems cleaner to me that all access protection is done in session handler. 

Otherwise I would argue that even the if from sendSignalAvailable() and sendSignalUnavailable() would need to be moved to the streaming server handler. Someone might to forget to wrap one function call leading to exposed signal. Anyone Implementing OnSignalSubscriptionCallback must also not forget to protect it with read access. It seems much cleaner to me that the callback is not even triggered if the user does not have access.

NikolaiShipilov (approved):

PR #474, opened 2024-08-29
Fix rare native streaming client crash on reconnect

dejancrn:


NikolaiShipilov (approved):

dejancrn:
Added more checks. Should the same thing be done on ServerSessionHandler?

JakaMohorkoDS (approved):

PR #475, opened 2024-09-02
<TBBAS-1772> Add support for bcrypt encoded passwords to authenticati…

janm-dw:


JakaMohorkoDS (approved):
Some Doxygen comments should be added indicating how a bcrypt has can be used as a password. Additionally, this should be used in an example.

PR #477, opened 2024-09-03
Add struct decoder function block

dejancrn:
Adds a struct decoder function block that decomposes the struct signal into sub-signals. Signals are created when the data descriptor changed event is handled.

Tested with Python GUI demo and CAN channel; unit tests are still missing.

Update 1:

Added unit tests

### DIFF
--- a/modules/ref_fb_module/src/struct_decoder_fb_impl.cpp
+++ b/modules/ref_fb_module/src/struct_decoder_fb_impl.cpp
@@ -1,0 +1,292 @@
+#include <ref_fb_module/struct_decoder_fb_impl.h>
+#include <opendaq/input_port_factory.h>
+#include <opendaq/data_descriptor_ptr.h>
+#include <opendaq/event_packet_ptr.h>
+#include <opendaq/custom_log.h>
+#include <opendaq/event_packet_params.h>
+#include <opendaq/data_packet.h>
+#include <opendaq/data_packet_ptr.h>
+#include <opendaq/event_packet_ids.h>
+#include <opendaq/packet_factory.h>
+#include <opendaq/component_status_container_private_ptr.h>
+#include <opendaq/sample_type_traits.h>
+#include <opendaq/search_filter_factory.h>
+
+BEGIN_NAMESPACE_REF_FB_MODULE
+namespace StructDecoder
+{
+
+static const char* InputDisconnected = "Disconnected";
+static const char* InputConnected = "Connected";
+static const char* InputInvalid = "Invalid";
+
+StructDecoderFbImpl::StructDecoderFbImpl(const ContextPtr& ctx, const ComponentPtr& parent, const StringPtr& localId)
+    : FunctionBlock(CreateType(), ctx, parent, localId)
+    , configured(false)
+{
+    createInputPorts();
+    initStatuses();
+}
+
+FunctionBlockTypePtr StructDecoderFbImpl::CreateType()
+{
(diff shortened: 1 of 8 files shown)


modules/ref_fb_module/src/struct_decoder_fb_impl.cpp
```diff
@@ -0,0 +1,289 @@
...
+            SampleType::UInt64,
+            SampleType::Float32,
+            SampleType::Float64,
+            SampleType::Struct,
+        };
+
+        const auto domainSignal = createAndAddSignal("__domain", inputDomainDataDescriptor);
+        domainSignal.asPtr<IComponentPrivate>().unlockAttributes(List<IString>("Visible"));
+        domainSignal.setVisible(false);
+        domainSignal.asPtr<IComponentPrivate>().lockAttributes(List<IString>("Visible"));
```
JakaMohorkoDS, line 82: I think there's an optional parameter in the createAndAddSignal for visibility.

modules/ref_fb_module/src/struct_decoder_fb_impl.cpp
```diff
@@ -0,0 +1,289 @@
...
+        const auto structFields = inputDataDescriptor.getStructFields();
+        for (const auto& field: structFields)
+        {
+            const auto fieldSampleType = field.getSampleType();
+            if (std::find(validFieldTypes.begin(), validFieldTypes.end(), fieldSampleType) == validFieldTypes.end())
+                throw std::runtime_error(fmt::format("Field \"{}\" has invalid sample type", field.getName()));
+
+            const auto signal = createAndAddSignal(field.getName(), field);
+            signal.setDomainSignal(domainSignal);
+        }
```
JakaMohorkoDS, line 93: This should probably be recursive for nested struct-type signals, unless this fb should only decode 1 layer of the structs. Similar comments apply to the data processing.

modules/ref_fb_module/src/struct_decoder_fb_impl.cpp
```diff
@@ -0,0 +1,289 @@
...
+    catch (const std::exception& e)
+    {
+        setInputStatus(InputInvalid);
+        LOG_W("Failed to configure output signals: {}", e.what())
+        signals.clear();
+    }
+}
+
+
+void StructDecoderFbImpl::onPacketReceived(const InputPortPtr& port)
```
JakaMohorkoDS, line 108: Should our examples use readers? This fb could also be written with a stream reader.

modules/ref_fb_module/src/struct_decoder_fb_impl.cpp
dejancrn: Ok will check that

modules/ref_fb_module/src/struct_decoder_fb_impl.cpp
dejancrn: It should decode just one layer and then additional decoder would be needed. I want to keep the implementation simple. Structs are anyway rare and nested structs even more :)

modules/ref_fb_module/src/struct_decoder_fb_impl.cpp
dejancrn: I want to keep the same packet size/sample count on output as on input. I am not sure if this is guaranteed with stream reader. Also stream reader does not support skipping samples when input is not configured, does it?

JakaMohorkoDS: No, not yet. That code is in a PR at the moment.

JakaMohorkoDS (approved):

PR #488, opened 2024-09-09
Fix object permissions without owner

JakaMohorkoDS:
## Type of change

# Description:

Property objects without parents had no access by default. Any type objects getters (eg. getAvailableFunctionBlockType) were crashing over native due to it.

# TODO:
- Add test

### DIFF
--- a/core/opendaq/component/include/opendaq/component_impl.h
+++ b/core/opendaq/component/include/opendaq/component_impl.h
@@ -238,13 +238,9 @@ ComponentImpl<Intf, Intfs...>::ComponentImpl(
 
     if (parent.assigned())
     {
+        this->permissionManager.setPermissions(PermissionsBuilder().inherit(true).build());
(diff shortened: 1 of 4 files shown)


core/opendaq/component/include/opendaq/component_impl.h
```diff
@@ -238,14 +238,10 @@ ComponentImpl<Intf, Intfs...>::ComponentImpl(
 
     if (parent.assigned())
     {
+        this->permissionManager.setPermissions(PermissionsBuilder().inherit(true).build());
```
janm-dw, line 241: I am not sure, this line is needed. I think permission manager inherits by default.

janm-dw: On second thought, it is safer to leave that line here.

janm-dw (approved):

PR #489, opened 2024-09-09
Make data packet buffers accessible in python bindings

dejancrn:
Add support for packet creation in python bindings.
Add test for packet creation in python bindings.


Yarikk26 (approved):
Looks ok for me

JakaMohorkoDS (approved):

PR #496, opened 2024-09-11
Retain the original value while converting Ordinal Object to Bool

sedovmo:
## Type of change

# Checklist:

# Description:
Assigning the result of getPropertyValue() to `uint8_t can be dangerous, because it implicitly casts to daq::Bool and converts the original value to either 0 or 1.
This is because daq::Bool is uint8_t by definition. During the conversion to Bool, the original value is replaced by daq::True or daq::False.
The suggestion is to keep the original during the conversion.

```cpp
uint8_t value = 33;
objPtr.setPropertyValue("StreamId", value);
uint16_t uint16Val = obj.getPropertyValue("StreamId");
ASSERT_EQ(int16Val, value); // pass
uint8_t uint8Val = obj.getPropertyValue("StreamId");
ASSERT_EQ(int8Val, value); // fail
```

### DIFF
--- a/core/coretypes/tests/test_integer.cpp
+++ b/core/coretypes/tests/test_integer.cpp
@@ -85,6 +85,14 @@ TEST_F(IntegerTest, CastBool)
     ASSERT_EQ(valBool, True);
 }
 
+TEST_F(IntegerTest, CastUint8_t)
+{
+    auto intObj = Integer(2);
+
+    uint8_t valUint8 = intObj;
+    ASSERT_EQ(valUint8, 2);
+}
+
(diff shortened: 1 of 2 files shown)


JakaMohorkoDS (approved):

dejancrn (changes requested):
Looks ok, but the test for this would look something like this:

```
TEST_F(IntegerTest, CastUInt8_t)
{
    auto intObj = Integer(2);

    uint8_t valInt = intObj;
    ASSERT_EQ(valInt, 2);
}
```

and it should be part of coretypes tests test_integer.cpp

dejancrn (approved):

PR #502, opened 2024-09-12
Add python example to enable native device & streaming monitoring

dejancrn:
Adds example how to enable device monitoring for reconnection on native configuration & streaming

### DIFF
--- a/examples/python/enable_device_monitoring.py
+++ b/examples/python/enable_device_monitoring.py
@@ -1,0 +1,40 @@
+from gc import enable
+import opendaq
+import time
+
+def set_device_config(config: opendaq.IPropertyObject):
+    
+    def enable_monitoring_on_transport_layer(config: opendaq.IPropertyObject):
+        transport_layer_config = opendaq.IPropertyObject.cast_from(config.get_property_value("TransportLayerConfig"))
+        transport_layer_config.set_property_value("MonitoringEnabled", True)
+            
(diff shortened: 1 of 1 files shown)


examples/python/enable_device_monitoring.py
```diff
@@ -0,0 +1,40 @@
...
+    native_device_config = opendaq.IPropertyObject.cast_from(device_config.get_property_value("OpenDAQNativeConfiguration"))
+    enable_monitoring_on_transport_layer(native_device_config)
+
+def main():
+    instance = opendaq.Instance()
+    
+    device_config = instance.create_default_add_device_config()
+    set_device_config(device_config)
+
+    device = instance.add_device("daq.nd://192.168.10.1", device_config)
```
NikolaiShipilov, line 25: It might be better to pass the IP address as a parameter to the py script

NikolaiShipilov (approved):

examples/python/enable_device_monitoring.py
dejancrn: It was agreed that for short examples we would just add localhost address and a comment as a reminder to change the address. This is just an example and not a tool, where it would make sense to have address as a command line argument.

But yes, the 192.168.10.1 is definitely not ok, I forgot about that. Thanks.

JakaMohorkoDS (approved):

Yarikk26 (approved):

PR #508, opened 2024-09-16
Improve thread-safety of config protocol client

NikolaiShipilov:
# Description:

* Ensure thread-safe generation of unique ID for config request
* Instantiate a separate deserializer for parsing each config request reply

## Type of change

# Checklist:

### DIFF
--- a/shared/libraries/config_protocol/include/config_protocol/config_protocol_client.h
+++ b/shared/libraries/config_protocol/include/config_protocol/config_protocol_client.h
@@ -96,11 +96,10 @@ public:
 
 private:
     ContextPtr daqContext;
-    uint64_t id;
+    std::atomic<uint64_t> id;
     SendRequestCallback sendRequestCallback;
     SendNoReplyRequestCallback sendNoReplyRequestCallback;
     ComponentDeserializeCallback rootDeviceDeserializeCallback;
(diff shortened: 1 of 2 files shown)


dejancrn (commented):
All the changes make sense, but I doubt this will be enough. For example when we do a configuration request, the `NativeDeviceHelper::doConfigRequestAndGetReply` callback is called. What happens if two threads do a configuration request at the same time.

NikolaiShipilov:
> All the changes make sense, but I doubt this will be enough. For example when we do a configuration request, the `NativeDeviceHelper::doConfigRequestAndGetReply` callback is called. What happens if two threads do a configuration request at the same time.

In this case each thread will receive a reply and process it using `ConfigProtocolClientComm::parseRpcReplyPacketBuffer` with independent deserializers. The request & reply registration by unique ID and future/promise instantiation within `NativeDeviceHelper` were already thread-safe. 
Therefore, from my perspective, it's now should be safe to perform multiple config requests simultaneously. However, I'm not certain if it's safe to apply the reply results if both affect the same components of the tree.

dejancrn (approved):
Ok then this looks fine. As of applying the two simultaneous requests from different threads may or may not be an issue. I'm not sure yet. But this PR can be merged

PR #518, opened 2024-09-19
[TBBAS-1825] Fix device username/location properties case in opcua

denise-opendaq:
## Type of change

# Checklist:

# Description:
device properties "userName" and "location" are shown in Opc Ua as "UserName" and "Location". Opendaq client can access to server properties with both ways (this made for backwards compatibility and testing)


ml-hbk (commented):
This change looks a bit strange to me. We aren't simply the string constants being fixed.

ml-hbk (commented):
Hi Denis,

could you please reduce the pull request to the fix only. There are 129 files that have changed here.

shared/libraries/opcuatms/opcuatms_server/src/objects/tms_server_property.cpp
```diff
@@ -59,16 +62,21 @@ std::string TmsServerProperty::getBrowseName()
     return this->object.getName();
 }
 
+std::string TmsServerProperty::getNodeName()
```
janm-dw, line 65: Can you check if we can avoid adding this method. Maybe we could try to override the existing getBrowseName() method.

janm-dw (approved):

PR #531, opened 2024-10-03
Fix enumeration factory with integer parameter

JakaMohorkoDS:
# Type of change:

# Description:

Enumeration construction with an integer parameter failed due to a type (`break` instead of `return`).


aljazfrancic (approved):
Seems legitimate.

PR #532, opened 2024-10-04
Other/min read count GitHub

dejancrn:
* Adds option to define minimum read count for multi reader.


core/opendaq/reader/src/multi_reader_impl.cpp
```diff
@@ -1270,24 +1311,17 @@ OPENDAQ_DEFINE_CLASS_FACTORY(LIBRARY_FACTORY,
...
+    LIBRARY_FACTORY, MultiReaderImpl, IMultiReader, createMultiReaderEx,
+    IList*, signals,
+    SampleType, valueReadType,
+    SampleType, domainReadType,
+    ReadMode, mode,
+    ReadTimeoutType, timeoutType,
+    Int, requiredCommonSampleRate,
+    Bool, startOnFullUnitOfDomain,
+    SizeT, minReadCount
+)
```
JakaMohorkoDS, line 1324: Do we need to modify this factory, or is the builder sufficient?

core/opendaq/reader/src/multi_reader_impl.cpp
```diff
@@ -601,6 +615,9 @@ ErrCode MultiReaderImpl::skipSamples(SizeT* count, IMultiReaderStatus** status)
         return OPENDAQ_IGNORED;
     }
 
+    if (minReadCount > *count)
+        return makeErrorInfo(OPENDAQ_ERR_INVALIDPARAMETER, "Count parameter has to be larger than minReadCount.");
```
JakaMohorkoDS, line 619: Should we throw an exception in these cases, or simply return that we read no samples?

JakaMohorkoDS (approved):

PR #536, opened 2024-10-09
 Add an option to mark functions and procedures as const

janm-dw:
-  Add an option to mark functions and procedures as const
- Allow triggering of const function properties under locked device over native config protocol

### DIFF
--- a/core/coreobjects/include/coreobjects/callable_info.h
+++ b/core/coreobjects/include/coreobjects/callable_info.h
@@ -72,6 +72,8 @@ DECLARE_OPENDAQ_INTERFACE(ICallableInfo, IBaseObject)
  * @brief Creates a CallableInfo object with the specified arguments and return type.
  * @param argumentInfo The list of `ArgumentInfo` type argument information.
  * @param returnType The return type of the described callable object.
+ * @param constFlag A flag indicating if the function is marked as const or not. A const function promises not to modify the state
+ * of the device or any other objects under the openDAQ instance.
  */
 OPENDAQ_DECLARE_CLASS_FACTORY(
     LIBRARY_FACTORY, CallableInfo,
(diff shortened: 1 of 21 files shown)


core/coreobjects/include/coreobjects/callable_info_impl.h
```diff
@@ -27,10 +27,11 @@ BEGIN_NAMESPACE_OPENDAQ
 class CallableInfoImpl : public GenericStructImpl<ICallableInfo, IStruct>
```
JakaMohorkoDS, line 27: This is a small-ish issue, as we're redefining a struct type here. I'm not 100% sure if these types are ever added to the type manager (I don't think they are), so it's not too big a deal. 

Notably, the struct type for CallableInfo must be updated to include isConst.

core/coreobjects/include/coreobjects/callable_info.h
```diff
@@ -68,7 +76,8 @@ DECLARE_OPENDAQ_INTERFACE(ICallableInfo, IBaseObject)
 OPENDAQ_DECLARE_CLASS_FACTORY(
     LIBRARY_FACTORY, CallableInfo,
     IList*, argumentInfo,
-    CoreType, returnType
+    CoreType, returnType,
```
JakaMohorkoDS, line 79: I would consider making a new factory here `CallableInfoWithConst` to preserve backwards compatibility. It's not 100% mandatory, as I don't think anyone is creating openDAQ properties in any bindings. 

At the very least the Integration notes of the changelog should notify that a factory signature got changed and needs to be fixed in all modules that use the C factory.

janm-dw: I have added a note to changelog file.

JakaMohorkoDS (approved):

dejancrn (approved):

PR #540, opened 2024-10-10
Other/property object sync

JakaMohorkoDS:
# Type of change:

# Checklist:

# Description:

- Introduces synchronization primitives to the property object implementation, making external API calls thread-safe.
- Introduces a recursive lock guard object for configuration call locking that's built on top a standard mutex

# Breaking changes:

- The `sync` mutex previously available in component implementations can no longer be locked normally within an event (onWrite, onRead...), nor does it need to be, as it's locked internally in the Property object implementation.
- If locking is needed (the function might be called from another thread), the `getRecursiveConfigLock()` method should be used instead.

### DIFF
--- a/core/coreobjects/include/coreobjects/property_object_impl.h
+++ b/core/coreobjects/include/coreobjects/property_object_impl.h
@@ -46,6 +46,7 @@
 #include <coreobjects/permission_manager_internal_ptr.h>
 #include <coreobjects/permission_mask_builder_factory.h>
 #include <coreobjects/permissions_builder_factory.h>
+#include <thread>
 
 BEGIN_NAMESPACE_OPENDAQ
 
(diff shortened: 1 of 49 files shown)


JakaMohorkoDS:
Should this PR be merged before the 3.10 release?

core/coreobjects/include/coreobjects/property_object_impl.h
```diff
@@ -57,6 +58,48 @@ struct PropertyNameInfo
...
+
+template <typename TMutex>
+class GenericRecursiveConfigLockGuard : public RecursiveConfigLockGuard
+{
+public:
+    GenericRecursiveConfigLockGuard(TMutex* lock, std::thread::id* threadId, int* depth)
+        : id(threadId)
+        , depth(depth)
+        , lock(std::lock_guard(*lock))
+    {
```
dejancrn, line 85: I would insert some assert statements that parameters are not nullptr.

dejancrn (approved):

PR #550, opened 2024-10-16
Inject openDAQ version from a single file

aljazfrancic:
# Type of change:

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

# Checklist:

- [x] Pull request title reflects its content
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes

# Description:
Inject openDAQ version string into multiple files throughout CI/repo so only one file needs to be edited during version bump.

PR #554, opened 2024-10-18
openDAQ versioning fixes

aljazfrancic:
* Rename placeholder because it is shown in https://opendaq.github.io/
* Showing the version as `latest_dev_opendaq_version` is more correct than showing future version

### DIFF
--- a/docs/Antora/modules/getting_started/pages/quick_start_setting_up_cpp.adoc
+++ b/docs/Antora/modules/getting_started/pages/quick_start_setting_up_cpp.adoc
@@ -7,7 +7,7 @@ To start working with openDAQ(TM), the requisite binaries are required. They can
 Windows::
 +
 --
-To install openDAQ(TM), download the `opendaq-latest_dev_opendaq_version-win64.exe` installer, run it, and follow its instructions. Make sure to add openDAQ(TM) to the system path to be able to run openDAQ(TM) application examples.
+To install openDAQ(TM), download the `opendaq-:version:-win64.exe` installer, run it, and follow its instructions. Make sure to add openDAQ(TM) to the system path to be able to run openDAQ(TM) application examples.
 
 We recommend using the default installation path to avoid issues requiring additional environment variable modifications and troubleshooting.
 --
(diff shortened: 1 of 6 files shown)


docs/Antora/modules/getting_started/pages/quick_start_setting_up_cpp.adoc
```diff
@@ -7,19 +7,19 @@ To start working with openDAQ(TM), the requisite binaries are required. They can
 Windows::
 +
 --
-To install openDAQ(TM), download the `opendaq-opendaq_version-win64.exe` installer, run it, and follow its instructions. Make sure to add openDAQ(TM) to the system path to be able to run openDAQ(TM) application examples.
+To install openDAQ(TM), download the `opendaq-latest_dev_opendaq_version-win64.exe` installer, run it, and follow its instructions. Make sure to add openDAQ(TM) to the system path to be able to run openDAQ(TM) application examples.
```
denise-opendaq, line 10: just a note: maybe it would be better to use `{opendaq_version}` or something like that so devs would see that its template argument?

aljazfrancic: Yes, but there are still at least other two problems with my previous PR (https://github.com/openDAQ/openDAQ/pull/550):
* In examples we use `set(OPENDAQ_EXAMPLES_VERSION ${OPENDAQ_PACKAGE_VERSION})` (https://github.com/openDAQ/openDAQ/blob/main/examples/CMakeLists.txt) which means you can't build the examples (which are also available as a standalone download on https://docs-dev.opendaq.com/) without building the whole project
* https://opendaq.github.io/ will show the placeholders either way

So I'm thinking we'll have to revert these changes... Because if you look at the `antora-playbook.yml` for that docs page I see no way to circumvent hard-coded strings (https://github.com/openDAQ/opendaq.github.io/blob/master/antora-playbook.yml)

denise-opendaq: i see in the official site that ascii doc supports Attributes. maybe it will help you [Attributes and substitutions](https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#attributes-and-substitutions)

aljazfrancic: This is true, but `config.adoc` would still have to be hard-coded, correct?

denise-opendaq: well, docs/version_injector.py can then modify only this file :)

aljazfrancic: True, I will change it like you suggested. But this wont work for: https://opendaq.github.io/

denise-opendaq (approved):

euttsov-odaq (approved):

PR #560, opened 2024-10-23
<TBBAS-1845> mitigate out-of-memory risk on server-side for native streaming

NikolaiShipilov:
Related PR in libNative - https://github.com/openDAQ/libNativeStreaming/pull/18

# Description:
Introduce an optional timeout for streaming packet writes:

* Add the "StreamingPacketSendTimeout" property to the native server for specifying the timeout value.
* Enable "PacketStreamingServer" to optionally add a timestamp to serialized openDAQ packets at "PacketBuffer" creation.
* Calculate the deadline for the start time point of the packet write operation using the serialized packet's timestamp and the specified timeout.
* Server resets the client connection if the deadline is exceeded.

# Type of change:

# Checklist:

### DIFF
--- a/shared/libraries/native_streaming_protocol/src/base_session_handler.cpp
+++ b/shared/libraries/native_streaming_protocol/src/base_session_handler.cpp
@@ -10,7 +10,8 @@ BaseSessionHandler::BaseSessionHandler(const ContextPtr& daqContext,
                                        SessionPtr session,
                                        const std::shared_ptr<boost::asio::io_context>& ioContextPtr,
                                        OnSessionErrorCallback errorHandler,
-                                       ConstCharPtr loggerComponentName)
+                                       ConstCharPtr loggerComponentName,
+                                       SizeT streamingPacketSendTimeout)
     : session(session)
     , configPacketReceivedHandler(nullptr)
     , packetBufferReceivedHandler(nullptr)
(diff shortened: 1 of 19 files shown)


shared/libraries/native_streaming_protocol/src/base_session_handler.cpp
```diff
@@ -358,7 +363,12 @@ void BaseSessionHandler::sendPacketBuffer(PacketBufferPtr&& packetBuffer)
     auto writeHeaderTask = createWriteHeaderTask(PayloadType::PAYLOAD_TYPE_STREAMING_PACKET, payloadSize);
     tasks.insert(tasks.begin(), writeHeaderTask);
 
-    session->scheduleWrite(std::move(tasks));
+    auto deadlineTime =
+        packetBuffer->timeStamp.has_value() && streamingPacketSendTimeout != std::chrono::milliseconds(0)
```
dejancrn, line 367: if streamingPacketSendTimeout == 0, will packet buffer contain the timestamp?

dejancrn (approved):

shared/libraries/native_streaming_protocol/src/base_session_handler.cpp
NikolaiShipilov: Currently, one of the two conditions is always redundant, as both are true when the feature is enabled and false when it’s disabled. 
However, since both conditions must be met to correctly set a deadline, I would keep both for now. This ensures compatibility if changes are made, such as if certain packet types, like event packets, will not have an attached timestamp.

PR #570, opened 2024-10-30
<TBBAS-1407> Add support for serialization of list properties on prop…

janm-dw:


core/coreobjects/include/coreobjects/property_object_impl.h
```diff
@@ -703,7 +703,8 @@ ErrCode GenericPropertyObjectImpl<PropObjInterface, Interfaces...>::checkContain
     if (coreType == ctList)
     {
         const auto itemType = prop.getItemType();
-        if (!iterate(value, itemType))
+
+        if (itemType != ctUndefined && !iterate(value, itemType))
```
janm-dw, line 707: I could also explicitly check if the list is empty, but this seems enough

JakaMohorkoDS (approved):

PR #579, opened 2024-11-05
Consolidate protocol specs and user guide into a single documentation site

NikolaiShipilov:
A separate archive for Protocol specifications is no longer created. 
Protocol specifications are now accessible from the drop-up menu in the bottom left corner of the documentation site.


NikolaiShipilov:
Thought, not sure regarding offline version of the documentation, as archive file is labeled `user_guide` but now also includes specifications.

JakaMohorkoDS (approved):

PR #580, opened 2024-11-06
<TBBAS-1900> Update locked status through core event

janm-dw:


core/opendaq/device/include/opendaq/device_impl.h
```diff
@@ -1581,6 +1619,17 @@ void GenericDevice<TInterface, Interfaces...>::deserializeCustomObjectValues(con
...
 
+
+    userLock.reset();
+
+    if (serializedObject.hasKey("Locked"))
+    {
+        const bool isLocked = serializedObject.readBool("Locked");
+
+        if (isLocked)
+            userLock = nullptr;
```
janm-dw, line 1630: Right now if we serialize and deserialize device, we preserve lock status but loose the user who locked the device.

On the client side this is not a problem because we only care about current lock status of the device. The user validation is done on the server side when Lock / Unlock RPC is triggered.

It should be possible to preserve the user if we serialize the username and retrieve the user instance from authentication provider in the deserialize method. For this, a new method IAuthenticationProvider::getUser(IString username) should be added. IUserLock can also be also added for nicer serialization.

core/coreobjects/include/coreobjects/authentication_provider.h
```diff
@@ -59,6 +59,14 @@ DECLARE_OPENDAQ_INTERFACE(IAuthenticationProvider, IBaseObject)
...
      */
     virtual ErrCode INTERFACE_FUNC authenticateAnonymous(IUser** userOut) = 0;
+
+    /*!
+     * @brief Find a user instance by its username. If no user with maching username is not found, null is returned.
+     *
+     * @param username The username.
+     * @param userOut[out] Found user instance. If no user with maching username is not found, null is returned.
+     */
+    virtual ErrCode INTERFACE_FUNC findUser(IString* username, IUser** userOut) = 0;
```
JakaMohorkoDS, line 69: Missing from changelog.

core/opendaq/device/include/opendaq/user_lock.h
```diff
@@ -0,0 +1,75 @@
...
+ * @ingroup objects_security
+ * @addtogroup objects_security_user UserLock
+ * @{
+ */
+
+/*!
+ * @brief This class manages the lock state of an object, usually a device. A device can be locked either with a specific user or without one.
+ * If the device is locked with a specific user, only that user can unlock it. If the device is locked without a user, any user can unlock it.
+ */
+DECLARE_OPENDAQ_INTERFACE(IUserLock, IBaseObject)
```
JakaMohorkoDS, line 37: Missing from changelog.

core/opendaq/device/include/opendaq/device_impl.h
```diff
@@ -375,6 +385,44 @@ ErrCode GenericDevice<TInterface, Interfaces...>::unlock(IUser* user)
...
+    *locked = this->userLock.isLocked();
+    return OPENDAQ_SUCCESS;
+}
+
+template <typename TInterface, typename... Interfaces>
+ErrCode GenericDevice<TInterface, Interfaces...>::forceUnlock()
+{
+    std::scoped_lock syncLock(this->sync);
+
+    ErrCode status = forceUnlockInternal();
```
JakaMohorkoDS, line 407: Should we return here if device is already unlocked?

shared/libraries/config_protocol/include/config_protocol/config_client_device_impl.h
```diff
@@ -212,11 +213,9 @@ ErrCode INTERFACE_FUNC GenericConfigClientDeviceImpl<TDeviceBase>::unlock(IUser*
 }
 
 template <class TDeviceBase>
-inline ErrCode INTERFACE_FUNC GenericConfigClientDeviceImpl<TDeviceBase>::isLocked(Bool* locked)
+inline ErrCode INTERFACE_FUNC GenericConfigClientDeviceImpl<TDeviceBase>::forceUnlock()
 {
-    OPENDAQ_PARAM_NOT_NULL(locked);
-
-    return daqTry([this, &locked] { *locked = this->clientComm->isLocked(this->remoteGlobalId); });
```
JakaMohorkoDS, line 219: This is probably fine to remove, as there has been no releases in between the 1st locking implementation and this one. It does break compatibility, however.

JakaMohorkoDS (approved):

core/opendaq/device/include/opendaq/device_impl.h
janm-dw: I would say no. Even if device is unlocked it might still have some locked children which you want to unlock. In this case you can call unlock on the root device and you can be sure that after the call, every device will be unlocked.

PR #582, opened 2024-11-06
Allow to disable flaky tests

NikolaiShipilov:
# Description:
* Introduces mechanism to skip "flaky" failing tests
* Makes majority of failing test listed in TBBAS-1735 skipped within regular CI test jobs

### DIFF
--- a/CMakeBasePresets.json
+++ b/CMakeBasePresets.json
@@ -158,7 +158,9 @@
                 "OPENDAQ_GENERATE_CSHARP_BINDINGS": true,
                 "APP_ENABLE_AUDIO_APP": true,
                 "APP_ENABLE_EXAMPLE_APPS": true,
-                "DAQSIMULATOR_ENABLE_SIMULATOR_APP": true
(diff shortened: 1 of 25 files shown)


aljazfrancic (approved):
Looks good to me! Additionally, we could also add a daily GitHub Action that runs flaky tests? Maybe that's overkill, though.

JakaMohorkoDS:
It'd be good to also add a scheduled job with these tests enabled that runs nightly on main and release/*  branches.

I'll make a note that from now on any flaky tests that fail independent of a PR should be wrapped with this flag.

JakaMohorkoDS (approved):

NikolaiShipilov:
> Looks good to me! Additionally, we could also add a daily GitHub Action that runs flaky tests? Maybe that's overkill, though.

> It'd be good to also add a scheduled job with these tests enabled that runs nightly on main and release/* branches.

The mechanism to skip flaky test is made with this end goal (CI job) in mind. The idea is to use prefix attached to tests name to filter those and run only flaky ones, likely multiple times in a row to ease debugging.

PR #584, opened 2024-11-08
Run flaky tests on nightly basis reusing existing CI jobs

NikolaiShipilov:
# Description:
Flaky test CI workflow reuse existing "Build and test openDAQ" workflow but with alternative cmake and ctest presets.
This workflow is to be run on main branch every night or can be manually triggered on any other branch


aljazfrancic (approved):
LGTM!

CMakeBasePresets.json
```diff
@@ -163,6 +163,17 @@
...
         },
+        {
+            "name": "flaky_tests_ci",
+            "hidden": false,
+            "inherits": [
+                "ci"
+            ],
+            "cacheVariables": {
+                "OPENDAQ_ENABLE_FLAKY_TEST_LABELS": true,
+                "OPENDAQ_SKIP_FLAKY_TESTS": false
```
JakaMohorkoDS, line 174: Also... while we're at it, could we rename this to "OPENDAQ_SKIP_UNSTABLE_TESTS".

.github/workflows/flaky_tests.yml
```diff
@@ -0,0 +1,14 @@
+name: Build and run flaky Tests openDAQ
```
JakaMohorkoDS, line 1: So, let's just name this something other  than "Flaky Tests" :) Something like, "Nightly full build and test".

JakaMohorkoDS (approved):

PR #606, opened 2024-11-20
Streaming connection statuses

NikolaiShipilov:
# Type of change:

# Checklist:

# Description:


core/opendaq/device/include/opendaq/connection_status_container_impl.h
```diff
@@ -0,0 +1,336 @@
...
+    OPENDAQ_PARAM_NOT_NULL(connectionString);
+    OPENDAQ_PARAM_NOT_NULL(initialValue);
+
+    const auto connectionStringObj = StringPtr::Borrow(connectionString);
+    if (connectionStringObj == "")
+        return OPENDAQ_ERR_INVALIDPARAMETER;
+
+    std::scoped_lock lock(sync);
+
+    if (statusNameAliases.hasKey(ConfigurationConnectionStatusAlias) ||
```
janm-dw, line 129: Is this condition correct? Shouldn't the keys of statusNameAliases be connection strings? So statusNameAliases.hasKey(ConfigurationConnectionStatusAlias) will probably always be false?

core/opendaq/device/include/opendaq/connection_status_container_impl.h
```diff
@@ -0,0 +1,336 @@
...
+    const auto connectionStringObj = StringPtr::Borrow(connectionString);
+    if (connectionStringObj == "")
+        return OPENDAQ_ERR_INVALIDPARAMETER;
+
+    std::scoped_lock lock(sync);
+
+    if (statuses.hasKey(connectionStringObj))
+        return OPENDAQ_ERR_ALREADYEXISTS;
+
+    ++streamingConnectionsCounter;
```
janm-dw, line 153: Counter is incremented in addStreamingConnectionStatus() but not decremented in removeStreamingConnectionStatus() is this intentional?

core/opendaq/device/include/opendaq/device_impl.h
```diff
@@ -1660,6 +1688,44 @@ void GenericDevice<TInterface, Interfaces...>::deserializeCustomObjectValues(con
     else
         userLock.forceUnlock();
 
+    if (serializedObject.hasKey("connectionStatuses"))
+    {
+        connectionStatusContainer = serializedObject.readObject(
+            "connectionStatuses",
+            context,
+            [this](const StringPtr& typeId,
```
janm-dw, line 1696: 1) Can you move this code above the deserialization of "UserLock" filed? It is probably safer if the user lock is the last field to be deserialized.

2) This looks like duplicated code to me. I suspect you are doing it this way, because you want to trigger the core event while deserializing? Would it be possible to use deserialize context parameter to pass the core event callback to ConnectionStatusContainerImpl::Deserialize method and have all the code there?

core/opendaq/device/include/opendaq/connection_status_container_impl.h
NikolaiShipilov: Yes, you're right - the condition was wrong.

core/opendaq/device/include/opendaq/connection_status_container_impl.h
NikolaiShipilov: Yes, the value is used to create numeration suffix in streaming status alias and it is not decremented intentionally to make sure any new alias has unique number.

core/opendaq/device/include/opendaq/device_impl.h
NikolaiShipilov: 1. Done. Could it be also unsafe that default folders and components are being deserialized after UserLock ?

2. Yes, the code in this deserialization method of device duplicates the one in ConnectionStatusContainer deserialization static method. Similarly, same approach is used for the ComponentStatusContainer and component Tags to pass the coreEvent callback into the constructors.
I think, it would be good to eliminate these duplications as well, though not sure if the deserialize context supports passing any type of parameter. Does it ?

core/opendaq/device/include/opendaq/connection_status_container_impl.h
```diff
@@ -0,0 +1,337 @@
...
+        return OPENDAQ_ERR_INVALIDPARAMETER;
+
+    std::scoped_lock lock(sync);
+
+    if (configConnectionStatusAdded || statuses.hasKey(connectionStringObj))
+        return OPENDAQ_ERR_ALREADYEXISTS;
+
+    statuses[connectionStringObj] = initialValue;
+    statusNameAliases[connectionStringObj] = ConfigurationConnectionStatusAlias;
+    configConnectionStatusAdded = true;
```
JakaMohorkoDS, line 135: Should this trigger a core event?

core/opendaq/device/include/opendaq/connection_status_container_impl.h
```diff
@@ -0,0 +1,337 @@
...
+    if (!statuses.hasKey(connectionString))
+        return OPENDAQ_ERR_NOTFOUND;
+
+    const StringPtr statusNameAlias =
+        statusNameAliases.hasKey(connectionString)
+            ? statusNameAliases.remove(connectionString)
+            : nullptr;
+
+    auto value = statuses.remove(connectionString);
+    value = "NotAvailable";
```
JakaMohorkoDS, line 189: I'd rename this to "Removed", as the status no longer actually exists in the container.

core/opendaq/device/include/opendaq/connection_status_container_private.h
```diff
@@ -0,0 +1,84 @@
...
+ * @brief Provides access to private methods for managing the Device's connection statuses.
+ *
+ * Enables adding, removing, and updating statuses stored in the connection status container.
+ * Statuses are identified by unique connection strings and can be accessed via
+ * IComponentStatusContainer using name aliases. A configuration status, accessible via alias "ConfigurationStatus",
+ * is unique per container and cannot be removed if added.
+ *
+ * On the other hand, multiple streaming statuses are supported, one per each streaming source attached to the device.
+ * Aliases for these follow the pattern "StreamingStatus_n," with optional protocol-based prefixes (e.g.,
+ * "OpenDAQNativeStreamingStatus_1", "OpenDAQLTStreamingStatus_2", "StreamingStatus_3" etc.).
```
JakaMohorkoDS, line 43: I'd maybe prefer "StreamingStatus_" to be the prefix for it to be more easily parseable. The '_' sign can be used as a delimeter. Eg. "StreamingStatus_OpenDAQNative_1".

Additionally, do we need the "_1" suffix?

core/opendaq/device/include/opendaq/connection_status_container_private.h
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * "ConnectionStatusChanged" Core events are triggered whenever there is a change in the connection status of the openDAQ Device,
+ * including parameters such as status name alias, value, protocol type, connection string, and streaming
+ * object (nullptr for configuration statuses). Removing streaming statuses also triggers said Core event with
+ * "NotAvailable" as the value and nullptr for the streaming object parameters.
```
JakaMohorkoDS, line 48: As above, I'd rename this to "Removed"

core/opendaq/device/include/opendaq/connection_status_container_private.h
```diff
@@ -0,0 +1,84 @@
...
+     */
+    virtual ErrCode INTERFACE_FUNC addConfigurationConnectionStatus(IString* connectionString, IEnumeration* initialValue) = 0;
+
+    /*!
+     * @brief Adds a new streaming connection status with the specified connection string, initial value, and streaming object.
+     * @param connectionString The connection string identifying the status.
+     * @param initialValue The initial value of the status.
+     * @param streamingObject The streaming object associated with the status.
+     */
+    virtual ErrCode INTERFACE_FUNC addStreamingConnectionStatus(IString* connectionString, IEnumeration* initialValue, IBaseObject* streamingObject) = 0;
```
JakaMohorkoDS, line 65: Is there a reason for the streaming object to not be `IStreaming*`?

modules/native_streaming_client_module/src/native_device_impl.cpp
```diff
@@ -482,33 +482,21 @@ NativeDeviceImpl::~NativeDeviceImpl()
...
-}
-
 // INativeDevicePrivate
-void NativeDeviceImpl::publishConnectionStatus(ConstCharPtr statusValue)
+void NativeDeviceImpl::publishConnectionStatus(const EnumerationPtr& status)
 {
-    auto newStatusValue = this->statusContainer.getStatus("ConnectionStatus");
-    newStatusValue = statusValue;
-
-    this->statusContainer.asPtr<IComponentStatusContainerPrivate>().setStatus("ConnectionStatus", newStatusValue);
```
JakaMohorkoDS, line 504: Should we maybe still keep this for backwards compatibility purposes? @dejancrn what are your thoughts on this?

core/opendaq/device/include/opendaq/connection_status_container_private.h
dejancrn: I think we can have multiple streamings per type.

dejancrn (approved):

core/opendaq/device/include/opendaq/connection_status_container_private.h
NikolaiShipilov: Initially, I thought this might require a dependency on the streaming core library, potentially causing cross-linking issues. However, it turns out not to be a problem since only a header include is needed. Therefore, it is now IStreaming.

modules/native_streaming_client_module/src/native_device_impl.cpp
NikolaiShipilov: Restored for backward compatibility

core/opendaq/device/include/opendaq/device_impl.h
NikolaiShipilov: 2. The ComponentDeserializationContext has been extended to support specifying a triggerCoreEvent procedure for component. It is now utilized to eliminate code duplication in the deserialization of status containers and tags.

core/opendaq/device/include/opendaq/connection_status_container_impl.h
```diff
@@ -0,0 +1,337 @@
+/*
```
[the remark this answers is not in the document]
NikolaiShipilov: The .cpp implementations is being built ok with GCC but cause an LNK2001 unresolved external symbol error with MSVC. This seems to be why the component status container and also tags were originally implemented directly in the header.

JakaMohorkoDS (approved):

PR #608, opened 2024-11-21
Fix a broken string in python gui_demo

adolenc:
At least in python 3.10 you get the following error at the start of `py -m opendaq`:
```
SyntaxError: unterminated string literal (detected at line 165)
```

Seems like python doesn't appreciate the line break there.


janm-dw:
Already fixes as part of #605

PR #625, opened 2024-12-02
Component status extension

aljazfrancic:
# Type of change:

# Checklist:

# Description:

* Add Component status types to the Type Manager in `context_impl.cpp` ("Ok", "Warning", and "Error")
* Define `ComponentErrorState`
* Declare and define `initComponentErrorStateStatus`, `setComponentErrorStateStatus`, and `setComponentErrorStateStatusWithMessage` in `component_impl.h`
* Add `getStatusMessage` in `component_status_container.h`, and `addStatusWithMessage`, `setStatusWithMessage` in `component_status_container_private.h` , implement all three in `component_status_container_impl.h`, and make other appropriate changes wherever necessary
* Use `initComponentErrorStateStatus` and `setComponentErrorStateStatusWithMessage` in all reference Function Blocks
* Add appropriate tests and supplement existing ones
* Modify Doxygen and Antora docs
* Fix typos, and other minor issues, etc. (unrelated)

### DIFF
--- a/core/coreobjects/include/coreobjects/core_event_args_impl.h
+++ b/core/coreobjects/include/coreobjects/core_event_args_impl.h
@@ -234,8 +234,6 @@ inline bool CoreEventArgsImpl::validateParameters() const
             return parameters.hasKey("DeviceDomain");
         case CoreEventId::DeviceLockStateChanged:
             return parameters.hasKey("IsLocked");
-        case CoreEventId::StatusChanged:
-            return parameters.getCount() == 2;
         case CoreEventId::ConnectionStatusChanged:
             return parameters.hasKey("StatusName")
                    && parameters.hasKey("StatusValue")
--- a/core/coreobjects/include/coreobjects/core_event_args_impl.h
+++ b/core/coreobjects/include/coreobjects/core_event_args_impl.h
@@ -235,7 +235,7 @@ inline bool CoreEventArgsImpl::validateParameters() const
         case CoreEventId::DeviceLockStateChanged:
             return parameters.hasKey("IsLocked");
         case CoreEventId::StatusChanged:
-            return parameters.getCount() == 1;
+            return parameters.getCount() == 2;
         case CoreEventId::ConnectionStatusChanged:
             return parameters.hasKey("StatusName")
                    && parameters.hasKey("StatusValue")
--- a/core/opendaq/component/include/opendaq/component.h
+++ b/core/opendaq/component/include/opendaq/component.h
@@ -39,13 +39,6 @@ BEGIN_NAMESPACE_OPENDAQ
  * [includeHeader("<coretypes/event_wrapper.h>")]
  */
 
-enum class ComponentErrorState : EnumType
-{
-    Ok = 0,
-    Warning,
-    Error
-};
-
 /*!
  * @brief Acts as a base interface for components, such as device, function block, channel and signal.
  *
--- a/core/opendaq/component/include/opendaq/component.h
+++ b/core/opendaq/component/include/opendaq/component.h
@@ -39,6 +39,12 @@ BEGIN_NAMESPACE_OPENDAQ
  * [includeHeader("<coretypes/event_wrapper.h>")]
  */
 
+enum class ComponentErrorState : EnumType
+{
+    Ok = 0,
+    Warning,
+    Error
+};
 
 /*!
  * @brief Acts as a base interface for components, such as device, function block, channel and signal.
--- a/core/opendaq/component/include/opendaq/component_impl.h
+++ b/core/opendaq/component/include/opendaq/component_impl.h
@@ -58,6 +58,13 @@ BEGIN_NAMESPACE_OPENDAQ
 
 #define COMPONENT_AVAILABLE_ATTRIBUTES {"Name", "Description", "Visible", "Active"}
 
+enum class ComponentErrorState : EnumType
+{
+    Ok = 0,
+    Warning,
+    Error
+};
+
 template <class Intf = IComponent, class ... Intfs>
 class ComponentImpl : public GenericPropertyObjectImpl<Intf, IRemovable, IComponentPrivate, IDeserializeComponent, Intfs ...>
 {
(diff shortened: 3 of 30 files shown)


core/coreobjects/include/coreobjects/core_event_args_impl.h
```diff
@@ -235,7 +235,7 @@ inline bool CoreEventArgsImpl::validateParameters() const
         case CoreEventId::DeviceLockStateChanged:
             return parameters.hasKey("IsLocked");
         case CoreEventId::StatusChanged:
-            return parameters.getCount() == 1;
+            return parameters.getCount() == 2;
```
JakaMohorkoDS, line 238: I'd delete this check. I'm not exactly sure if this is backwards compatible, as older devices will send out events with parameterCount == 1

core/opendaq/component/include/opendaq/component.h
```diff
@@ -39,6 +39,12 @@ BEGIN_NAMESPACE_OPENDAQ
  * [includeHeader("<coretypes/event_wrapper.h>")]
  */
 
+enum class ComponentErrorState : EnumType
+{
+    Ok = 0,
+    Warning,
+    Error
+};
```
JakaMohorkoDS, line 47: I would not have this on a public header. This should be in the component implementation file.

core/opendaq/component/include/opendaq/component_impl.h
```diff
@@ -1114,6 +1120,43 @@ bool ComponentImpl<Intf, Intfs...>::validateComponentId(const std::string& id)
     return id.find(' ') == std::string::npos;
 }
 
+template <class Intf, class... Intfs>
+void ComponentImpl<Intf, Intfs...>::initComponentErrorStateStatus() const
+{
+    // Component error state status is added ("Ok" when a component is created)
+    const auto statusContainerPrivate = this->statusContainer.template asPtr<IComponentStatusContainerPrivate>();
```
dejancrn, line 1127: Call with borrow parameter =True

core/opendaq/component/include/opendaq/component_impl.h
```diff
@@ -1114,6 +1120,43 @@ bool ComponentImpl<Intf, Intfs...>::validateComponentId(const std::string& id)
...
+        auto dummy = this->statusContainer.getStatus("ComponentStatus");
+    }
+    catch (const NotFoundException&)
+    {
+        throw NotFoundException("ComponentStatus has not been added to statusContainer. initComponentErrorStateStatus needs to be called "
+                                "before setComponentErrorStateStatus.");
+    }
+
+    // Set status if initialized
+    const auto statusContainerPrivate = this->statusContainer.template asPtr<IComponentStatusContainerPrivate>();
```
dejancrn, line 1154: Borrow

core/opendaq/component/tests/test_component.cpp
```diff
@@ -12,6 +12,9 @@
 #include <opendaq/component_deserialize_context_factory.h>
 #include <opendaq/component_status_container_private_ptr.h>
 #include <opendaq/component_exceptions.h>
+#include <opendaq/component_impl.h>
+
+#include "testutils/testutils.h"
```
JakaMohorkoDS, line 17: Should wrap with "<>"

core/opendaq/component/tests/test_component_status_container.cpp
```diff
@@ -5,6 +5,8 @@
 #include <opendaq/context_factory.h>
 #include <opendaq/component_deserialize_context_factory.h>
 
+#include "opendaq/instance_factory.h"
```
JakaMohorkoDS, line 8: Wrap with "<>"

JakaMohorkoDS (commented):
@NikolaiShipilov Can you check if we need to add anything to the streaming connection statuses here?

dejancrn (approved):

core/opendaq/component/include/opendaq/component_impl.h
aljazfrancic: @dejancrn Fixed! However I'm just curious why this is needed or preferred. Does it affect performance? My understanding is that that the variable will go out of scope at the end of `initComponentErrorStateStatus`, thereby decreasing the reference count to the original value.

dejancrn: By borrowing you avoid unnecessary reference count increment when smart ptr is created and decrement when destroyed (out of scope). Not super important in configuration part of openDAQ (more important on signal path). However, it makes sense to get a grasp why borrowing is better in cases when you do not keep a reference to object.

Reference count is an atomic variable and as such requires memory barrier and memory bus locks when incremented and decremented.

aljazfrancic: @dejancrn Thank you for the nice explanation! I'll keep this in mind!

core/opendaq/modulemanager/src/context_impl.cpp
```diff
@@ -304,6 +304,10 @@ void ContextImpl::registerOpenDaqTypes()
                                     .build();
 
     typeManager->addType(ptpSyncInterface);
+
```
NikolaiShipilov, line 307: Could you please also move the registration of `"ConnectionStatusType"` to this method instead of registering it directly in the constructor? I initially overlooked the fact that there’s a `registerOpenDaqTypes` method, which is a more appropriate place for such registration.

NikolaiShipilov (approved):

NikolaiShipilov:
> @NikolaiShipilov Can you check if we need to add anything to the streaming connection statuses here?

I think the additional messages are unnecessary for connection statuses since the status values have a fixed meaning and don’t require further clarification. Therefore, connection statuses can remain as they are for now.

core/opendaq/modulemanager/src/context_impl.cpp
aljazfrancic: I moved it and also removed additional `hasType` check, since this is only called once per Context.

JakaMohorkoDS (approved):

PR #635, opened 2024-12-16
Remove root device on instance destruction

JakaMohorkoDS:
# Type of change:

# Description:

- `rootDevice.remove()` is now called on instance destruction.
- openDAQ components can hook onto the `onRemoved()` that will be called when the openDAQ program terminates.

### DIFF
--- a/core/opendaq/opendaq/src/instance_impl.cpp
+++ b/core/opendaq/opendaq/src/instance_impl.cpp
@@ -64,6 +64,7 @@ InstanceImpl::InstanceImpl(IInstanceBuilder* instanceBuilder)
 InstanceImpl::~InstanceImpl()
 {
     stopAndRemoveServers();
+    rootDevice.remove();
     rootDevice.release();
 }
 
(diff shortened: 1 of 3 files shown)


JakaMohorkoDS:
Should all components call `removed()` in their destructor if not yet removed?

core/opendaq/opendaq/src/instance_impl.cpp
```diff
@@ -132,8 +133,13 @@ static ContextPtr ContextFromInstanceBuilder(IInstanceBuilder* instanceBuilder)
     return Context(scheduler, logger, typeManager, moduleManager, authenticationProvider, options, discoveryServers);
 }
 
-void InstanceImpl::stopAndRemoveServers()
+void InstanceImpl::stopAndRemoveServers() const
 {
+    Bool removed = false;
```
dejancrn, line 138: you can use `rootDevice.isRemoved()`

dejancrn (approved):

PR #637, opened 2024-12-17
Make attributes editable in Python GUI Demo Application

aljazfrancic:
# Type of change:

# Checklist:

# Description:

This fixes the issue of attributes not being editable in Python GUI Demo Application
* This works, because there is an unexpected `.` at the beginning of `sel`
* Because the code that handles the logic for this action was already there I'm assuming this either:
  * Never worked as intended
  * Stopped working due to to updates in either Python or tkinter versions (most likely the latter)
    * Tested using:
      * `Python 3.12.8`
      * `tkinter 8.6`
      * `opendaq-3.11.0+301c940-cp312-cp312-win_amd64.whl`
      * Not sure what else could be relevant
* My other worry is that `utils.treeview_get_first_selection` function has the possibility of returning `None`, and, in turn, this generating an error, but I'm guessing `None` was intended for other cases where this function is used

### DIFF
--- a/examples/python/gui_demo/components/attributes_dialog.py
+++ b/examples/python/gui_demo/components/attributes_dialog.py
@@ -233,7 +233,7 @@ class AttributesDialog(Dialog):
             locked = utils.yes_no[locked_flag]
             iid = f'{parent}.{key}'
             if type(value) is bool:
-                value = yes_no[value]
+                value = utils.yes_no[value]
             elif type(value) is dict:
                 self.tree.insert(
                     parent, tk.END, iid=iid, text=key, values=('', locked))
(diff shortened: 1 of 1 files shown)


aljazfrancic:
I'm assuming the issues raised in the description will have to be checked before merging.

aljazfrancic:
I have now also tested with:
* `Python 3.8.10`
* `tkinter 8.6` (this one actually comes bundled with Python, so there's no way to change it really)
* `opendaq-3.11.0+301c940-cp38-cp38-win_amd64.whl`

So I guess fix the fix is valid.

Yarikk26 (approved):

JakaMohorkoDS (approved):

PR #645, opened 2024-12-19
Properly set component status on reconnect

dejancrn:
Component status was not a part of remote-update on configuration client objects.

Slightly off-topic, I don't like the way we have deserialization/update implemented on remote objects. Every change applied on openDAQ components must be applied to configuration client remote objects. I prefer the remote update to be implemented on the components themselves.

### DIFF
--- a/shared/libraries/config_protocol/include/config_protocol/config_client_component_impl.h
+++ b/shared/libraries/config_protocol/include/config_protocol/config_client_component_impl.h
@@ -54,6 +54,7 @@ protected:
                                                     const FunctionPtr& factoryCallback);
 
     void handleRemoteCoreObjectInternal(const ComponentPtr& sender, const CoreEventArgsPtr& args) override;
+    void remoteUpdateStatuses(const SerializedObjectPtr& serializedStatuses);
     void onRemoteUpdate(const SerializedObjectPtr& serialized) override;
 
 private:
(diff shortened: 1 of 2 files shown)


shared/libraries/config_protocol/include/config_protocol/config_client_component_impl.h
```diff
@@ -193,6 +194,40 @@ void ConfigClientComponentBaseImpl<Impl>::handleRemoteCoreObjectInternal(const C
...
+            StringPtr msg;
+            if (messagesDict.hasKey(name))
+                msg = messagesDict.get(name);
+            else
+                msg = String("");
+
+            if (statuses.hasKey(name))
+                statusContainerPrivate.setStatusWithMessage(name, value, msg);
+            else
+                statusContainerPrivate.addStatusWithMessage(name, value, msg);
```
JakaMohorkoDS, line 226: I suppose this the general state of our statuses, but do we need to ever remove a status on update?

JakaMohorkoDS (approved):

shared/libraries/config_protocol/include/config_protocol/config_client_component_impl.h
dejancrn: I don't think there is a way to remove the status. There's no method available for this.

PR #649, opened 2024-12-24
Upload Python Wheels to S3 even if not `refs/heads/main`

aljazfrancic:
Enables uploading Python Wheels to S3 from branches other than `refs/heads/main` to be more in line with other files (nuget package, docs, etc.) being uploaded (e.g. via `workflow_dispatch`) for `deploy.yml` workflow.

Noticed the files were missing when I tried deploying files to AWS S3 for "manual" update of  https://docs.opendaq.com to `3.10.2`.


Yarikk26 (approved):
It's hard to review due to lots of changes, but it looks ok for me :)

PR #650, opened 2024-12-27
Component status fixes

aljazfrancic:
# Type of change:

# Checklist:

# Description:

Component status fixes:
* Streamline naming for Component Status building blocks:
  * `ComponentErrorStateStatus` to `ComponentStatus` (all variable names)
  * `ComponentErrorState` to `ComponentStatus` (enum)
* Equip all existing Function Block implementations  with additional `setComponentStatus` calls, setting the Component Status back to `ComponentStatus::Ok`
* Avoid redundant code in Function Block implementations and improve logging:
  * `LOG_` (`W`/`E`/`I` for Warning/Error/Ok) on each `setComponentStatusWithMessage`
  * Delete all "manual" `LOG_` calls in Function Block implementations adjacent/related to `setComponentStatus`/`setComponentStatusWithMessage`
* Improve Component Status messages (and thereby log messages) in reference Function Blocks with additional info wherever possible
* Add comments to `initComponentStatus`, `setComponentStatus`, and `setComponentStatusWithMessage` in `component_impl.h` that hopefully show up in users' IDEs (works in Visual Studio Community 2022)
* Supplement unit tests

Unrelated changes:
* Change all changelog dates to ISO 8601
* Make minor cosmetic/grammar changes to `README.md`
* Fix typos, minor changes

### DIFF
--- a/modules/ref_fb_module/src/fft_fb_impl.cpp
+++ b/modules/ref_fb_module/src/fft_fb_impl.cpp
@@ -179,7 +179,7 @@ void FFTFbImpl::configure()
     }
     catch (const std::exception& e)
     {
-        setComponentStatusWithMessage(ComponentStatus::Warning, fmt::format("FFT: Failed to set descriptor for signal: {}", e.what()));
+        setComponentStatusWithMessage(ComponentStatus::Error, fmt::format("FFT: Failed to set descriptor for signal: {}", e.what()));
         outputSignal.setDescriptor(nullptr);
     }
 }
--- a/modules/ref_fb_module/src/fft_fb_impl.cpp
+++ b/modules/ref_fb_module/src/fft_fb_impl.cpp
@@ -190,6 +190,7 @@ void FFTFbImpl::configure()
         LOG_W("FFT: Failed to set descriptor for signal: {}", e.what())
         outputSignal.setDescriptor(nullptr);
     }
+    setComponentStatusWithMessage(ComponentStatus::Ok, "Configuration successful");
 }
 
 void FFTFbImpl::processEventPacket(const EventPacketPtr& packet)
--- a/modules/ref_fb_module/src/fft_fb_impl.cpp
+++ b/modules/ref_fb_module/src/fft_fb_impl.cpp
@@ -186,8 +186,7 @@ void FFTFbImpl::configure()
     }
     catch (const std::exception& e)
     {
-        setComponentStatusWithMessage(ComponentStatus::Warning, "Failed to set descriptor for signal");
-        LOG_W("FFT: Failed to set descriptor for signal: {}", e.what())
+        setComponentStatusWithMessage(ComponentStatus::Warning, fmt::format("FFT: Failed to set descriptor for signal: {}", e.what()));
         outputSignal.setDescriptor(nullptr);
     }
     setComponentStatus(ComponentStatus::Ok);
--- a/modules/ref_fb_module/src/fft_fb_impl.cpp
+++ b/modules/ref_fb_module/src/fft_fb_impl.cpp
@@ -22,7 +22,7 @@ namespace FFT
 FFTFbImpl::FFTFbImpl(const ContextPtr& ctx, const ComponentPtr& parent, const StringPtr& localId)
     : FunctionBlock(CreateType(), ctx, parent, localId)
 {
-    initComponentErrorStateStatus();
+    initComponentStatus();
     initProperties();
     createSignals();
     createInputPorts();
@@ -100,7 +100,7 @@ void FFTFbImpl::configure()
     {
         if (inputDataDescriptor.getSampleType() == SampleType::Struct || inputDataDescriptor.getDimensions().getCount() > 0)
         {
-            setComponentErrorStateStatusWithMessage(ComponentErrorState::Error, "Incompatible input value data descriptor");
+            setComponentStatusWithMessage(ComponentStatus::Error, "Incompatible input value data descriptor");
             throw std::runtime_error("Incompatible input value data descriptor");
         }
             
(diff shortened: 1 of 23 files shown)


modules/ref_fb_module/src/renderer_fb_impl.cpp
```diff
@@ -727,6 +730,8 @@ void RendererFbImpl::renderLoop()
     auto waitTime = defaultWaitTime;
     while (!stopRender && window.isOpen())
     {
+        setComponentStatus(ComponentStatus::Ok);
```
aljazfrancic, line 733: Please pay attention to this method call during the review, as it may be problematic. The reason being that we do this inside a loop (the `renderLoop`), which might result in Component status being changed many times, if something is wrong. E.g. this will reset it to `Ok` every time, then inside the loop the status might change again during execution (again multiple times for each loop iteration). Which in turn might cause the status to "blink" in the user interface if implemented. One solution is to put it outside of the loop, but then it won't reset once set. Also whatever Warning/Error might happen inside the loop will repeat multiple times regardless.

modules/ref_fb_module/src/scaling_fb_impl.cpp
```diff
@@ -110,23 +110,21 @@ void ScalingFbImpl::processSignalDescriptorChanged(const DataDescriptorPtr& inpu
 
 void ScalingFbImpl::configure()
 {
+    setComponentStatus(ComponentStatus::Ok);
```
dejancrn, line 113: Why do you set to OK at the start of method. Simply call this as last step in try..catch:

try
{
  ..
  ..
  ..
  SetStatusToOK
}
catch
{
  SetStatusToWarn/Err
}

modules/ref_fb_module/src/renderer_fb_impl.cpp
dejancrn: Maybe you can introduce internal member that remembers the last state and only change the state when current state is different than previous.

dejancrn (changes requested):
I don't like that status is changed first to "OK" and then again to OK or error state on each configure on function blocks.

Also render should only change the state when something changes.

dejancrn:
Another potential solution would be NOT to trigger status changed event when status is set and the value is the same as previous.
@JakaMohorkoDS What do you think?

modules/ref_fb_module/src/fft_fb_impl.cpp
```diff
@@ -186,10 +186,10 @@ void FFTFbImpl::configure()
     }
     catch (const std::exception& e)
     {
-        setComponentErrorStateStatusWithMessage(ComponentErrorState::Warning, "Failed to set descriptor for signal");
-        LOG_W("FFT: Failed to set descriptor for signal: {}", e.what())
+        setComponentStatusWithMessage(ComponentStatus::Warning, fmt::format("FFT: Failed to set descriptor for signal: {}", e.what()));
         outputSignal.setDescriptor(nullptr);
     }
+    setComponentStatus(ComponentStatus::Ok);
```
JakaMohorkoDS, line 192: Considering we always catch exceptions, we will always override the component status with "Ok" at the end here.

modules/ref_fb_module/src/fft_fb_impl.cpp
```diff
@@ -186,10 +186,10 @@ void FFTFbImpl::configure()
     }
     catch (const std::exception& e)
     {
-        setComponentErrorStateStatusWithMessage(ComponentErrorState::Warning, "Failed to set descriptor for signal");
-        LOG_W("FFT: Failed to set descriptor for signal: {}", e.what())
+        setComponentStatusWithMessage(ComponentStatus::Warning, fmt::format("FFT: Failed to set descriptor for signal: {}", e.what()));
```
JakaMohorkoDS, line 189: This overrides any component statuses set above.

modules/ref_fb_module/src/renderer_fb_impl.cpp
```diff
@@ -174,9 +174,13 @@ void RendererFbImpl::readProperties()
...
-        setComponentErrorStateStatusWithMessage(ComponentErrorState::Error,
-                                                "Property custom2dMaxRange have to be more then custom2dMinRange");
-        LOG_E("Property custom2dMaxRange have to be more then custom2dMinRange");
+        setComponentStatusWithMessage(ComponentStatus::Error,
+                                                "Property custom2dMaxRange has to be more than custom2dMinRange");
+        LOG_E("Property custom2dMaxRange has to be more than custom2dMinRange");
+    }
+    else
+    {
+        setComponentStatus(ComponentStatus::Ok);
```
JakaMohorkoDS, line 183: This might not be the best pattern for log handling, as we set the status to OK no matter the previous state. Likely the logs should be cleared in a separate method that checks all preconditions. 

This pattern might be okay in function blocks where the status is changed only within 1 function.

JakaMohorkoDS (changes requested):

modules/ref_fb_module/src/power_fb_impl.cpp
```diff
@@ -485,10 +468,10 @@ void PowerFbImpl::configure(bool resync)
     }
     catch (const std::exception& e)
     {
-        setComponentErrorStateStatusWithMessage(ComponentErrorState::Warning, "Failed to set descriptor for power signal");
-        LOG_W("Failed to set descriptor for power signal: {}", e.what())
+        setComponentStatusWithMessage(ComponentStatus::Warning, fmt::format("Failed to set descriptor for power signal: {}", e.what()));
```
JakaMohorkoDS, line 471: Should probably be Error.

JakaMohorkoDS (approved):

PR #655, opened 2024-12-31
Device module template

JakaMohorko:
# Type of change:

# Description:


NikolaiShipilov (approved):
lgtm, there are a few considerations regarding future-proofing and upcoming improvements:
* Would there be in-code documentation within the template framework (e.g., Doxygen) detailing template and hooks classes, or  only guidance available within the documentation site?
* If developers implementing a device need to extend certain implementations, e.g. adding a custom IP configuration handling mechanism within the device implementation, could they still use the device template and extend the hooks with appropriate overrides? Or should they follow an implementation model that does not rely on templates instead, as seen in the existing reference device module example ?
* The newly introduced CMake options will likely need to be documented in CMake-Options.md .

PR #657, opened 2025-01-03
Show time domain signal last value as time

dejancrn:
This PR will properly show time domain signal last value

{6BFE310B-96DC-4D8A-B5F5-4B618586601E}

However, this is only a temporary solution. As a final solution, I propose a new core type, `IAbsTime`, and return this one as an object when calling `ISignal.getLastValue`. Cpp bindings should convert this type to `std::chrono::system_clock`˙and Python bindings should convert it to Python `datetime` object.

### DIFF
--- a/examples/python/gui_demo/utils.py
+++ b/examples/python/gui_demo/utils.py
@@ -2,6 +2,7 @@ import os
 import tkinter as tk
 from tkinter import ttk
 from tkinter import messagebox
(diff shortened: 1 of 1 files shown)


JakaMohorkoDS (approved):
Looks good. As for the IAbsTime object, we'd likely need to think on whether it should include information about time zones/types (GPS, TAI).

PR #661, opened 2025-01-10
Show device connection status and refresh on connection status changed event

dejancrn:


examples/python/gui_demo/app_context.py
```diff
@@ -160,3 +163,10 @@ def properties_of_component(self, component: daq.IComponent):
             return daq.IList()
 
         return component.all_properties if self.view_hidden_components else component.visible_properties
+
+    def on_core_event(self, sender: Optional[daq.IComponent], args: daq.IEventArgs):
+        if sender is not None and daq.IDevice.can_cast_from(sender) and args.event_name == "StatusChanged" and self.on_needs_refresh is not None:
+            core_event_args: daq.ICoreEventArgs = daq.ICoreEventArgs.cast_from(args)
+            has_connection_status = any(k == "ConnectionStatus" for k in core_event_args.parameters.keys())
```
adolenc, line 170:
```suggestion
            has_connection_status = "ConnectionStatus" in core_event_args.parameters
```
Maybe you need parameters.keys(), I don't know if parameters is a normal python dict.

examples/python/gui_demo/utils.py
```diff
@@ -275,3 +275,13 @@ def prettify_bool(value):
     'read_only': prettify_bool,
     'visible': prettify_bool
 }
+
+def is_device_connected(device: daq.IDevice):
+    status_container = device.status_container
+    try:
+        status_container.statuses
```
adolenc, line 282: Leftover? Otherwise this doesn't feel right...

dejancrn: `is_device_connected` is used here
is_device_connected

examples/python/gui_demo/app_context.py
dejancrn: On a second check, python will user "__iter__" operator if "__contains__" is not supported, so "in" operator should work

examples/python/gui_demo/utils.py
adolenc: No, I mean just that one specific line, `status_container.statuses`. It's weird to see a statement like that in Python. It looks like a property accessor and I would not really expect it to "do" anything on its own.

dejancrn: I see, yeah just a leftover

PR #662, opened 2025-01-15
Use generic Property Treeview in Python GUI Demo Application

aljazfrancic:
# Brief

Introduce `generic_properties_treeview.py` in Python GUI Demo Application, use it in `add_config_dialog.py` and `properties_view.py`

# Description

* Introduce `generic_properties_treeview.py`, a Tkinter Treeview class that can be used to modify properties
* Use `generic_properties_treeview.py` in `add_config_dialog.py`
* Use `generic_properties_treeview.py` in `properties_view.py`
* Minor fixes and changes

# Usage example

```py
from .generic_properties_treeview import PropertiesTreeview

PropertiesTreeview(parent, node, context)
```

Where `parent` represents a Tkinter widget to act as the parent of the current object, `node` is the Property Object to display in Treeview and `context` is Python `AppContext`.

# Required integration changes

## Breaking application changes

None.

## Breaking module changes

None.

# API changes

None.


Yarikk26 (commented):
I ran it on my machine. This is what I found:
- the height of properties view for device is different in comparison with channel (it's personal, didn't find any constraints in GUI specifications)
- type mismatch causes unhandled exception (when entering string into int field)

Yarikk26 (approved):

JakaMohorkoDS (approved):

PR #670, opened 2025-01-21
remove limitation on amount of txt records for mDNS

denise-opendaq:
# Type of change:

# Checklist:

# Description:

### DIFF
--- a/shared/libraries/discovery_server/src/mdnsdiscovery_server.cpp
+++ b/shared/libraries/discovery_server/src/mdnsdiscovery_server.cpp
@@ -33,6 +33,7 @@ MdnsDiscoveredDevice::MdnsDiscoveredDevice(const std::string& serviceName, uint3
 
 void MdnsDiscoveredDevice::populateRecords(std::vector<mdns_record_t>& records) const 
 {
+    this->recordSize = 1024;
     for (const auto & [key, value] : properties)
     {
         mdns_record_t record;
@@ -43,6 +44,8 @@ void MdnsDiscoveredDevice::populateRecords(std::vector<mdns_record_t>& records) 
         record.rclass = 0,
         record.ttl = 0;
         records.push_back(record);
+
+        this->recordSize += key.size() + value.size() + 2;
     }
 }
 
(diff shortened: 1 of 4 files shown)


shared/libraries/discovery_server/src/mdnsdiscovery_server.cpp
```diff
@@ -604,30 +620,38 @@ int MDNSDiscoveryServer::serviceCallback(int sock, const sockaddr* from, size_t
                     records.push_back(createARecord(device));
                 if (serviceAddressIpv6.sin6_family == AF_INET6)
                     records.push_back(createAaaaRecord(device));
+                device.populateRecords(records);
 
+                std::vector<char>sendBuffer(device.recordSize);
                 send_mdns_query_answer(unicast, sock, from, addrlen, sendBuffer, query_id, rtype, name, answer, records);
             }
         } 
         else if (name == device.serviceQualified)
```
NikolaiShipilov, line 629: In general, it seems fine that we now attach TXT records to responses for any DNS requests with a matching name. 
Just to clarify - currently, all other branches of the `if (name == smth)` statement, except for `else if (name == serviceName)`, aren't typically executed during normal operation since the request with `serviceName` is the only one being actually ever formed by the OpenDAQ discovery client. Correct?

shared/libraries/discovery_server/src/mdnsdiscovery_server.cpp
```diff
@@ -33,6 +33,7 @@ MdnsDiscoveredDevice::MdnsDiscoveredDevice(const std::string& serviceName, uint3
 
 void MdnsDiscoveredDevice::populateRecords(std::vector<mdns_record_t>& records) const 
 {
+    this->recordSize = 1024;
```
NikolaiShipilov, line 36: Would it make sense to set `1024` as the default value (e.g., in the constructor) instead of initializing it here?

shared/libraries/discovery_server/src/mdnsdiscovery_server.cpp
```diff
@@ -73,6 +78,7 @@ MDNSDiscoveryServer::MDNSDiscoveryServer(void)
     if (WSAStartup(versionWanted, &wsaData))
     {
         printf("Failed to initialize WinSock\n");
+        return;
```
NikolaiShipilov, line 81: I would consider throwing an exception here (and the same for call failure in the mDNS client), as this failure makes the server non-functional anyway. A `printf` could easily be missed in the logs (or might not even appear), making debugging more difficult.

shared/libraries/discovery_server/src/mdnsdiscovery_server.cpp
denise-opendaq: yes, but I checked mdns server with external mdns client as well, and without txt records, the client can not see inforamation about txt records, if client requesting for data while server is already working

PR #671, opened 2025-01-21
Apply `run_rtgen.sh`

aljazfrancic:
# Brief

Commit changes after running `run_rtgen.sh`, thereby updating Python bindings
(all changes were auto-generated and manually fixed afterwards)


aljazfrancic:
Due to the need for manual fixes I think we should update rtgen in such a way that there is no need for manual fixes? @Yarikk26 What do you think?

Yarikk26:
> Due to the need for manual fixes I think we should update rtgen in such a way that there is no need for manual fixes? @Yarikk26 What do you think?

If I'm correct, we already have some tasks in backlog for that

Yarikk26 (approved):

PR #673, opened 2025-01-21
Initialize `futureComponentStatus` and `futureComponentMessage` in `renderer_fb_impl.cpp`, thereby preventing unwarranted crashes

aljazfrancic:
I think this should fix [TBBAS-2070] as well (must be tested to confirm)


Yarikk26 (approved):
It works

Issue #677, opened 2025-01-23
[BUG] Can't change List Property remotely with Python GUI


sedovmo:
### Current behavior

I have an Opedaq server and a Function block with a List property instantiated on the server. I can't change this List property remotely using the Python GUI. I get "Can't add item: Unsupported core type: CoreType.ctUndefined" error.

![Image](https://github.com/user-attachments/assets/c0efd82c-41e7-4197-bb60-4c6fb9bd6867)

But I can change this property using C++, so it seems like the Python only issue

### Expected behavior

I can change this property if I add the Function block locally.

![Image](https://github.com/user-attachments/assets/0d5ff598-f549-4546-b7d9-fa864d63e912)

### Code or steps

Run a device server. Open the Python GUI.
In Python GUI
1. Add device with Connection string "daq.nd://127.0.0.1"
2. Add Function Block with List property in it
3. Double click this List property, you'll see a new "Edit" window
4. Press "Add" button, you'll see a new "Add item" window
5. Type some text and press OK button
You'll see the error

### Environment

* openDAQ version tag or commit SHA:
Python GUI under Ubuntu -  Opendaq 3.11.0.d3f7343
Python GUI under Windows 11 - Opendaq 3.11.0.ae8716f

* OS: server - Ubuntu 22.04.5, python GUI run under Ubuntu 22.04.5  and Windows 11.
* programming language: Python

aljazfrancic:
Trying to reproduce this, I made a server like this:

```cpp
#include <iostream>
#include <opendaq/opendaq.h>

using namespace daq;

int main(int /*argc*/, const char* /*argv*/[])
{
    const InstancePtr instance = Instance("");
    instance.setRootDevice("daqref://device0");
    auto dummy = instance.addStandardServers();
    instance.addProperty(ListProperty("listProp", List<IBaseObject>("foo", "bar")));
    std::cin.get();
    return 0;
}
```
On the latest main.

Then I add device with Connection string `daq.nd://127.0.0.1` in Python GUI app.

This is somewhat different because it's on a Device instead of a Function Block.

This works fine, with either `IBaseObject` or `IString` as the list templated type.

However, if I, for example, do:

```cpp
instance.addProperty(ListProperty("listProp", List<IString>()));
```

(e.g. empty list of whatever type)

this will fail with the error you get as well.

Anyways, here's the fix:
https://github.com/openDAQ/openDAQ/pull/686

aljazfrancic:
[...]

PR #686, opened 2025-01-30
Allow adding items to an empty `List<IBaseObject>` Property in Python GUI demo app

aljazfrancic:
Allow adding items to an empty `List<IBaseObject>` Property in Python GUI demo app thereby solving https://github.com/openDAQ/openDAQ/issues/677

### DIFF
--- a/examples/python/gui_demo/components/edit_container_property.py
+++ b/examples/python/gui_demo/components/edit_container_property.py
@@ -141,8 +141,7 @@ class EditContainerPropertyDialog(Dialog):
             if value is None:
                 return
             try:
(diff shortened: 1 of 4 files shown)


examples/python/gui_demo/components/edit_container_property.py
```diff
@@ -127,8 +127,7 @@ def add_item(self):
             if value is None:
                 return
             try:
-                self.data.append(utils.value_to_coretype(
-                    value, self.property.item_type))
+                self.data.append(value)
```
JakaMohorkoDS, line 130: I'd think we want to try and convert to the target type. We could have the 2nd option as a fallback.

PR #689, opened 2025-01-31
Mdns discovering multiple devices with the same ip

denise-opendaq:
# Brief
- Fixing the discovering multiple devices on the same ip
- fix serviceQualified in mdns server

### DIFF
--- a/shared/libraries/discovery/include/daq_discovery/mdnsdiscovery_client.h
+++ b/shared/libraries/discovery/include/daq_discovery/mdnsdiscovery_client.h
@@ -108,6 +108,7 @@ protected:
         SRVRecord SRV;
         std::string A;
         std::string AAAA;
+        std::string linkLocalInterface;
         std::vector<std::pair<std::string, std::string>> TXT;
     } DeviceData;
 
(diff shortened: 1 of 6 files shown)


shared/libraries/discovery/include/daq_discovery/mdnsdiscovery_client.h
```diff
@@ -255,7 +256,6 @@ inline MDNSDiscoveryClient::~MDNSDiscoveryClient()
 
 inline std::vector<MdnsDiscoveredDevice> MDNSDiscoveryClient::getAvailableDevices()
 {
-    devicesMap.clear();
```
NikolaiShipilov, line 258: So, by moving the clearing process after issuing the next mDNS request, the timed-out responses (those received after the `discoveryDuration` has expired but still readable from the input socket) will now be taken into account during the device "rescan", correct?

shared/libraries/discovery/include/daq_discovery/mdnsdiscovery_client.h
```diff
@@ -684,28 +683,31 @@ inline int MDNSDiscoveryClient::discoveryQueryCallback(int sock,
...
+    char nameBuffer[512];
 
     std::lock_guard lg(devicesMapLock);
-    
-    std::string deviceAddr = ipAddressToString(from, addrlen);
 
-    auto it = devicesMap.insert({deviceAddr, DeviceData{}});
+    std::string deviceName;
+    {
+        mdns_string_t namestr = mdns_string_extract(buffer, size, &rname_offset, nameBuffer, sizeof(nameBuffer));
```
NikolaiShipilov, line 696: You can use the `DiscoveryUtils::extractRecordName` common method to read the record name.

NikolaiShipilov (approved):
Looks good! only added a few clarification questions.
Also, since the way how client enumerates discovered devices has changed, which may introduce potential limitations related to name conflicts, it would be worth mentioning this in the changelog.

shared/libraries/discovery/include/daq_discovery/mdnsdiscovery_client.h
denise-opendaq: the more devices we read, the better :)

shared/libraries/discovery/include/daq_discovery/mdnsdiscovery_client.h
```diff
@@ -684,28 +683,31 @@ inline int MDNSDiscoveryClient::discoveryQueryCallback(int sock,
     if (opcode)
         return 0;
 
-    char nameBuffer[256];
+    // There is no need to handle PTR records
+    if (rtype == MDNS_RECORDTYPE_PTR)
+        return 0;
+
+    char nameBuffer[512];
```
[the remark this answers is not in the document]
denise-opendaq: yes, but it also used for parsing the records, which are including the name

PR #693, opened 2025-02-07
Install missing CMake utilities

JakaMohorkoDS:
# Brief

Standard example openDAQ modules make use of non-installed openDAQ cmake utilities. This PR adds said cmake utilities to the installed openDAQ package.

# Description

The files DaqUtils.cmake, DaqInternal.cmake, as well as the input templates for version files have been added as targets for openDAQ installation. 

Module examples such as the [Simple FB Module](https://github.com/openDAQ/SimpleFBModule) and [Simple Device Module](https://github.com/openDAQ/SimpleDeviceModule) as of now fail on CMake configure when using an installed version of openDAQ due to the missing cmake utilities.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -746,6 +746,21 @@ install(FILES cmake/${SDK_NAME}Utils.cmake
         COMPONENT ${SDK_NAME}_Development
 )
 
+install(FILES cmake/DaqUtils.cmake
+        DESTINATION share/cmake
+        COMPONENT ${SDK_NAME}_Development
+)
+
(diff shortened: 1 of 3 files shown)


NikolaiShipilov (approved):
Have tested it on Ubuntu 20.04. Successfully configured and built `SimpleFbModule` with the fixed version of package installed, whereas the configuration step fails when using the old package (latest main). So, can confirm that the fix works !

Noticed that `SimpleDeviceModule` ignores the installed 3.11.0 and instead fetches and builds openDAQ from source; though, this is likely fine, as it specifically looks-up for version 3.10.2.

PR #697, opened 2025-02-12
initial BasicRecorder implementation

dvnrrs:
# Brief

Implement the BasicRecorderModule with the BasicRecorderImpl function block.

# Description

This PR adds the "BasicRecorderModule," which exposes a single function block type, "BasicRecorder." This function block allows arbitrary numbers of signals to be connected to its dynamically-managed set of input ports. When recording is activated by setting the `RecordingActive` property or calling `startRecording()` (which could belong to the not-yet-existent `IRecorder` interface), CSV files are created for each of the attached input signals.

# Limitations / Todo

Skeletal unit tests are added which verifies the module can be loaded, but there are not yet any tests to verify the actual functionality of the recorder.

Other known/accepted limitations are:

- Only scalar integer and floating-point types are supported (no complex, raw, multidimensional etc. signals)
- No time domain alignment is performed (requiring each signal to be recorded in its own CSV file)
- No performance optimizations of any kind are made

In other words, this is only realistically suitable as an example or a template for how more advanced recorders can be implemented.

# Usage example

```cpp
auto recorder = instance.addFunctionBlock("BasicRecorder");
recorder.getInputPorts(daq::search::LocalId("Value1"))[0].connect(/* any scalar signal */);
recorder.getInputPorts(daq::search::LocalId("Value2"))[0].connect(/* any scalar signal */);
recorder.setPropertyValue("Path", "/tmp");
recorder.setPropertyValue("RecordingActive", true);
...
recorder.setPropertyValue("RecordingActive", false);
```

# API changes

None.

None.

# Required module changes

None.


core/opendaq/functionblock/include/opendaq/recorder.h
```diff
@@ -0,0 +1,58 @@
...
+ * @{
+ */
+
+/*!
+ * @brief Recorders represent objects which record input data to a persistent storage medium such
+ * as a file, database, or cloud storage bucket. Recorders implement the `IRecorder` interface and
+ * may expose additional properties for configuring the storage (such as file type, storage
+ * location, etc.). tags.
+ */
+DECLARE_OPENDAQ_INTERFACE(IRecorder, IBaseObject)
```
JakaMohorkoDS, line 38: There should probably be another method `getIsRecording` here.

modules/basic_recorder_module/src/basic_recorder_impl.cpp
```diff
@@ -0,0 +1,157 @@
...
+void BasicRecorderImpl::onPacketReceived(const InputPortPtr& port)
+{
+    if (auto signal = findSignal(port))
+        signal->onPacketReceived(port);
+}
+
+void BasicRecorderImpl::addProperties()
+{
+    objPtr.addProperty(StringProperty(Props::PATH, ""));
+    objPtr.addProperty(BoolProperty(Props::RECORDING_ACTIVE, false));
```
JakaMohorkoDS, line 83: Do we still want this property, given that we have start/stop on the interface?

modules/basic_recorder_module/include/basic_recorder_module/basic_recorder_impl.h
```diff
@@ -0,0 +1,152 @@
...
+ * A basic recorder function block which records data from its input signals into a CSV file. A
+ * separate CSV file is created for each recorded signal, in a directory specified by the user via
+ * a property. Recording can be started and stopped by calling member functions or by setting a
+ * property. Signals can be dynamically connected and disconnected. Initially, the function block
+ * has a single input port named "Value1"; additional input ports "Value2" etc. are created so
+ * that at least one unconnected input port is always available. Signals can be connected or
+ * disconnected while the recording is active. Doing so does not disturb ongoing recording of
+ * other signals.
+ */
+class BasicRecorderImpl final : public FunctionBlockImpl<IFunctionBlock, IRecorder>
```
JakaMohorkoDS, line 40: An override of `onActiveChanged` might be good here. We should probably call `stopRecording` when active goes from `true` to `false`. I would not automatically start when going the other direction though.

Or is that behaviour not entirely correct?

modules/basic_recorder_module/src/basic_recorder_signal.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+/**
+ * Generates a CSV filename (without path) for a signal to be recorded based on the signal's
+ * global ID. The ID is sanitized by converting characters not allowed in filenames to
+ * underscores. Only alphanumeric characters, hyphens, periods and underscores are permitted. If
+ * a file with the chosen name already exists, a Windows-style sequence number of the form " (1)"
+ * (using the lowest number that results in an unused filename) is appended.
+ *
+ * @todo There is a race condition finding an unused filename. It is possible for two recorders
+ *     to simultaneously select the same filename and then overwrite each other.
```
JakaMohorkoDS, line 26: Would it make sense to only allow a single instance of the recorder in the module?

modules/basic_recorder_module/src/basic_recorder_signal.cpp
```diff
@@ -0,0 +1,253 @@
...
+        {
+            case PacketType::Data:
+                onDataPacketReceived(packet);
+                break;
+
+            default:
+                break;
+        }
+    }
+}
```
JakaMohorkoDS, line 164: If we are not handling event packets, we should probably stop recording on event packets, as the descriptor of the signal will have likely changed.

modules/basic_recorder_module/src/basic_recorder_signal.cpp
```diff
@@ -0,0 +1,253 @@
...
+ * @param stream The output stream to write to.
+ * @param signal The signal whose unit name to write.
+ *
+ * @return @p stream.
+ */
+static std::ostream& appendUnitInfo(std::ostream& stream, const SignalPtr& signal)
+{
+    bool parentheses = false;
+
+    if (auto descriptor = signal.getDescriptor(); descriptor.assigned())
```
JakaMohorkoDS, line 69: In general, in function blocks we should never call `signal.getDescriptor()`. The descriptors obtained from the input signal can be out-of-date compared to those that come through the connections in the form of event packets. In general, the function blocks should always parse the descriptor information from their received packets.

The first packet to arrive in an input port will always be a `SignalDescriptorChagned` packet, and `onPacketReceived` is triggered synchronously within the `IInputPort::connect` method.

modules/basic_recorder_module/src/basic_recorder_signal.cpp
```diff
@@ -0,0 +1,253 @@
...
+        case SampleType::Int64:     writeSamples<std::int64_t>(packet, writer); return;
+        case SampleType::UInt8:     writeSamples<std::uint8_t>(packet, writer); return;
+        case SampleType::UInt16:    writeSamples<std::uint16_t>(packet, writer); return;
+        case SampleType::UInt32:    writeSamples<std::uint32_t>(packet, writer); return;
+        case SampleType::UInt64:    writeSamples<std::uint64_t>(packet, writer); return;
+        case SampleType::Float32:   writeSamples<float>(packet, writer); return;
+        case SampleType::Float64:   writeSamples<double>(packet, writer); return;
+        default: break;
+    }
+}
```
JakaMohorkoDS, line 251: This could be significantly simplified by using a `StreamReader` with undefined data types. Of course, the simplest option would be to use one that automatically converts to double.

You can then use `descriptor.getSampleSize()` to determine the sample size in bytes.

modules/basic_recorder_module/src/basic_recorder_impl.cpp
dvnrrs: No, we don't. It was left in temporarily because the OneHBK software team has a version of this module which only supports the property (before `IRecorder` was implemented). It should be removed for the merge.

modules/basic_recorder_module/src/basic_recorder_signal.cpp
dvnrrs: No, there are use cases where multiple recorders (of the same or different types) can coexist in a system, so it shouldn't be a singleton.

modules/basic_recorder_module/include/basic_recorder_module/basic_recorder_impl.h
dvnrrs: I think your proposed behavior makes sense to me, and will implement it that way.

JakaMohorkoDS (approved):
Looks good for the initial implementation, we should at some point, however, rework this module to use readers.

The only thing missing at the moment is to add this PR to the list of features in the repository changelog.

JakaMohorkoDS:
@dejancrn Had some questions/suggestions for the PR and will do a code review as well.

dejancrn (commented):
I have already reviewed it

modules/basic_csv_recorder_module/include/basic_csv_recorder_module/basic_csv_recorder_impl.h
```diff
@@ -0,0 +1,162 @@
...
+
+        void addProperties();
+        void addInputPort();
+        void reconfigure();
+
+        bool recordingActive = false;
+
+        std::shared_ptr<BasicCsvRecorderSignal> findSignal(IInputPort *port);
+
+        std::map<IInputPort *, std::shared_ptr<BasicCsvRecorderSignal>> signals;
```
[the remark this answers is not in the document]
dvnrrs: This is a non-owning (weak) reference and I think shouldn't require a reference count. This `std::map` is used as a lookup table. The stored `IInputPort *` key is never itself dereferenced, it's only used to find the associated `BasicCsvRecorderSignal` - and if it's found, it means the caller is holding a corresponding `InputPortPtr`.

I also considered the case of an input port being destroyed, then replaced by a new one at the same memory location (so the pointer is the same). However this case should result in separate sequential calls to `onDisconnected()` and `onConnected()`. This will cause the "stale" pointer to be removed - without being dereferenced - and then a new entry added to the lookup table.

So I believe the behavior is correct in all cases, but let me know if I missed something.

modules/basic_csv_recorder_module/src/basic_csv_recorder_impl.cpp
```diff
@@ -0,0 +1,180 @@
...
+    if (recordingActive)
+    {
+        // We will update the 'signals' map by emplacing new BasicCsvRecorderSignal objects for
+        // newly-connected input ports, and destroying BasicCsvRecorderSignal objects for ports
+        // that are gone or no longer connected. Notably, we will not disturb
+        // BasicCsvRecorderSignal objects for input ports that have not changed.
+
+        // Keep track of the currently-connected input ports we have seen, so we can later destroy
+        // BasicCsvRecorderSignal objects for ports that are gone or no longer connected.
+        std::set<IInputPort *> ports;
```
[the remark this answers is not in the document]
dvnrrs: Refer to above comment: this pointer is used only for lookups and is never dereferenced (a weak non-owning reference).

modules/basic_csv_recorder_module/src/basic_csv_recorder_impl.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+void BasicCsvRecorderImpl::addProperties()
+{
+    objPtr.addProperty(StringProperty(Props::PATH, ""));
+    objPtr.getOnPropertyValueWrite(Props::PATH) += std::bind(&BasicCsvRecorderImpl::reconfigure, this);
+}
+
+void BasicCsvRecorderImpl::addInputPort()
+{
+    auto c = createAndAddInputPort("Value" + std::to_string(++portCount), PacketReadyNotification::SameThread);
```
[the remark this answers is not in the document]
dvnrrs: A worker thread has been created to address this (your "option 3").

modules/basic_csv_recorder_module/src/basic_csv_recorder_impl.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+void BasicCsvRecorderImpl::activeChanged()
+{
+    if (!active)
+        stopRecording();
+}
+
+void BasicCsvRecorderImpl::onPacketReceived(const InputPortPtr& port)
+{
+    auto signal = findSignal(port);
```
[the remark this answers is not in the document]
dvnrrs: The `findThreadForSignal()` function calls `getRecursiveConfigLock()` - is this enough; specifically, does it rule out any concurrent invocations of `onConnected()` and `onDisconnected()`? I assumed so but if that's wrong, and there is no built-in way to do this, I can add a mutex.

modules/basic_csv_recorder_module/src/basic_csv_recorder_impl.cpp
```diff
@@ -0,0 +1,181 @@
...
+    *isRecording = recordingActive;
+
+    return OPENDAQ_SUCCESS;
+}
+
+void BasicCsvRecorderImpl::onConnected(const InputPortPtr& port)
+{
+    addInputPort();
+    reconfigure();
+}
```
JakaMohorkoDS, line 66: These operations likely need to be under a lock, as we have no guarantees that configuration calls are done from the same thread. the `getRecursiveConfigLock` should be used in this case.

JakaMohorkoDS (commented):
Issue in multi threaded configuration calls, and missing test cases. The latter can be added later on, as this is an example function block.

JakaMohorkoDS (changes requested):
On 2nd review, this PR is missing the IRecorder native implementation. With this PR the start/stop recording methods can only be called locally. This needs to be added as a native RPC. Additionally, recorded function blocks require a serialization flag noting that they are recorder for the client to be aware of their type.

nilsRoettgerAtBB:
For usage of the recorded files, it would be very useful to have this additional information (added in the second row in recursive). So, second row has meta data information about the signal.

AI0Time,AI0
_resolution=1/1000000000;EPOCH=1970-01-01T00:00:00+00:00; unit=seconds;delta=1000, unit=voltage_   
1744120476656782,-2.40877
1744120476657782,-2.1289
1744120476658782,-1.84062
1744120476659782,-1.54508
1744120476660782,-1.24345

NikolaiShipilov:
> On 2nd review, this PR is missing the IRecorder native implementation. With this PR the start/stop recording methods can only be called locally. This needs to be added as a native RPC. Additionally, recorded function blocks require a serialization flag noting that they are recorder for the client to be aware of their type.

Added with https://github.com/openDAQ/openDAQ/pull/772

JakaMohorkoDS (approved):

core/opendaq/opendaq/mocks/advanced_components_setup_utils.cpp
```diff
@@ -375,4 +381,35 @@ MockSrvImpl::MockSrvImpl(const ContextPtr& ctx, const DevicePtr& rootDev, const
...
 
+MockRecorderFb1Impl::MockRecorderFb1Impl(const ContextPtr& ctx, const ComponentPtr& parent, const StringPtr& localId)
+    : Super(FunctionBlockType("test_recorder_uid", "test_recorder_name", "test_recorder_description"), ctx, parent, localId, "MockClass")
+{
+}
+
+ErrCode MockRecorderFb1Impl::startRecording()
+{
+    if (this->isRecording)
+        return OPENDAQ_ERR_INVALIDSTATE;
```
denise-opendaq, line 392: hello! we are trying to make the project easier to debug if the error was returned. to do this, was intrdoced macros `DAQ_MAKE_ERROR_INFO(errCode)`, additionally you can leave the readable message `return DAQ_MAKE_ERROR_INFO(OPENDAQ_ERR_INVALIDSTATE, "the function block is already recording")`. so in checking the error code, you will see the error message and path from which place it was returned

denise-opendaq (approved):

NikolaiShipilov:
@dvnrrs , I noticed that the Python bindings are missing for `IRecorder` API

NikolaiShipilov:
> > @dvnrrs , I noticed that the Python bindings are missing for `IRecorder` API
> 
> Who can do this? I don't know anything about Python bindings; what has to be done manually?

Yes, Py binding are modified manually. I'll create additional PR for this.

NikolaiShipilov:
> > > @dvnrrs , I noticed that the Python bindings are missing for `IRecorder` API
> > 
> > 
> > Who can do this? I don't know anything about Python bindings; what has to be done manually?
> 
> Yes, Py binding are modified manually. I'll create additional PR for this.

Here it is https://github.com/openDAQ/openDAQ/pull/782 , I think you can merge it into your branch without waiting for CI https://github.com/openDAQ/openDAQ/actions/runs/14752169446?pr=782 passed, since at least this one https://github.com/openDAQ/openDAQ-Pipelines/actions/runs/14750985862 passed - so, I believe all others should succeed as well.
If some jobs will fail as it happened here https://github.com/openDAQ/openDAQ/actions/runs/14748795525?pr=697 just re-run failed jobs - sometimes it could take several tries to achieve all green :disappointed: as some tests are unstable (work is in progress to address unstable tests)

NikolaiShipilov:
And one last thing - please add reference to this PR into `Features` section of changelog file https://github.com/openDAQ/openDAQ/blob/main/changelog/changelog_3.10-3.20.md

dvnrrs:
Merged via command line (to skip known-good CI checks while also being able to do a rebase instead of merge/squash).

PR #699, opened 2025-02-14
Add additional resolution options to renderer

JakaMohorkoDS:
# Brief

Two new resolution options have been added to the renderer function block: 1920x1080 and 2560x1440.

# Usage example

```cpp
auto instance = Instance();
auto fb = instance.addFunctionBlock("RefFBModuleRenderer");
fb.setPropertyValue("Resolution", 4);
```


aljazfrancic (approved):
LGTM!

PR #704, opened 2025-02-18
Device operation modes

denise-opendaq:
# Brief

Implementing the setting of the operation mode for the device, which notifies all subcomponents of the tree.
The operation mode can be:
- **Idle** – The component can be configured but is not processing signal packets.
- **Safe Operation** – The component can be configured and is processing signal packets, but it does not access physical sinks -(e.g., the writer component reads packets but does not write information to a file).
- **Operation** – The component can be configured, is processing signal packets, and has access to physical sinks.

# Description

Setting the operation mode can only be performed by setting the device's operation mode using:
- **setOperationModeRecursive** – Sets the operation mode for the entire device tree, including subdevices.
- **setOperationMode** – Sets the operation mode for the device without affecting subdevices.
The device may not support all operation modes. The list of available operation modes can be retrieved using the method **getAvailableOperationModes**.

When setting the operation mode, the entire tree is locked, and notifications about the new operation mode are propagated recursively. By default, functions, blocks, and channels set output signals to inactive. However, developers can implement custom logic by overriding the method **onOperationModeChanged**.

When the device's operation mode changes, it triggers a core event with ID **DeviceOperationModeChanged (180)**.

In OPC UA, the server publishes two new properties:
- **OperationModeOptions** – A list of available operation modes for the device.
- **OperationMode** – A string property that indicates the current operation mode. Setting this property's value will call **setOperationMode** on the server. If the value starts with the prefix "**Recursive**" (e.g., "**RecursiveIdle**"), it will call **setOperationModeRecursive** instead.

# Usage example

In C++ use:

imagine there is a device and it has a subdevice.
The device implemented the method **onGetAvailableOperationModes** and **syncOperationMode**
```cpp
std::set<OperationModeType> onGetAvailableOperationModes() override
{
    return {OperationModeType::Idle, OperationModeType::Operation};
}

ErrCode syncOperationMode() override
{
    return updateOperationMode(OperationModeType::Idle);
}

```
The subdevice has the default method of  supports all operation mode/

```cpp
// get available operation modes of device returns {"Idle", "Operation"}
ListPtr<IString> deviceAvailableOpModes = device.getAvailableOperationModes();

// returns {"Idle", "Operation", "SafeOperation"}
ListPtr<IString> subDeviceAvailableOpModes = subdevice.getAvailableOperationModes();

// returns the default operation mode which is "Idle", as we override the syncOperationMode
StringPtr deviceOpMode = device.getOperationMode();

// returns "Operation"
StringPtr subDeviceOpMode = subdevice.getOperationMode();

// setting operation mode for the full tree.
// device will have op mode "Operation" (op mode not changed as the device is not supporting the safe operation mode)
// subdevice will have op mode "SafeOperation"
device.setOperationModeRecursive("SafeOperation");

// device will have op mode "Idle"
// subdevice will have op mode "SafeOperation" (op mode is not changed as setting op mode is not recursive)
device.setOperationMode("Idle");

// device will have op mode "Operation"
// subdevice will have op mode "Operation"
device.setOperationModeRecursive("Operation");

```

# API changes

> [!NOTE]
> breaks binary compatibility of module shared libraries

```diff
+ [interface] ILockGuard: public IBaseObject
+ [function] IPropertyObjectInternal::getLockGuard(ILockGuard** lockGuard)
+ [function] IPropertyObjectInternal::getRecursiveLockGuard(ILockGuard** lockGuard)

+ [function] IComponentPrivate::updateOperationMode(OperationModeType modeType)

+ [function] IDevice::getAvailableOperationModes(IList** availableOpModes)
+ [function] IDevice::setOperationMode(IString* modeType)
+ [function] IDevice::setOperationModeRecursive(IString* modeType)
+ [function] IDevice::getOperationMode(IString** modeType)

+ [factory] CoreEventArgsPtr CoreEventArgsDeviceOperationModeChanged(Int operationMode)
```


core/coreobjects/include/coreobjects/property_object_impl.h
```diff
@@ -102,6 +103,37 @@ class GenericRecursiveConfigLockGuard : public RecursiveConfigLockGuard
     std::lock_guard<TMutex> lock;
 };
 
+template <typename TMutex>
+struct LockGuardImpl : public ImplementationOf<ILockGuard>
```
JakaMohorkoDS, line 107: This is then a recursive lock guard implementation? I'd make two different ones, since the thread ID and depth management is redundant in the case of a non-recursive implementation.

I also feel like these implementations should be moved to a separate utilities file that is not header-only.

core/coreobjects/include/coreobjects/property_object_impl.h
```diff
@@ -2685,6 +2719,32 @@ ErrCode GenericPropertyObjectImpl<PropObjInterface, Interfaces...>::hasUserReadA
...
+    OPENDAQ_PARAM_NOT_NULL(lockGuard);
+    if (externalCallThreadId != std::thread::id() && externalCallThreadId == std::this_thread::get_id())
+        *lockGuard = createWithImplementation<ILockGuard, LockGuardImpl<object_utils::NullMutex>>(this->objPtr, &nullSync, &externalCallThreadId, &externalCallDepth).detach();
+    else
+        *lockGuard = createWithImplementation<ILockGuard, LockGuardImpl<std::mutex>>(this->objPtr, &sync, &externalCallThreadId, &externalCallDepth).detach();
+    return OPENDAQ_SUCCESS;
+}
+
+template <typename PropObjInterface, typename... Interfaces>
+ErrCode GenericPropertyObjectImpl<PropObjInterface, Interfaces...>::getRecursiveLockGuard(IList* lockGuardList)
```
JakaMohorkoDS, line 2734: Should these two not be other way around?

core/opendaq/device/include/opendaq/device.h
```diff
@@ -353,6 +354,26 @@ DECLARE_OPENDAQ_INTERFACE(IDevice, IFolder)
      * @param[out] statusContainer The container for the device connection statuses.
      */
     virtual ErrCode INTERFACE_FUNC getConnectionStatusContainer(IComponentStatusContainer** statusContainer) = 0;
+
+    // [templateType(availableOpModes, IString, IInteger)]
+    /*!
+     * @brief Gets a list of available operation modes for the device.
+     * @param[out] availableOpModes The list of available operation modes.
+     */
+    virtual ErrCode INTERFACE_FUNC getAvailableOperationModes(IDict** availableOpModes) = 0;
```
JakaMohorkoDS, line 363: If this is a dictionary by intention, the doxygen comment is incorrect.

core/opendaq/device/include/opendaq/device.h
```diff
@@ -353,6 +354,26 @@ DECLARE_OPENDAQ_INTERFACE(IDevice, IFolder)
...
+
+    /*!
+     * @brief Sets the operation mode of the device and subtree.
+     * @param modeType The operation mode to set.
+     * @param includeSubDevices If true, the operation mode will be set for all sub-devices as well.
+     */
+    virtual ErrCode INTERFACE_FUNC setOperationMode(OperationModeType modeType, Bool includeSubDevices = true) = 0;
```
JakaMohorkoDS, line 370: Would we maybe want to split this into two methods? `setOperationMode` and `setOperationModeRecursive`?

It'd look nicer in Python.

core/opendaq/device/include/opendaq/device_impl.h
```diff
@@ -149,6 +156,11 @@ class GenericDevice : public SignalContainerImpl<TInterface, IDevicePrivate, IDe
     static ConstCharPtr SerializeId();
     static ErrCode Deserialize(ISerializedObject* serialized, IBaseObject* context, IFunction* factoryCallback, IBaseObject** obj);
 
+    // IComponentPrivate
+    ErrCode INTERFACE_FUNC updateOperationMode(OperationModeType modeType) override;
+
+    // IPropertyObjectInternal
+    ErrCode INTERFACE_FUNC getRecursiveLockGuard(IList* lockGuardList) override;
```
JakaMohorkoDS, line 163: Should this be an overridden method, or just an internal implementation that calls `getRecursiveConfigLock` on children?

core/coreobjects/include/coreobjects/property_object_internal.h
```diff
@@ -44,6 +49,10 @@ DECLARE_OPENDAQ_INTERFACE(IPropertyObjectInternal, IBaseObject)
     virtual ErrCode INTERFACE_FUNC getPropertySelectionValueNoLock(IString* name, IBaseObject** value) = 0;
     virtual ErrCode INTERFACE_FUNC setPropertyValueNoLock(IString* name, IBaseObject* value) = 0;
     virtual ErrCode INTERFACE_FUNC clearPropertyValueNoLock(IString* name) = 0;
+
+    virtual ErrCode INTERFACE_FUNC getLockGuard(ILockGuard** lockGuard) = 0;
+    // [templateType(lockGuard, ILockGuard)]
+    virtual ErrCode INTERFACE_FUNC getRecursiveLockGuard(IList* lockGuard) = 0;
```
JakaMohorkoDS, line 55: These two methods are a bit inconsistent in how they behave considering their naming.

- `PropertyObject::getLockGuard` creates a `lock_guard` with a non-recursive mutex
- `PropertyObject::getRecursiveLockGuard` creates a `lock_guard` with a recursive mutex
- `SignalContainer::getRecursiveLockGuard` creates a `lock_guard` with a recursive mutex for itself and all of its children
- `Device::getRecursiveLockGuard` creates a `lock_guard` with a recursive mutex for all of its non-device children.

I don't think that `Device/SignalContainer` should have a special implementation of these interface methods. A device should just collect all of its non-device subcomponents and call `getRecursiveLockGuard` on them.

```cpp
const auto items = this->objPtr.getItems(Recursive(Not(InterfaceId(IDevice::Id)));
ListPtr<ILockGuard> guards;

for (const auto& item : items)
    guards.pushBack(item.asPtr<IPropertyObjectInternal>().getRecursiveLockGuard();
```

core/opendaq/device/include/opendaq/device_impl.h
```diff
@@ -1068,6 +1085,131 @@ ErrCode GenericDevice<TInterface, Interfaces...>::getConnectionStatusContainer(I
...
+    if (includeSubDevices)
+    {
+        for (const DevicePtr & dev: this->devices.getItems())
+        {
+            errCode = dev->setOperationMode(modeType, includeSubDevices);
+            if (OPENDAQ_FAILED(errCode))
+                return errCode;
+        }
+    }
+    return OPENDAQ_SUCCESS;
```
JakaMohorkoDS, line 1202: We need a new core event here called `OperationModeChanged`. That should be triggered and used to notify native clients of the change.

core/opendaq/component/include/opendaq/folder_impl.h
```diff
@@ -235,25 +234,16 @@ ErrCode FolderImpl<Intf, Intfs...>::addItem(IComponent* item)
...
         const auto args = createWithImplementation<ICoreEventArgs, CoreEventArgsImpl>(
-                CoreEventId::ComponentAdded,
-                Dict<IString, IBaseObject>({{"Component", component}}));
+            CoreEventId::ComponentAdded,
+            Dict<IString, IBaseObject>({{"Component", component}}));
 
         this->triggerCoreEvent(args);
+        
+        if (auto componentPrivate = component.template asPtrOrNull<IComponentPrivate>(true); componentPrivate.assigned())
+            componentPrivate->syncOperationMode();
```
JakaMohorkoDS, line 245: This should be before `triggerCoreEvent` in all likelihood? At least for devices, as the client will receive invalid information about the op mode state. Either that, or it's called after `enableCoreEvent`.

JakaMohorkoDS (approved):
Need to double-check if setting `active` to false on `idle` is correct or not.

JakaMohorkoDS (approved):

PR #713, opened 2025-02-24
<TBBAS-2120> Add cmake option to disable access control (#708)

janm-dw:
# Brief

Add a CMake option to disable access control at compile time. Disabling access control can improve the startup performance of a device while its structure is being built.

# Usage example

In Cmake use `OPENDAQ_ENABLE_ACCESS_CONTROL=OFF` to disable access control. By default this option is set to `ON` to ensure backwards compatibility.

By default, access control is enabled, so no integration changes are required. If you disable access control, the PermissionManager object will always allow unrestricted access to all users for all property objects.


JakaMohorkoDS:
I would skip this for the main branch and keep it as a workaround in release. The performance issues are addressed by #714

PR #714, opened 2025-02-24
Permission manager optimizations

JakaMohorkoDS:
# Brief

Set of permission manager optimizations that reduce the number of Dictionary object creations on `Property`/`PropertyObject` construction.

# Description

- Define static permission objects for default initializations of the permission managers of `PropertyObject`, `Property`, and Component
- Replace `PermissionBuilder` dictionaries with `std::map`
- Only create `PermissionManager` on `Property` creation when creating an object-type `Property`

### DIFF
--- a/core/coreobjects/src/permissions_impl.cpp
+++ b/core/coreobjects/src/permissions_impl.cpp
@@ -11,14 +11,17 @@ PermissionsImpl::PermissionsImpl()
 }
 
 PermissionsImpl::PermissionsImpl(Bool inherited,
-                                 const DictPtr<IString, Int>& allowed,
-                                 const DictPtr<IString, Int>& denied,
-                                 const DictPtr<IString, Int>& assigned)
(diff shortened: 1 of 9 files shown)


core/coreobjects/src/permissions_impl.cpp
```diff
@@ -53,14 +56,10 @@ ErrCode INTERFACE_FUNC PermissionsImpl::getAssigned(IDict** permissions)
     return OPENDAQ_SUCCESS;
 }
 
-DictPtr<IString, Int> PermissionsImpl::cloneDict(const DictPtr<IString, Int>& dict)
+void PermissionsImpl::cloneDict(const std::unordered_map<StringPtr, Int, StringHash, StringEqualTo>& dict, DictPtr<IString, Int>& target)
```
janm-dw, line 59: It might be better rename this method to addAll()

janm-dw (approved):

PR #721, opened 2025-02-28
Logging improvement

denise-opendaq:
# Brief

- openDAQ stores not just the last error info message but a list of error messages, providing a traceback of the error.
- Implemented macros that enhance existing error info and exception-handling functions by adding details about the error's origin.

# Description
- **IErrorInfo** has a new method to get and set the file name and line number where the error occurred.
- The base openDAQ exception class, **DaqException**, also has a method to provide information about the exception's origin.
- The **errorFromException** method, which receives a **DaqException**, now forwards the file name and line number to the error info.
- For the methods **makeErrorInfo**, **errorFromException**, and **setErrorInfoWithSource**, macros have been implemented that, in debug builds, add the error's origin.
- In debug builds, the **OPENDAQ_PARAM_NOT_NULL** macro creates error info with a message specifying which argument is **null**.
- The method **checkErrorInfo** throws an exception with the error info traceback message

# Usage example

In C++ use:

```cpp
DECLARE_OPENDAQ_INTERFACE(IErrorTest, IBaseObject)
{
    ErrCode INTERFACE_FUNC exceptionExample() = 0;
};

class ErrorTestImpl : public ImplementationOf<IErrorTest>
{
public:
    ErrCode INTERFACE_FUNC exceptionExample(IString* someWord) override
    {
        OPENDAQ_PARAM_NOT_NULL(someWord);
        const ErrCode errCode = daqTry([this, &someWord] 
        {
            onExceptionExample(someWord);
        });

        if (OPENDAQ_FAILED(errCode))
            return DAQ_MAKE_ERROR_INFO(errCode, "exceptionExample failed");
        return OPENDAQ_SUCCESS;
    }

protected:
    virtual void onExceptionExample(const StringPtr& someWord)
    {
        if (someWord.getLength() == 0)
            DAQ_THROW_EXCEPTION(InvalidParameterException, "someWord must not be empty");

        // ... proccesing the word
    }
};

static ObjectPtr<IErrorTest> CreateTestObject()
{
    IErrorTest* testObject = new ErrorTestImpl();
    testObject->addRef();
    return testObject;
}

int main(int argc, char *argv[])
{
    auto obj = CreateTestObject();
    std::cout << " - Calling the method with nullptr" << std::endl;
    try 
    {
        checkErrorInfo(obj->exceptionExample(nullptr));
    }
    catch (const daq::DaqException& e)
    {
        std::cout << e.what() << std::endl;
    }

    std::cout << "\n - Calling the method with empty string" << std::endl;
    try
    {
        checkErrorInfo(obj->exceptionExample(String("")));
    }
    catch(const std::exception& e)
    {
        std::cerr << e.what() << '\n';
    }
}
```

The output will be:
``` bash
 - Calling the method with nullptr
Parameter someWord must not be null

Traceback (most recent call last):
[ main.cpp:11 ] : Parameter someWord must not be null

 - Calling the method with empty string
exceptionExample failed

Traceback (most recent call last):
[ main.cpp:26 ] : someWord must not be empty
[ main.cpp:18 ] : exceptionExample failed

```

# API changes

```diff
+ [function] IErrorInfo::setFileName(ConstCharPtr fileName)
+ [function] IErrorInfo::getFileName(ConstCharPtr* fileName)
+ [function] IErrorInfo::setFileLine(Int fileLine)
+ [function] IErrorInfo::getFileLine(Int* fileLine)

+ [function] daqGetErrorInfoList(daq::IList** errorInfoList)

+ [function] DaqException::getFileName() const
+ [function] DaqException::getFileLine() const

+ [macros] DAQ_ERROR_FROM_STD_EXCEPTION(const std::exception& e, IBaseObject* source, ErrCode errCode)
+ [macros] DAQ_THROW_EXCEPTION(Exception, ExceptionParams... params)
+ [macros] DAQ_MAKE_ERROR_INFO(ErrCode errCode, const std::string& message, Params... params)
```


euttsov-odaq (approved):

euttsov-odaq:
@denise-opendaq One note about passing template arguments pack of type `Params...`: maybe it is better to pass them by universal reference and forward the further with `std::forward<Params>(params)...` ?

core/coretypes/include/coretypes/ctutils.h
```diff
@@ -89,32 +89,110 @@ inline std::string objectToString(IBaseObject* object)
...
+                if (errorInfoObj != nullptr)
+                    errorInfoObj->releaseRef();
+            }
         }
 
-        throwExceptionFromErrorCode(errCode, msgCharPtr ? msgCharPtr : "");
+        if (errorInfoList != nullptr)
+            errorInfoList->releaseRef();
+
+        throwExceptionFromErrorCode(errCode, ss.str());
```
NikolaiShipilov, line 195: As I understand it, these changes include enabling additional information, such as the filename and line number where an error originates via the `makeErrorInfo` call.
Would it be feasible (perhaps not as part of this PR but as a potential future improvement) to provide similar details for error-codes returned by interface functions but not wrapped with error-Info, i.e., simply returned as `OPENDAQ_ERR_x`, as most interface-functions do?
In such cases, it could trace back to the location of `checkErrorInfo` call itself, which is typically in auto-generated `_ptr.h` code, so would require modifications to RTGen.

NikolaiShipilov: I mean that `checkErrorInfo` could provide filename, line (or maybe also a function name) using macroses , so this will be used within `_ptr.h` headers like this (e.g. in auto-generated `property_object_ptr.h`) :
```
    /*!
     * @brief Adds the property to the Property object.
     * @param property The property to be added.
     * @throws InvalidValueException if the property has no name.
     * @throws AlreadyExistsException if a property with the same name is already part of the Property object.
     * @throws FrozenException if the Property object is frozen.
     * 
     * The Property is frozen once added to the Property object, making it immutable. The same Property cannot
     * be added to multiple different Property objects.
     */
    void addProperty(const daq::PropertyPtr& property) const
    {
        if (this->object == nullptr)
            throw daq::InvalidParameterException();

        auto errCode = this->object->addProperty(property);
        daq::checkErrorInfo(errCode, __FILENAME__, __LINE__, __FUNCTION__);
    }
```
so the user will see not only smth like this 
`terminate called after throwing an instance of 'daq::AlreadyExistsException' ...`
but also information about where the error coming from - file, line or function name

core/coretypes/include/coretypes/exceptions.h
```diff
@@ -63,24 +63,30 @@ BEGIN_NAMESPACE_OPENDAQ
...
+    }                                                                   \
+    catch (...)                                                         \
+    {                                                                   \
+        return OPENDAQ_ERR_GENERALERROR;                                \
     }
 
+#ifdef NDEBUG
+    #define THROW_OPENDAQ_EXCEPTION(opendaqException) throw opendaqException
+#else
+    #define THROW_OPENDAQ_EXCEPTION(opendaqException) throw opendaqException.setFileName(__FILE__).setFileLine(__LINE__)
```
Unsttopabull, line 87: Shouldn't we follow the existing convention of macros being `DAQ_XYZ` so `DAQ_THROW_EXCEPTION`
Why the need for a builder? Isn't this just complicating it for no reason?
I'd just pass FILE and LINE as constructor parameters that can be defaulted to empty string if necessary.

core/corecontainers/include/coretypes/dictobject_iterator_impl.h
```diff
@@ -40,7 +40,7 @@ class DictIterator : public IteratorBaseImpl<T, IDictElementType, KeyValueSelect
     ErrCode INTERFACE_FUNC getKeyInterfaceId(IntfID* idKey) override
     {
         if (idKey == nullptr)
-            return this->makeErrorInfo(OPENDAQ_ERR_ARGUMENT_NULL, "Key-id output parameter must not be null.");
+            return this->MakeErrorInfo(OPENDAQ_ERR_ARGUMENT_NULL, "Key-id output parameter must not be null.");
```
Unsttopabull, line 43: I would UPPERCASE_NAME all macros WITHOUT exceptions. You can hide `this->` into the macro anyway.
But even there I wonder if we ever actually check the `source` object and if this needs to be a member function.

Unsttopabull (approved):

PR #728, opened 2025-03-10
Add scan timeout in `createDevice`

JakaMohorkoDS:
# Brief
Add timeout on to re-scan after for available devices after 5s in the module manager call `createDevice`.

# Description
The openDAQ module manager scans for available devices before attempting to create a device. This is done to gather all device capabilities for configuration/streaming, and to allow adding a device via the connection string format "daq://Manufacturer_SerialNumber".

This was only done if `getAvailableDevices` has not been called at all in the lifetime of the module manager. This PR changes said behaviour by introducing a 5s  timeout post `getAvailableDevices`. If a device is created after said timeout, the module manager now re-scans before creating the device.

# Usage example

When re-booting a device with manufacturer "openDAQ" and serial "SN123",  the following code snippet previously never connected, if the device was not yet discoverable during the first loop iteration:

```cpp
bool connected = false;
while (!connected)
{
    try
    {
        auto dev = instance.addDevice("daq://openDAQ_SN123")
        connected = dev.assigned();
    }
    catch(...)
    {
        using namespace std::chrono_literals;
        std::cout << "Failed to connect, retrying..." << std::endl;
        std::this_tread::sleep_for(0.5s);
    }
}
```

### DIFF
--- a/changelog/changelog_3.10-3.20.md
+++ b/changelog/changelog_3.10-3.20.md
@@ -2,7 +2,6 @@
 
 ## Features
 
-- [#719](https://github.com/openDAQ/openDAQ/pull/719) Fixes error when accessing selection property values using "dot" notation (eg. `getPropertySelectionValue("child.val")`).
 - [#718](https://github.com/openDAQ/openDAQ/pull/718) Adds new Native Configuration Protocol RPCs for handling sub-function blocks (function blocks that are children of other FBs.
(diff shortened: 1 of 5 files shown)


dejancrn:
It would be nice if scan timeout would be a configuration parameter.

JakaMohorkoDS:
> It would be nice if scan timeout would be a configuration parameter.

Agreed that this should be a configuration parameter, it's just a bit tricky, as the module manager does not have access to the context. I'll see what I can do to get that in.

denise-opendaq (approved):

PR #729, opened 2025-03-10
Python GUI app: attributes dialog, changeable device info, metadata fields fix

aljazfrancic:
# Brief

Improve Python GUI app: attributes dialog update, device info via generic properties treeview, metadata fields fix

# Description

- Attributes dialog update
  - Move all logic for topmost TreeView in `attributes_dialog.py` to `generic_attributes_treeview.py`
  - Use `generic_properties_treeview.py` for `additional_tree_frame` if `daq.IDevice.can_cast_from(node)` allowing for modifications
  - Introduce `data_descriptor_treeview.py` and use it for `additional_tree_frame` if `daq.ISignal.can_cast_from(node)` (it's read only)
  - Set parent for `simpledialog.ask...` to prevent pop-ups under current window
- Enable nested property objects in device info dialog via generic properties treeview
- Fix metadata fields not showing up
- Fix crash if no `related_signals`
- Fix typos, minor fixes

### DIFF
--- a/examples/python/gui_demo/components/data_descriptor_treeview.py
+++ b/examples/python/gui_demo/components/data_descriptor_treeview.py
@@ -1,0 +1,72 @@
+import tkinter as tk
+from tkinter import ttk
+
+import opendaq as daq
+
+from .. import utils
+from ..app_context import AppContext
+
+import opendaq
+
+class DataDescriptorTreeview(ttk.Treeview):
+    def __init__(self, parent, data_descriptor : opendaq.IDataDescriptor = None, context: AppContext = None, **kwargs):
+        ttk.Treeview.__init__(self, parent, columns=('value', 'access', *context.metadata_fields), show='tree headings', **kwargs)
+
+        self.context = context
+
+        scroll_bar = ttk.Scrollbar(
+            self, orient=tk.VERTICAL, command=self.yview)
+        self.configure(yscrollcommand=scroll_bar.set)
+        scroll_bar.pack(side=tk.RIGHT, fill=tk.Y)
+        scroll_bar_x = ttk.Scrollbar(
+            self, orient=tk.HORIZONTAL, command=self.xview)
+        self.configure(xscrollcommand=scroll_bar_x.set)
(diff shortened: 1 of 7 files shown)


examples/python/gui_demo/components/generic_attributes_treeview.py
```diff
@@ -0,0 +1,232 @@
...
+
+import opendaq as daq
+import os
+
+from .. import utils
+from ..app_context import AppContext
+from ..event_port import EventPort
+
+class AttributesTreeview(ttk.Treeview):
+    def __init__(self, parent, frame, node=None, context: AppContext = None, **kwargs):
```
aljazfrancic, line 12: I wasn't sure how to avoid having both `parent` (used for `EventPort`, to properly update the parent GUI when there is a new value and we call `emit`) and `frame` (used as the parent of the Treeview).

examples/python/gui_demo/components/data_descriptor_treeview.py
```diff
@@ -0,0 +1,72 @@
+import tkinter as tk
+from tkinter import ttk
+
+import opendaq as daq
+
+from .. import utils
+from ..app_context import AppContext
+
+import opendaq
```
Yarikk26, line 9: redundant import here

examples/python/gui_demo/components/metadata_fields_selector_dialog.py
```diff
@@ -1,6 +1,8 @@
 import tkinter as tk
 from tkinter import ttk
 
+import opendaq
```
Yarikk26, line 4: `import opendaq as daq` just for consistency

examples/python/gui_demo/components/generic_attributes_treeview.py
Yarikk26: I had the same concerns about similar case somewhere inside the app, but it was working ok. `EventPort` should be replaced with `CoreEvents` `Event`ually :)

Yarikk26 (approved):
Looks ok in general

PR #730, opened 2025-03-11
Provide list of connected clients via DeviceInfo

NikolaiShipilov:
# Brief

Provides the information about connected clients within the `DeviceInfo`.
Fixes issue with the device dropping all native streaming clients when `ExclusiveControl` client with `DropOthers` flag connected.

# Description

Introduces `IConnectedClientInfo` interface to set and access the information of connected client; such as client address, host name, connection protocol type and name, client connection type ("Control", "ExclusiveControl", "ViewOnly") if supported by protocol, and additional custom string-type properties. The mentioned information come with the following limitations depending on the client connection protocol:
  - Client address is unavailable for OPC UA.
  - Only IP address without port number is available for LT.
  - Hostname is available only for native protocol clients (streaming and configuration), excluding legacy versions.

The `ConnectedClientInfo` object is derived from a `PropertyObject`, supports only string-type properties, and defines a set of default ones.
This object is transmitted as a set of TXT records over the mDNS discovery protocol using the following format (key = value):
```
connectedClient--<clientId>--Address = 192.168.1.103:44032  
connectedClient--<clientId>--ProtocolName = OpenDAQNativeConfiguration  
connectedClient--<clientId>--HostName = hostname  
connectedClient--<clientId>--ClientType = Control  
connectedClient--<clientId>--ProtocolType = Configuration  
connectedClient--<clientId>--<customProp> = customValue
```
Where:
- `connectedClient--` is a required prefix for keys.
- `<clientId>` is a unique string identifier for each connected client. In the openDAQ discovery server implementation, numeric IDs (e.g., 00, 01, ...) are used for simplicity, with zero-padding applied when the client count exceeds 10.
- The final part of the key (after `connectedClient--<clientId>--`) specifies the property name, which may be one of the 5 predefined properties or a custom-defined one.

Restrictions:
- `<clientId>` and property names must not contain non-printable ASCII characters or the `=` symbol to comply with mDNS TXT record limitations.
- The combined length of the TXT key and its corresponding value must not exceed 254 bytes.
- Property values containing non-printable ASCII characters or the `=` symbol will have those characters replaced with a space (' ') prior to being encoded into the TXT record.

> [!NOTE]
> The information about connected clients is also propagated by the device through the native configuration protocol, allowing it to be retrieved on the client side from the DeviceInfo of a connected native configuration device. However, this feature is currently not supported over the OPC UA protocol.

The native configuration protocol version is incremented to 11, and the openDAQ serializer version is increased to 3. If the native protocol version is 10, serializer version 2 is used; for versions below 10, serializer version 1 is applied.

# Usage example

In C++ use:

To get the detailed information about client already connected to discovered device
```cpp
auto deviceInfo = instance.getAvailableDevices()[0];
auto connectedClientInfo = deviceInfo.getConnectedClientsInfo()[0];

std::cout << "Connected client address: " << connectedClientInfo.getAddress() << std::endl;
std::cout << "Connected client host name: " << connectedClientInfo.getHostName() << std::endl;
std::cout << "Connected client connection protocol name: " << connectedClientInfo.getProtocolName() << std::endl;
std::cout << "Connected client type: " << connectedClientInfo.getClientTypeName() << std::endl;
```

# Required module changes
Updates the openDAQ discovery client API and base Module implementation, providing helper method populateDiscoveredDeviceInfo to create discovered DeviceInfo and populate its properties.

Therefore, the following module implementation snippet should be modified as follows:

```diff
DeviceInfoPtr ExampleClientModule::populateDiscoveredDevice(const MdnsDiscoveredDevice& discoveredDevice)
{
-   PropertyObjectPtr deviceInfo = DeviceInfo("");
-   DiscoveryClient::populateDiscoveredInfoProperties(deviceInfo, discoveredDevice);

    auto cap = ServerCapability(...);
    ...

-   deviceInfo.asPtr<IDeviceInfoInternal>().addServerCapability(cap);
-   deviceInfo.asPtr<IPropertyObjectProtected>().setProtectedPropertyValue("connectionString", cap.getConnectionString());
-   deviceInfo.asPtr<IDeviceInfoConfig>().setDeviceType(createDeviceType());

-   return deviceInfo;

+   return populateDiscoveredDeviceInfo(DiscoveryClient::populateDiscoveredInfoProperties, discoveredDevice, cap, createDeviceType());
}
```

# API changes

> [!NOTE]
> Breaks binary compatibility of module shared libraries

```diff
+ [interface] IConnectedClientInfo : public IPropertyObject
+ [function]  IConnectedClientInfo::getAddress(IString** address)
+ [function]  IConnectedClientInfo::getProtocolType(ProtocolType* type)
+ [function]  IConnectedClientInfo::getProtocolName(IString** protocolName)
+ [function]  IConnectedClientInfo::getClientTypeName(IString** type)
+ [function]  IConnectedClientInfo::getHostName(IString** hostName)
+ [factory] ConnectedClientInfoPtr ConnectedClientInfo(const StringPtr& address, ProtocolType protocolType, const StringPtr& protocolName, const StringPtr& clientType, const StringPtr& hostName)
+ [factory] ConnectedClientInfoPtr ConnectedClientInfo()

+ [function]  IDeviceInfo::getConnectedClientsInfo(IList** connectedClientsInfo)
+ [function]  IDeviceInfoInternal::addConnectedClient(SizeT* clientNumber, IConnectedClientInfo* clientInfo)
+ [function]  IDeviceInfoInternal::removeConnectedClient(SizeT clientNumber)
```


core/opendaq/device/include/opendaq/connected_client_info.h
```diff
@@ -0,0 +1,94 @@
...
+ * [interfaceLibrary(IPropertyObject, "coreobjects")]
+ * [interfaceSmartPtr(IPropertyObject, GenericPropertyObjectPtr, "<coreobjects/property_object_ptr.h>")]
+ */
+DECLARE_OPENDAQ_INTERFACE(IConnectedClientInfo, IPropertyObject)
+{
+    /*!
+     * @brief Gets the client URL string.
+     * @param[out] url The client URL string.
+     */
+    virtual ErrCode INTERFACE_FUNC getUrl(IString** url) = 0;
```
JakaMohorkoDS, line 40: To keep it consistent with our other interfaces, should this be named `getAddress`?

shared/libraries/config_protocol/include/config_protocol/config_client_property_object_impl.h
```diff
@@ -861,8 +861,9 @@ void ConfigClientPropertyObjectBaseImpl<Impl>::checkCanSetPropertyValue(const St
 template <class Impl>
 bool ConfigClientPropertyObjectBaseImpl<Impl>::isBasePropertyObject(const PropertyObjectPtr& propObj)
 {
-    return !propObj.supportsInterface<IServerCapabilityConfig>() 
-            && !propObj.supportsInterface<IAddressInfo>();
+    return !propObj.supportsInterface<IServerCapabilityConfig>()
+            && !propObj.supportsInterface<IAddressInfo>()
+            && !propObj.supportsInterface<IConnectedClientInfo>();
```
JakaMohorkoDS, line 866: In my PR #731  I'm also adding this code to the OPC UA client property object. So likely needs to be changed there/here after one has been merged.

JakaMohorkoDS (commented):
Lgtm! One thing that we need is the TXT record description for a 3rd party server implementation to be able to replicate this. Just what fields are required and how to set them up.

Even a utility method somewhere in our discovery internals would be useful that converts the connected client info into a dictionary of TXT records.

NikolaiShipilov:
> One thing that we need is the TXT record description for a 3rd party server implementation to be able to replicate this. Just what fields are required and how to set them up.

Where would be the best place to include this description ? Adding it as a separate [specification](https://docs-dev.opendaq.com/manual/opendaq-specs/dev/overview.html) for the discovery protocol (importing most of the information from Confluence pages) seems like a reasonable approach.

However, I would maybe also reconsider this conversion to make it more straightforward for 3rd-party implementations. Instead of serializing the openDAQ object and then splitting the resulting string into appropriately sized chunks for the TXT record (which adds extra complexity by involving the serialization step) we could encode the nested properties in a format like this instead (`key`=`value`):
`connectedClientsInfo--NESTED--clientId--Address`=`192.168.0.1`
`connectedClientsInfo--NESTED--clientId--ProtocolName`=`OpenDAQNativeConfiguration`

Though, this will likely constrain the nested properties type of `ConnectedClientInfo` (including custom ones) to Strings only. But, anyway, all default properties are already of String type.

aljazfrancic (approved):

JakaMohorkoDS (approved):

PR #734, opened 2025-03-17
[TBBAS-2132] Config objects are not saved/loaded from JSON

denise-opendaq:
# Brief

Enables storing of the component initialization config for the device and function block to use it while saving configuration.

# Description

- Previously, the saved tree configuration did not contain the component config, which is used when creating components such as devices and function blocks. Moreover, the device config contains additional properties that are used for setting up a streaming connection. Now, the configuration object used for creating devices and function blocks is saved within the save/load mechanism with help of `setComponentConfig` method in `IComponentPrivate`. This ensures that creating a component behaves exactly the same as during its initial creation, including properties for the module manager. Saving and loading with configuration now works both locally and via native methods. In OPC UA, configuration applies only to function blocks, as other components cannot be added or removed from the client side.

- Fixed an issue with loading function blocks via OPC UA. Previously, when adding a function block, the old one was removed only on the client side without invoking the corresponding method on the server side. Now, function block removal is processed correctly. Additionally, there was an issue when passing the config: while loading a function block, the client was passing the FB **local ID** property to the config. The property object was transferred via OPC UA as a dictionary, and the server applied new parameters to the default FB config. However, because the local ID was not part of the default config, setting the new property was failing.

- Corrects the implementation type used during device deserialization in the native configuration protocol. Previously, all mirrored devices created during deserialization used the custom root device implementation if provided. Now, the custom implementation is applied only to the root device, while nested devices use the default implementation.

- Adds setter and getter methods for component configuration to `IComponentPrivate`, deprecating similar methods from `IDevicePrivate`.

- Introduces the `"GetComponentConfig"` RPC and updates the native configuration protocol version to 13.

- Prevents mirrored devices from being set as the client root.

- Fixes Doxygen comments for `IDeviceInfo` and `IStreamingType`.

- Enables completion of the ServerCapabilities and ConnectInfo in DeviceInfo of updated native device, including the merging it with info obtained from device discovery utilizing new helper method `completeDeviceCapabilities` of the `IModuleManagerUtils` interface.

- Reworks configuration properties population for added devices, ensuring completeness of saved configuration object.

- Moves streaming setup initialization and modification (handled via core events) for mirrored devices and their nested components from `ModuleManager` and client modules (native and OPC UA) into the `MirroredDeviceBase` core implementation.

- Enhances **OPENDAQ_PARAM_NOT_NULL** macro to provide error information (including function and parameter names) in release builds.

- Renames namespace `daq::detail` to `daq::details` in `core/opendaq/utility/include/opendaq/mem_pool_allocator.h` to avoid conflicts

- Enhances streaming setup tests for subdevices.

# Usage example

To get function block config use:
```cpp
FunctionBlockPtr fb = instance.getFunctionBlocks()[0];
auto config = fb.asPtr<IComponentPrivate>().getComponentConfig();
```

To get device config use:
```cpp
DevicePtr device = instance.getDevices()[0];
auto config = device.asPtr<IComponentPrivate>().getComponentConfig();
```

# API changes
> breaks binary compatibility

```diff
+ [function] IComponentPrivate::setComponentConfig(IPropertyObject* config)
+ [function] IComponentPrivate::getComponentConfig(IPropertyObject** config)
+ [function] IModuleManagerUtils::completeDeviceCapabilities(IDevice* device)
```

# Required module changes

None, however methods of `IDevicePrivate` used to manage device initialization config are deprecated now and similar methods from `IComponentPrivate` is to be used instead.

Instead of:

```cpp
PropertyObjectPtr config = device.asPtr<IDevicePrivate>(true).getDeviceConfig();
```

Do:

```cpp
PropertyObjectPtr config = device.asPtr<IComponentPrivate>(true).getComponentConfig();
```


core/coretypes/include/coretypes/validation.h
```diff
@@ -31,7 +31,7 @@
 
 #ifdef NDEBUG
     #define OPENDAQ_PARAM_NOT_NULL(param) \
-        do { if (nullptr == (param)) return OPENDAQ_ERR_ARGUMENT_NULL; } while (0)
+        do { if (nullptr == (param)) return DAQ_MAKE_ERROR_INFO(OPENDAQ_ERR_ARGUMENT_NULL, "Parameter %s must not be null in the function \"%s\"", #param, __func__); } while (0)
```
JakaMohorkoDS, line 34: Good change

core/opendaq/signal/include/opendaq/signal_container_impl.h
```diff
@@ -38,14 +38,16 @@ class GenericSignalContainerImpl : public ComponentImpl<Intf, Intfs ...>
                                const ComponentPtr& parent,
                                const StringPtr& localId,
                                const StringPtr& className = nullptr,
-                               const StringPtr& name = nullptr);
+                               const StringPtr& name = nullptr,
+                               const PropertyObjectPtr& config = nullptr);
```
JakaMohorkoDS, line 42: I would set the config in the module manager via an internal interface method. Adding this here requires fb developers to always pass this parameter up to the parent constructor which is something we want to avoid.  This is incompatible with current implementations of fbs, and is a hidden change that will be hard to spot when upgrading openDAQ.

As far as I can see, that's already done for devices via `set/getDeviceConfig` (though that so far was never used from what I can tell). We should likely just add `set/getComponentConfig` to `IComponentPrivate`, though I'd prefer it to be on some internal interface.

JakaMohorkoDS (changes requested):

NikolaiShipilov (approved):

core/opendaq/modulemanager/include/opendaq/module_impl.h
```diff
@@ -179,8 +179,9 @@ class Module : public ImplementationOf<IModule>
         FunctionBlockPtr block;
         errCode = wrapHandlerReturn(this, &Module::onCreateFunctionBlock, block, id, parent, localId, mergeConfig(config, type));
 
-        if (const auto & componentPrivate = block.asPtrOrNull<IComponentPrivate>(true); componentPrivate.assigned())
-            componentPrivate.setComponentConfig(config);
+        if (block.assigned())
```
denise-opendaq, line 182: if the block is not assigned, asptrornull will return empty ptr

denise-opendaq (commented):
i think we can merge it :)

PR #737, opened 2025-03-18
Set thread names

dejancrn:
# Brief

Adds `daqNameThread`, which sets a thread's name. This is helpful during debugging and resource monitoring. It also sets the name of some threads.

# Description

Debuggers usually have a "threads" view that displays a list of threads. However, it can be challenging to identify them. Most debuggers and resource monitoring tools can also print the thread's name. `daqNameThread` must be called within the context of the target thread.

# Usage example

In C++ use:

```cpp

#include <opendaq/thread_name>

std::thread([] {
    // ...
    daqNameThread("CustomThread");
    // ...
}
```
# API Changes

```diff
+ daqNameThread(const char* name)
```

### DIFF
--- a/core/opendaq/logger/src/logger_thread_pool_impl.cpp
+++ b/core/opendaq/logger/src/logger_thread_pool_impl.cpp
@@ -1,11 +1,10 @@
 #include <opendaq/logger_thread_pool_impl.h>
-
+#include <opendaq/thread_name.h>
 #include <coretypes/impl.h>
 
 BEGIN_NAMESPACE_OPENDAQ
 
-LoggerThreadPoolImpl::LoggerThreadPoolImpl() :
-    spdlogThreadPool(std::make_shared<ThreadPool>(8192, 1, []{}))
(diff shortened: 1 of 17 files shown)


denise-opendaq (approved):
i was just thinking about this a week ago:) looks great!

NikolaiShipilov (approved):
lgtm. Would it be possible to merge it somehow with the [`shared/libraries/utils/src/thread_ex.cpp`](https://github.com/openDAQ/openDAQ/blob/main/shared/libraries/utils/src/thread_ex.cpp#L117) to eliminate code duplication ?

dejancrn:
> lgtm. Would it be possible to merge it somehow with the [`shared/libraries/utils/src/thread_ex.cpp`](https://github.com/openDAQ/openDAQ/blob/main/shared/libraries/utils/src/thread_ex.cpp#L117) to eliminate code duplication ?

I thought about that too. However, "utils" library does not link to opendaq lib, therefore we cannot use that code in opendaq shared library. On the other hand, I cannot include utils in opendaq as it is declared after opendaq. I'm sure it can be done I just did not want to mess with CMake and dependencies too much.

PR #738, opened 2025-03-19
Set default value as args newValue on clear instead of empty ptr

JakaMohorkoDS:
# Brief
When clearing a property value, the `PropertyValueArgs` object obtained in the `onWrite` callback now contains the default value instead of an empty ptr.

# Description
When a `onWrite` event was previously triggered by a `clearPropertyValue` call, the `args.getValue()` method would return an empty ptr. Users needed to always check for `args.getPropertyEventType()` being `Clear` to know whether or not the args contain a newly written value.

With this change, the same code can be used whether the event is `Clear` or `Write`, as the args value on `Clear` will contain the default property value.

# Usage example

```cpp
propObj.getOnPropertyValueWrite("FloatProperty") +=
	[](PropertyObjectPtr& /*sender*/, PropertyValueEventArgsPtr& args)
	{
              // Works regardless of PropertyEventType
              valueChanged(args.getValue());
	};
```

### DIFF
--- a/core/coreobjects/include/coreobjects/property_object_impl.h
+++ b/core/coreobjects/include/coreobjects/property_object_impl.h
@@ -718,6 +718,7 @@ ErrCode GenericPropertyObjectImpl<PropObjInterface, Interfaces...>::callProperty
                                                                                            bool isUpdating)
 {
     const auto name = prop.getName();
+    const auto defaultValue = prop.getDefaultValue();
 
     if (!updatePropertyStack.registerPropertyUpdating(name, newValue))
         return OPENDAQ_IGNORED;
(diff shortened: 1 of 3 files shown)


core/coreobjects/include/coreobjects/property_object_impl.h
```diff
@@ -768,6 +774,9 @@ ErrCode GenericPropertyObjectImpl<PropObjInterface, Interfaces...>::callProperty
     if (shouldUpdate)
     {
         // setting the final value is only done in the top level of the stack
+        if (changeType == PropertyEventType::Clear && args.getValue() == defaultValue)
```
denise-opendaq, line 777: set new values as the default

JakaMohorkoDS: PropertyEventType::Clear is set only when calling `clearPropertyValue`. In clear, if `newValue` stays unassigned, the value will simply be deleted from the map of local values. `args.getValue() == defaultValue` in this case means that the written value was not overridden, as such we should stick to the default behaviour and clear the local value.

```cpp
BaseObjectPtr newVal;
const ErrCode err = callPropertyValueWrite(prop, newVal, PropertyEventType::Clear, isUpdating);

if (OPENDAQ_FAILED(err))
    return err;
                
if (err == OPENDAQ_IGNORED)
    return OPENDAQ_SUCCESS;
                
if (!newVal.assigned())
{
    auto it = propValues.find(prop.getName());
    propValues.erase(it);
}
```

denise-opendaq (approved):

PR #743, opened 2025-03-25
high-performance WebSocket streaming server

dvnrrs:
# Brief

Implement the `new_websocket_streaming_server_module`.

# Description

This PR adds the high-performance implementation of the WebSocket Streaming Server. The module exists side-by-side with the pre-existing server, and clients may choose which server to instantiate. This implementation was developed by HBK to address some performance and reliability concerns with the older server. More detailed documentation on the design goals can be found at https://github.com/openDAQ/openDAQ/blob/norris/hbk-ws-streaming-server/modules/new_websocket_streaming_server_module/docs/index.md .

To instantiate, call `addServer("OpenDAQNewLTStreaming", nullptr)`.

After this module has been tested for a release cycle and validated that there are no critical bugs or feature parity issues with the older server, the hope is that the older server can be removed and replaced with this one.


aljazfrancic:
I've opened #774, which partially addresses problems encountered while building the new High-Performance WebSocket Streaming Server Module.

Fixed:
* `Boost::asio` in CMake
* `strcasecmp`

Still pending:
* `htole32` problem fix (seems system specific)
* `packed` problem fix (?) 
* Don't use designated initializers (C++20 feature)
* Changelog entry missing
* Supplementary image missing from `modules/new_websocket_streaming_server_module/docs/index.md`
* Consider integrating module doc into Antora, where it won't get lost

dvnrrs:
> I've opened #774, which partially addresses problems encountered while building the new High-Performance WebSocket Streaming Server Module.
> 
> Fixed:
> 
>     * `Boost::asio` in CMake
> 
>     * `strcasecmp`
> 
> 
> Still pending:
> 
>     * `htole32` problem fix (seems system specific)
> 
>     * `packed` problem fix (?)
> 
>     * Don't use designated initializers (C++20 feature)
> 
>     * Changelog entry missing
> 
>     * Supplementary image (`class-diagram`) missing from `modules/new_websocket_streaming_server_module/docs/index.md`
> 
>     * Consider integrating module doc into Antora, where it won't get lost

I think a commit of mine must have gotten lost somewhere, because most (not all) of these points I'd already fixed. I'll have a look.

modules/CMakeLists.txt
```diff
@@ -43,6 +43,7 @@ if (DAQMODULES_OPENDAQ_SERVER_MODULE)
     if (OPENDAQ_ENABLE_WEBSOCKET_STREAMING)
         message(STATUS "    - Websocket server module")
         add_subdirectory(websocket_streaming_server_module)
+        add_subdirectory(new_websocket_streaming_server_module)
```
NikolaiShipilov, line 46: @dvnrrs , as I recall from our discussion with @JakaMohorkoDS, a CMake option should be added to include or exclude the new module from the build and mark it as experimental for now. I assume you're already aware of this.

Also, please add a brief description of this CMake option to https://github.com/openDAQ/openDAQ/blob/main/CMake-Options.md

dvnrrs: > @dvnrrs , as I recall from our discussion with @JakaMohorkoDS, a CMake option should be added to include or exclude the new module from the build and mark it as experimental for now. I assume you're already aware of this.
> 
> Also, please add a brief description of this CMake option to https://github.com/openDAQ/openDAQ/blob/main/CMake-Options.md

I wasn't aware of this, but can certainly add such an option.

external/streaming_protocol/CMakeLists.txt
```diff
@@ -6,8 +6,8 @@ endif()
 
 opendaq_dependency(
     NAME                streaming_protocol
-    REQUIRED_VERSION    1.2.2
+    REQUIRED_VERSION    1.2.4
     GIT_REPOSITORY      https://github.com/openDAQ/streaming-protocol-lt.git
-    GIT_REF             v1.2.2
+    GIT_REF             3b56a7a066b059e7eb1692de5f6704815200a222
```
NikolaiShipilov, line 11:
```
-    GIT_REF             3b56a7a066b059e7eb1692de5f6704815200a222
+    GIT_REF             v1.2.4

File: modules/new_websocket_streaming_server_module/src/websocket_client_negotiating.cpp [hunk 3]
```diff
@@ -0,0 +1,122 @@
...
+    boost::system::error_code ec;
+
+    std::size_t bytes_read = socket.read_some(boost::asio::buffer(read_buffer), ec);
+
+    if (ec)
+    {
+        if (ec == boost::asio::error::would_block)
+            return true;
+
+        std::cerr << "[ws-streaming] client (negotiating): recv returned errno " << ec << std::endl;
```
NikolaiShipilov, line 28: Here and elsewhere, it would be preferable to use the built-in openDAQ logger instead of `std::cerr`. 
It also could help to make code more verbose if needed  by supporting various log levels such as warning, info, debug, and trace, not just errors.

PR #745, opened 2025-03-31
[TBBAS-1957] .NET bindings usage guide on Linux

Daniel-BB:
# Brief

General update of the quick-start documentation for usage of .NET on Windows and Linux. 

# Description

General alignment of the .NET examples to the C++ and Python examples.
Now the .NET examples and commands can be run from both, Windows and Linux console (tested with WSL/Windows Subsystem for Linux).
There is also a small fix regarding the preloading of openDAQ libraries under Linux (in the static constructor of the factories) so that the dependencies can be found.

# Usage example

n/a

# API changes

n/a

n/a

# Required module changes

n/a


Yarikk26 (approved):
Looks ok

PR #752, opened 2025-04-02
Fix switch in EvalValue decoding the default value wrong

WolfgangPretl:
# Brief

Invalid behavior with default values in switch statements inside EvalValue.

# Description

If there is an default value in the switch statement for EvalValue like in
`EvalValue($Measurment, 0, %PropertyA, 1, %PropertyB, %PropertyDefault)`
the current solution will have it's size set to 5 and will check the pair of [0,1], [2,3] and [4,5]. The last one will either fail (as the actual value can't be compared to the property reference) or will read from an invalid index.

This will only check the correct number of available pairs (size of valueNodes divided by two will output the correct number of pairs. As the default value will be just rounded down).


JakaMohorkoDS:
@WolfgangPretl Apologies for the late response. Did we ever support default values in switch statements of evals?

WolfgangPretl:
At least it was decoded in the openDAQ code that way (right below the changes I made for the loop), as if the number of nodes is odd, there is no key-value pair but only a value - which is used, if no match could be found.
```
    if (valueNodes->size() % 2 == 1)
        return valueNodes->back()->getResult();
```

And I've seen it several times being used on object models, but never for referenced properties.

WolfgangPretl:
@adolenc As mentioned above: It's basically fixing accessing an invalid index. Before this change, it was already prepared to return the last value in case there is an uneven number of values for the case selection (as all cases are in pairs `case, value, case, value`).

I don't know, how the switch statement is supposed to work in openDAQ, as I never found a proper documentation for EvalValue regarding allowed syntax and priorities. But I find it highly useful, to have the option to have a default value already prepared:
`switch(expression, case, value, case, value)`
works as well (and will throw an error if no case has been found)
`switch(expression, case, value, case, value, default)`
this just helps to provide at least a properly defined value.

PR #753, opened 2025-04-02
Fix reserved struct type deserialization

JakaMohorkoDS:
# Brief

Fixes crash when deserializing struct types that have a type name that's present in the list of protected type names.

# Description

The native streaming protocol serializes the struct types, and deserializes them on the client. On deserialization, the type adds itself back to the type manager, resulting in a crash when the deserialized struct type is one of the protected ones (eg. range, complex number). This PR prevents said crash.


denise-opendaq (approved):

aljazfrancic (approved):
LGTM!

PR #757, opened 2025-04-04
Merge discovery info and device info based on SN and manufacturer

JakaMohorkoDS:
# Brief

Device info obtained from device and discovery matching is changed from checking for connection string equality to SN + manufacturer equality.

# Description

When connecting to a device, after connection, the capabilities of the device made known through discovery and those published on the device's info object are merged. Previously, we compared connection strings of server capabilities to check which discovered device info matches the device we connected to. This caused issues when equivalent connection strings that were not completely equal were used (eg. when a connection string contained a trailing slash). 

As such, we now use the combination of serial number and manufacturer to check which discovered device info belongs to the device we connected to.

Ensure that the same connection protocol is used for streaming as for configuration. For example, if the server connection uses an IPv6 connection string, the streaming connection should also use IPv6.

If the connection string used differs from the connection strings found via discovery, the address info corresponding to the manually used connection string will be stored in the server capability.

# Usage example

The below code previously failed to merge the discovery information with the connected-to device's info object, as the connection string of the capability differed by a trailing slash from the one obtained through discovery.

```cpp
auto instance = Instance();
auto capability = instance.getAvailableDevices()[0].getServerCapabilities()[0];

// Returns "daq.nd://127.0.0.1/"
auto connStr = capability.getConnectionString(); 

// Connection string differs from the one obtained through discovery by the trailing slash.
// Merge of info objects now works, as SN + manufacturer is used for checking if the info belongs
// to the same device.
auto device = instance.addDevice("daq.nd://127.0.0.1");
```


modules/native_streaming_client_module/src/native_streaming_client_module_impl.cpp
```diff
@@ -429,6 +429,17 @@ DevicePtr NativeStreamingClientModule::onCreateDevice(const StringPtr& connectio
...
+    for (const auto& capability : device.getInfo().getServerCapabilities())
+    {
+        if (capability.getProtocolId() == protocolId)
+        {
+            auto capConf = capability.asPtr<IServerCapabilityConfig>(true);
+            capConf.addAddressInfo(addressInfo);
+            capConf.addConnectionString(connectionString);
+            break;
+        }
+    }
```
JakaMohorkoDS, line 441: Shouldn't this be done in `onCompleteServerCapabilities`?

modules/opcua_client_module/src/opcua_client_module_impl.cpp
```diff
@@ -115,18 +114,23 @@ DevicePtr OpcUaClientModule::onCreateDevice(const StringPtr& connectionString,
                   .addAddressInfo(addressInfo)
                   .freeze();
 
+    completeServerCapabilities(device, addressInfo);
+
     return device;
 }
 
-void OpcUaClientModule::completeServerCapabilities(const DevicePtr& device, const StringPtr& deviceAddress)
+void OpcUaClientModule::completeServerCapabilities(const DevicePtr& device, const AddressInfoPtr& deviceAddress)
```
denise-opendaq, line 122: have no clue why we need this

modules/native_streaming_client_module/src/native_streaming_client_module_impl.cpp
```diff
@@ -665,14 +662,18 @@ Bool NativeStreamingClientModule::onCompleteServerCapability(const ServerCapabil
     }
     
     const auto path = target.hasProperty("Path") ? target.getPropertyValue("Path") : "";
+    const auto targerAddress = target.getAddresses();
```
JakaMohorkoDS, line 664: Typo

modules/native_streaming_client_module/src/native_streaming_client_module_impl.cpp
```diff
@@ -82,7 +82,7 @@ void NativeStreamingClientModule::SetupProtocolAddresses(const MdnsDiscoveredDev
             protocolPrefix,
             discoveredDevice.ipv4Address,
             discoveredDevice.servicePort,
-            discoveredDevice.getPropertyOrDefault("path", "/")
+            discoveredDevice.getPropertyOrDefault("path", "")
```
JakaMohorkoDS, line 84: Is there a reason to remove the path? I think this is likely to break test cases of integrators.

modules/opcua_client_module/src/opcua_client_module_impl.cpp
```diff
@@ -203,7 +189,7 @@ StringPtr OpcUaClientModule::formConnectionString(const StringPtr& connectionStr
     std::smatch match;
 
     std::string prefix = "";
-    std::string path = "/";
+    std::string path = "";
```
JakaMohorkoDS, line 181: Same as above

modules/opcua_client_module/src/opcua_client_module_impl.cpp
```diff
@@ -287,13 +267,18 @@ Bool OpcUaClientModule::onCompleteServerCapability(const ServerCapabilityPtr& so
     }
 
     const auto path = target.hasProperty("Path") ? target.getPropertyValue("Path") : "";
+    const auto targerAddress = target.getAddresses();
```
JakaMohorkoDS, line 259: Typo

modules/tests/test_opendaq_device_modules/test_native_device_modules.cpp
```diff
@@ -863,26 +862,29 @@ TEST_F(NativeDeviceModulesTest, TestDiscoveryReachabilityAfterConnectIPv6)
...
+                    continue;
 
-            // ASSERT_EQ(ipv4Info.getConnectionString(), capability.getConnectionStrings()[0]);
-            ASSERT_EQ(ipv6Info.getConnectionString(), capability.getConnectionStrings()[0]);
-            
-            // ASSERT_EQ(ipv4Info.getAddress(), capability.getAddresses()[0]);
-            ASSERT_EQ(ipv6Info.getAddress(), capability.getAddresses()[0]);
-            return;
+                if (info.getConnectionString() != deviceConnectionString)
+                    continue;
```
JakaMohorkoDS, line 989: We can probably now have more than 1 address with a IPv4/6 type (eg. 127.0.0.1 + the address obtained via discovery), so this test might fail, if the discovery address is 1st.

modules/tests/test_opendaq_device_modules/test_opcua_device_modules.cpp
```diff
@@ -235,16 +235,19 @@ TEST_F(OpcuaDeviceModulesTest, TestDiscoveryReachabilityAfterConnectIPv6)
...
     auto path = "/test/opcua/discoveryReachabilityAfterConnectIPv6/";
     serverConfig.setPropertyValue("Path", path);
 
     instance.addServer("OpenDAQOPCUA", serverConfig).enableDiscovery();
 
     auto client = Instance();
-    // client.getAvailableDevices();
-    DevicePtr device = client.addDevice(std::string("daq.opcua://[::1]") + path);
+    StringPtr deviceConnectionString = std::string("daq.opcua://[::1]:4840") + path;
+    DevicePtr device = client.addDevice(deviceConnectionString);
```
JakaMohorkoDS, line 250: Is the one without a port no longer valid?

modules/tests/test_opendaq_device_modules/test_opcua_device_modules.cpp
```diff
@@ -257,23 +260,26 @@ TEST_F(OpcuaDeviceModulesTest, TestDiscoveryReachabilityAfterConnectIPv6)
...
+                continue;
 
-        // ASSERT_EQ(ipv4Info.getConnectionString(), capability.getConnectionStrings()[0]);
-        ASSERT_EQ(ipv6Info.getConnectionString(), capability.getConnectionStrings()[0]);
-        
-        // ASSERT_EQ(ipv4Info.getAddress(), capability.getAddresses()[0]);
-        ASSERT_EQ(ipv6Info.getAddress(), capability.getAddresses()[0]);
-        return;
+            if (info.getConnectionString() != deviceConnectionString)
+                continue;
```
JakaMohorkoDS, line 273: Same as above, might fail in case of more than 1 IPv6 address.

modules/websocket_streaming_client_module/src/websocket_streaming_client_module_impl.cpp
```diff
@@ -190,11 +187,14 @@ Bool WebsocketStreamingClientModule::onCompleteServerCapability(const ServerCapa
     }
 
     const auto path = target.hasProperty("Path") ? target.getPropertyValue("Path") : "";
+    const auto targerAddress = target.getAddresses();
```
JakaMohorkoDS, line 189: Typo

modules/tests/test_opendaq_device_modules/test_native_device_modules.cpp
denise-opendaq: it will have the connection string that we are actually using

denise-opendaq (approved):

PR #768, opened 2025-04-14
Creating error info for error codes

denise-opendaq:
# Brief

Improving error info system in openDAQ 

# Description

- The `DAQ_MAKE_ERROR_INFO` macro can be used without a comment message; in that case, it will use the default exception message.
- All places in openDAQ that previously returned error codes now create error info instead.
- If a method receives an error code from another method, it also creates error info. This allows the user to see the full stack trace when checking the error info.
- Two new macros have been introduced:
    - `OPENDAQ_RETURN_IF_FAILED` – creates error info if the returned error code is not a success.
    - `OPENDAQ_RETURN_IF_FAILED_EXCEPT` - if the error is excepted - cleans error info. otherwise creates a new one and returns

# Usage example

In C++ use:

```cpp
ErrCode updateValue(const StringPtr& key, const StringPtr& value)
{
    if (!key.assigned())
        return DAQ_MAKE_ERROR_INFO(OPENDAQ_ERR_ARGUMENT_NULL);

    auto it = map.find(key);
    if (it == map.end())
    {
        // create error info with default message
        return DAQ_MAKE_ERROR_INFO(OPENDAQ_ERR_NOTFOUND);
    }
    it->second = value;
    return OPENDAQ_SUCCESS;
}

ErrCode createOrUpdateValue(const StringPtr& key, const StringPtr& value)
{
    const ErrCode errCode = updateValue(key, value);
    // return with error info if update values has failed not because key is not existing, but because its not assigned
    OPENDAQ_RETURN_IF_FAILED_EXCEPT(errCode, OPENDAQ_ERR_NOTFOUND);
    if (errCode == OPENDAQ_ERR_NOTFOUND)
    {
        map.emplace(key, value);
    }

    return OPENDAQ_SUCCESS;
}
```

# API changes

> [!NOTE]
> method `makeErrorInfo` has a different order of arguments!
```diff
-m [function] ErrCode makeErrorInfo(ErrCode errCode, const std::string& message, IBaseObject* source, Params... params)
+m [function] ErrCode makeErrorInfo(ErrCode errCode, IBaseObject* source, const std::string& message, Params... params)

```


core/coreobjects/include/coreobjects/property_impl.h
```diff
@@ -1142,138 +1137,111 @@ class PropertyImpl : public ImplementationOf<IProperty, ISerializable, IProperty
...
-        {
-            return errCode;
-        }
-        if (errCode != OPENDAQ_ERR_NOTFOUND)
-        {
+        if (errCode == OPENDAQ_ERR_NOTFOUND)
+            daqClearErrorInfo();
+        else if (OPENDAQ_FAILED(errCode))
+            return DAQ_MAKE_ERROR_INFO(errCode);
+        else
```
NikolaiShipilov, line 1158: Would it possible to introduce additional macro to shrink repetitive snippets such:
```
        if (errCode == OPENDAQ_ERR_X)
            daqClearErrorInfo();
        else if (OPENDAQ_FAILED(errCode))
            return DAQ_MAKE_ERROR_INFO(errCode);
        else
```
into
```
        OPENDAQ_RETURN_IF_FAILED_EXCEPT(errCode, OPENDAQ_ERR_X);
```

PR #772, opened 2025-04-17
Other/recorder native rpc

NikolaiShipilov:
# Brief

Adds native config protocol RPCs for IRecorder

### DIFF
--- a/shared/libraries/config_protocol/src/config_protocol_client.cpp
+++ b/shared/libraries/config_protocol/src/config_protocol_client.cpp
@@ -319,6 +319,21 @@ StringPtr ConfigProtocolClientComm::getOperationMode(const std::string& globalId
     return sendComponentCommand(globalId, ClientCommand("GetOperationMode", 9));
 }
 
+void ConfigProtocolClientComm::startRecording(const std::string& globalId)
+{
+    sendComponentCommand(globalId, ClientCommand("StartRecording", 12));
+}
+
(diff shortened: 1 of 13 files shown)


shared/libraries/config_protocol/src/config_protocol_client.cpp
```diff
@@ -319,6 +319,21 @@ StringPtr ConfigProtocolClientComm::getOperationMode(const std::string& globalId
...
+
+void ConfigProtocolClientComm::stopRecording(const std::string& globalId)
+{
+    sendComponentCommand(globalId, ClientCommand("StopRecording", 12));
+}
+
+BooleanPtr ConfigProtocolClientComm::getIsRecording(const std::string& globalId)
+{
+    return sendComponentCommand(globalId, ClientCommand("GetIsRecording", 12));
+}
```
JakaMohorkoDS, line 335: I think this could be implemented as a core event instead of RPC that retrieves the state.

JakaMohorkoDS (approved):

JakaMohorkoDS:
LGTM with the exception of the getter being replaced with an RPC

shared/libraries/config_protocol/src/config_protocol_client.cpp
NikolaiShipilov: This is supposed to be a new core event type, right?
The issue with this approach is that every new recorder implementation would need to trigger this core event on its own, since there's currently no base implementation of Recorder that could do it.

PR #789, opened 2025-05-07
Optimize Signal Packet Handling by Deferring Last Value Calculation

denise-opendaq:
# Brief

Modified signal implementation to defer calculation of the last value when sending packets. Instead of computing the last value immediately, raw data is now stored and the last value is only computed on-demand via `ISignal::getLastValue` method. 

# Description

Optimized the `Signal` implementation by deferring the calculation of the last value until it is explicitly requested via `getLastValue()`. The `Signal` now allocates memory to store the raw value and descriptor. This change separates the logic for packet transmission and value calculation, significantly reducing the time spent when sending data.

Performance Improvements (10 million packets, 100 samples each):

- Raw data: Reduced from **12.1s** to **10.6s**
- Scaled data: Reduced from **18.2s** to **10.4s**
- Linear rule data: Reduced from **23.9s** to **14.8s**

This optimization improves throughput by avoiding unnecessary computation during packet transmission, especially beneficial for high-frequency or high-volume data acquisition scenarios.

# API changes

```diff
+ [function] IDataPacket::getRawLastValue(void** value) = 0;
+ [function] IDataPacket::getRawValueByIndex(void** value, SizeT index) = 0;

+ [function] IDataRuleCalcPrivate::calculateLastSample(const NumberPtr& packetOffset, SizeT sampleCount, void* input, SizeT inputSize, void** output)
```

None

# Required module changes

None


core/opendaq/signal/include/opendaq/data_rule_calc.h
```diff
@@ -358,16 +386,61 @@ inline void DataRuleCalcTyped<T>::calculateConstantSample(const SizeT sampleInde
...
+
+        if (sampleCount - 1 > upToSamples)
+        {
+            *outputTyped = value;
+            return;
+        }
+    }
+
+    // If all upToSamples are >= sampleCount - 1, use the initial constant
+    *outputTyped = initialConstant;
```
JakaMohorkoDS, line 441: I think all of the constant packet entries should be ordered, so just taking the last one should suffice.

core/opendaq/signal/include/opendaq/data_packet_impl.h
```diff
@@ -431,227 +609,73 @@ bool DataPacketImpl<TInterface>::isDataEqual(const DataPacketPtr& dataPacket) co
...
 {
-    const auto dimensions = descriptor.getDimensions();
-
-    if (!dimensions.assigned())
-        DAQ_THROW_EXCEPTION(NotAssignedException, "Dimensions of data descriptor not assigned.");
+    OPENDAQ_PARAM_NOT_NULL(value);
+    OPENDAQ_PARAM_NOT_NULL(*value);
 
-    const auto dimensionCount = dimensions.getCount();
+    const auto dimensionCount = descriptor.getDimensions().getCount();
```
dejancrn, line 624: This requires unnecessary atomic inc and dec. On ARM they are quite expensive, so we should try to avoid them if possible. I would ignore dimension check here and move it to getlastvalue. sampleSize and rawSampleSize already include dimension count

dejancrn (approved):

PR #790, opened 2025-05-07
Fix input port to signal notification

NikolaiShipilov:
# Brief

Adds missing input port notification sent to previously connected signal when it is replaced by new connected signal.

# Description

- Stabilizes client-to-device streaming tests by fixing issues with input port to signal notifications and test environment

### DIFF
--- a/core/coreobjects/include/coreobjects/property_impl.h
+++ b/core/coreobjects/include/coreobjects/property_impl.h
@@ -69,12 +69,6 @@ namespace details
     }
 }
 
-namespace permissions
-{
-    static const auto DefaultPermissions =
-        PermissionsBuilder().inherit(false).assign("everyone", PermissionMaskBuilder().read().write().execute()).build();
-}
-
 class PropertyImpl : public ImplementationOf<IProperty, ISerializable, IPropertyInternal, IOwnable>
 {
 protected:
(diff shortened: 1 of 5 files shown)


core/coreobjects/include/coreobjects/property_impl.h
```diff
@@ -314,7 +308,9 @@ class PropertyImpl : public ImplementationOf<IProperty, ISerializable, IProperty
     {
 #ifdef OPENDAQ_ENABLE_ACCESS_CONTROL
         defaultPermissionManager = PermissionManager();
-        defaultPermissionManager.setPermissions(permissions::DefaultPermissions);
+        const auto DefaultPermissions =
+            PermissionsBuilder().inherit(false).assign("everyone", PermissionMaskBuilder().read().write().execute()).build();
```
JakaMohorkoDS, line 312: This includes a large amount of heap allocations - it'd be good if we could keep this as a static variable.

JakaMohorkoDS (approved):

dejancrn (approved):

core/coreobjects/include/coreobjects/property_impl.h
NikolaiShipilov: I would then just leave this particular change out of the current PR.
The static variable in triggers false positives in memory leak detection — when built with `gcc (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0`.
This doesn’t occur in our CI environment, and interestingly when using Clang on my local machine, the issue also results to such false positives but affects far fewer tests.

PR #792, opened 2025-05-08
Component property search

NikolaiShipilov:
# Brief

Enables property search within a property object, as well as recursive search through a component and its child components.

# Description

- Adds property search filters based on visibility, read-only status and name. Also supports aggregation filters including conjunction, disjunction, negation, a recursive wrapper, and support for user-defined custom filters.
- Relocates the `IRecursiveSearch` interface definition to `core/coretypes/include/coretypes/recursive_search.h` and `ISearchFilter` to `core/coretypes/include/coretypes/search_filter.h`

# Usage example

To create property search filter:

```cpp
#include <coreobjects/search_filter_factory.h>
// ...

using namespace daq::search::properties;
// `visibleWithWriteAccess` filter accepts only visible properties with write access
SearchFilterPtr visibleWithWriteAccess = And(Visible(), Not(ReadOnly()));

// `nonObjectWithName12345` filter accepts only properties with name "12345" except object-properties
SearchFilterPtr nonObjectWithName12345 = And(Name("12345"), Not(Type(CoreType::ctObject)));

// `finalFilter` filter accepts only visible properties with name "12345" and write access
// except object-properties, enabling recursive search in child property objects 
SearchFilterPtr finalFilter = Recursive(And(visibleWithWriteAccess, nonObjectWithName12345));
```

To find and change specific property(-ies):

```cpp
using namespace daq;

// Assumes `propertyObject` is a previously defined instance of type `PropertyObjectPtr`
// and `device` is a previously defined instance of type `DevicePtr`

// filter by property name accepting all properties which names start with "Frequency" word
SearchFilterPtr propertyFilter = search::properties::Name("Frequency.*");

// find properties within the property object
ListPtr<IProperty> properties = propertyObject.findProperties(propertyFilter);

// find properties within the device and all its child components recursively
ListPtr<IProperty> componentProperties = device.findProperties(propertyFilter, search::Recursive(search::Any()));

// change properties values
properties[0].setValue(newValue);
componentProperties[0].setValue(otherValue);
```

# API changes

> [!NOTE]
> breaks API compatibility

```diff
- [function]  ISearchFilter::acceptsComponent(IComponent* component, Bool* accepts)
+ [function]  ISearchFilter::acceptsObject(IBaseObject* obj, Bool* accepts)

+ [factory] SearchFilterPtr Visible()
+ [factory] SearchFilterPtr ReadOnly()
+ [factory] SearchFilterPtr Name(const StringPtr& regex)

+ [function] IPropertyObject::findProperties(IList** properties, ISearchFilter* propertyFilter, ISearchFilter* componentFilter = nullptr)
```

# Required module changes

Breaks the `ISearchFilter` API by changing the input parameters of its methods and renaming the accept method.

As a result, the following implementation example must be updated accordingly:

```diff
class MySearchFilterImpl final : public ImplementationOf<ISearchFilter>
{
public:
    explicit MySearchFilterImpl();

-   ErrCode INTERFACE_FUNC acceptsComponent(IComponent* component, Bool* accepts) override;
+   ErrCode INTERFACE_FUNC acceptsObject(IBaseObject* obj, Bool* accepts) override;

-   ErrCode INTERFACE_FUNC visitChildren(IComponent* component, Bool* visit) override;
+   ErrCode INTERFACE_FUNC visitChildren(IBaseObject* obj, Bool* visit) override;
};
```

Breaking changes to the `ISearchFilter` API also necessitate modifications to the input parameter type of the accept and visit functions within custom search filters:

```diff
FunctionPtr customAcceptorFunc = Function(
-   [](const ComponentPtr& component)
+   [](const BaseObjectPtr& obj)
    {
+       auto component = obj.asPtr<IComponent>();
        return isComponentAccepted(component);
    }
);
FunctionPtr customVisitorFunc = Function(
-   [](const ComponentPtr& component)
+   [](const BaseObjectPtr& obj)
    {
+       auto component = obj.asPtr<IComponent>();
        return isComponentVisited(component);
    }
);

auto foundComponents = folder.getItems(search::Recursive(search::Custom(customAcceptorFunc, customVisitorFunc)));
```


core/coreobjects/include/coreobjects/property_filter_factory.h
```diff
@@ -0,0 +1,65 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+#include <coretypes/search_filter_ptr.h>
+#include <coreobjects/property_filter.h>
+#include <coretypes/string_ptr.h>
```
JakaMohorkoDS, line 20: I would do cascading includes. This should include `coretypes/search_filter_factory.h`. Same for the openDAQ - it should include `coreobjects/search_filter_factory.h`

bindings/python/core_types/CMakeLists.txt
```diff
@@ -45,6 +45,7 @@ set(SRC_Cpp
     generated/py_enumeration_type.cpp
     generated/py_enumeration.cpp
     generated/py_version_info.cpp
+    generated/py_search_filter.cpp
```
JakaMohorkoDS, line 48: This probably must be manually generated now? It seems that we're missing the coreobjects and opendaq factories.

core/coreobjects/include/coreobjects/property_filter_factory.h
```diff
@@ -0,0 +1,65 @@
...
+/*!
+ * @ingroup objects_property_filter
+ * @addtogroup objects_property_filter_factories Factories
+ * @{
+ */
+
+/*!
+ * @brief Creates a search filter that accepts only visible properties.
+ */
+inline SearchFilterPtr Visible()
```
JakaMohorkoDS, line 36: Not sure whether we really need to differentiate between the property one and the component one. Wouldn't it be possible to simply check for the object type in 1 implementation?

core/coreobjects/include/coreobjects/property_object.h
```diff
@@ -375,6 +377,17 @@ DECLARE_OPENDAQ_INTERFACE(IPropertyObject, IBaseObject)
...
     virtual ErrCode INTERFACE_FUNC getPermissionManager(IPermissionManager** permissionManager) = 0;
+
+    // [elementType(properties, IProperty)]
+    /*!
+     * @brief Returns a list of properties contained in the Property object that are accepted by the specified property filter.
+     * @param filter An optional filter used to exclude unwanted properties and optionally enable recursive search.
+     * @param[out] properties The list of properties found.
+     *
+     * If no filter is provided, only visible properties directly belonging to the current object are returned.
+     * Recursion enabled through corresponding wrapper filter includes properties from child property objects into search results.
```
JakaMohorkoDS, line 388: The doxygen comment hasn't been updated. Should contain info on the difference between the property and component filter.

JakaMohorkoDS:
Another small comment - before merge, the API changes and PR content should be updated to match the last changes. We should indicate that this could be a breaking change.

core/coreobjects/include/coreobjects/property_filter_factory.h
NikolaiShipilov: I don't think we can achieve this without introducing more significant changes.

One potential option could be to introduce a new `IVisible:: public IBaseObject` interface with a single method `getVisible(Bool* visible)`, and then have both component (technically should be `IPropertyObject:: public IVisible`) and property interfaces inherit from it.

With that in place, we could have a common implementation `VisibleSearchFilterImpl` in `coretypes` lib, where we’d cast the object passed to `ISearchFilter` methods to `VisiblePtr` and call `*outParam = visiblePtr.getVisible()` to get the result.

However, such solution feels overly complex and not good overall.

Alternatively, we could introduce a base class, `VisibleSearchFilterBaseImpl` in coretypes , to cover the basic common filter logic, with dedicated final classes for property and component searches that would do type checking and visibility extraction. This also implies making public an additional impl.h header from the `coretypes` lib.
I'm not sure this option is justified either.

NikolaiShipilov: The third option might be `IVisible::public IBaseObject { getVisible(Bool* visible) }` being just an addition for `getVisible` methods of `IProperty` and `IComponent` instead of being replacement. This would keep existing interface inheritance hierarchy, with just `IVisible` being additional interface for component and property.

bindings/python/core_types/CMakeLists.txt
NikolaiShipilov: In addition to auto-generated binding in `coretypes` py module there are now extra bindings files also in `coreobject` and `opendaq` py modules used to register type-specific search filter factories.

JakaMohorkoDS (approved):
Lgtm!

PR #807, opened 2025-05-16
Device operation mode switching in Python GUI Demo Application

aljazfrancic:
# Brief

Enable Device operation mode switching in Python GUI Demo Application.

# Description

- Add `operation_mode` getter to Python bindings
- Add logic to handle operation mode switching though Python GUI Demo App

# Usage example

In Python GUI Demo App use:
image


aljazfrancic:
I've tested by using the Virtual Box Simulator from latest main and opening two copies of modified Python GUI Demo App, etc.

Yarikk26 (approved):
lgtm

denise-opendaq (approved):

PR #808, opened 2025-05-19
Fix examples CMake

aljazfrancic:
# Brief

Fix examples CMake, which fixes examples when downloaded as solo archive via https://docs.opendaq.com/ or https://docs-dev.opendaq.com/ (when not part of whole openDAQ project)

# Description

* Set missing CMake variable `OPENDAQ_EXAMPLES_VERSION` that is amiss only if examples are downloaded solo via downloads webpage
* Set CMake variable `MODULE_PATH` in examples

- Rename a function
- ...

# Usage example

https://docs-dev.opendaq.com/manual/opendaq/dev/tutorials/quick_start_setting_up_cpp.html

### DIFF
--- a/examples/cpp/CMakeLists.txt
+++ b/examples/cpp/CMakeLists.txt
@@ -1,7 +1,14 @@
 cmake_minimum_required(VERSION 3.20)
 
-set(OPENDAQ_EXAMPLES_VERSION ${OPENDAQ_PACKAGE_VERSION})
+if(NOT DEFINED OPENDAQ_PACKAGE_VERSION)
+  set(OPENDAQ_EXAMPLES_VERSION "0.0.0")
+elseif ()
+  set(OPENDAQ_EXAMPLES_VERSION ${OPENDAQ_PACKAGE_VERSION})
+endif()
+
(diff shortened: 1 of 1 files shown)


examples/cpp/CMakeLists.txt
```diff
@@ -1,8 +1,15 @@
...
-set(OPENDAQ_EXAMPLES_VERSION ${OPENDAQ_PACKAGE_VERSION})
+if(NOT DEFINED OPENDAQ_PACKAGE_VERSION)
+  set(OPENDAQ_EXAMPLES_VERSION "0.0.0")
+elseif ()
+  set(OPENDAQ_EXAMPLES_VERSION ${OPENDAQ_PACKAGE_VERSION})
+endif()
+
 project(opendaq_examples VERSION ${OPENDAQ_EXAMPLES_VERSION})
 
+add_compile_definitions(MODULE_PATH="${OPENDAQ_MODULES_DIR}")
```
mojca, line 6: I would call this `OPENDAQ_MODULE_PATH` (and change the code that uses this variable).

But also: shouldn't this be `target_compile_definition`?

PR #809, opened 2025-05-19
Fixing freezing the gui demo on calling function property with args. gui demo improvements

denise-opendaq:
…window in focus. add closing the windows on escape

# Brief

# Description

- Rename a function
- ...

# Usage example

In C++ use:

```cpp
auto variable = "foo";
```

In terminal use:

```shell
cd folder
```

# API changes

> [!NOTE]

```diff
+ this
- that
```

- ...

Instead of:

```cpp
foo();
```

Do:

```cpp
bar();
```

# Required module changes

- ...

Instead of:

```cpp
foo();
```

Do:

```cpp
bar();
```

### DIFF
--- a/examples/python/gui_demo/components/dialog.py
+++ b/examples/python/gui_demo/components/dialog.py
@@ -15,6 +15,9 @@ class Dialog(tk.Toplevel):
         self.configure(padx=10, pady=5)
         self.attributes('-topmost', True)
         self.transient(parent)
+        
+        # Bind Escape key to close the dialog
+        self.bind("<Escape>", lambda event: self.close())
 
     def center_window(self):
         main_window = self.parent.winfo_toplevel()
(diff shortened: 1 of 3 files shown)


examples/python/gui_demo/components/dialog.py
```diff
@@ -10,11 +10,13 @@ def __init__(self, parent, title, context: AppContext, **kwargs):
         self.title(title)
         self.parent = parent
         self.context = context
-        self.initial_update_func = None
```
Yarikk26, line 13: Is it a non-breaking change here?

examples/python/gui_demo/components/function_dialog.py
```diff
@@ -100,13 +100,13 @@ def create_argument(self, value_type_num: int, *args):
     def exec_clicked(self):
         ret = None
         try:
-            args = []
+            list = daq.List()
```
Yarikk26, line 103: It looks like it was the problem with the GIL, so one of the possible ways would be removing the lock release in C++ code, but it is ok also.

Yarikk26 (approved):

PR #810, opened 2025-05-20
Load individual modules

NikolaiShipilov:
# Brief

Enables functionality for loading individual modules from a specified absolute filesystem path.

# Description

- Allows manual loading of a specific module after the default set has been loaded.
- The default module directory can be rescanned to detect and load newly added modules without affecting loaded ones.

# Usage Example

In C++:

```cpp
// Create an instance and load modules from the default MODULE_PATH
const InstancePtr instance = Instance(MODULE_PATH);

// Load a specific module from an absolute path
ModulePtr module = instance.getModuleManager().loadModule("/home/users/user1/openDAQ/modules/MyModule.module.so");

// Rescan the default module path to pick up any newly added modules, without affecting the ones already loaded
instance.getModuleManager().loadModules(instance.getContext());
```

# API changes

```diff
+ [function] IModuleManager::loadModule(IString* path, IModule** module)
```


JakaMohorkoDS (approved):
Lgtm. Could we also add this to the Python application in the "File" dropdown?

NikolaiShipilov:
> Lgtm. Could we also add this to the Python application in the "File" dropdown?

Absolutely. This actually will be a great mechanism to test the feature thoroughly, so I’m going to include it in this PR.

PR #811, opened 2025-05-21
Color coding and message display for Component statuses in Python GUI application

aljazfrancic:
# Brief

Add color coding and message display for Component statuses in Python GUI demo application.

# Description

- Addresses TBBAS-2194
- Add color coding for Component statuses in Python GUI demo application
- Add Component status message display (if there is one) in Python GUI demo application
- Currently auto-refresh is not implemented (clicking the "Refresh" button does work, however)
- Stick operation mode selection to west (unrelated)

# Usage example

OK Component status example:

image

Warning Component status example with message:

image

Error Component status example with message:

image

### DIFF
--- a/examples/python/gui_demo/components/block_view.py
+++ b/examples/python/gui_demo/components/block_view.py
@@ -13,7 +13,7 @@ from .attributes_dialog import AttributesDialog
 class BlockView(ttk.Frame):
 
     def __init__(self, parent, node, context=None, expanded=False, **kwargs):
-        ttk.Frame.__init__(self, parent, **kwargs)
+        ttk.Frame.__init__(self, parent, style="Custom.TFrame", **kwargs)
         self.parent = parent
         self.expanded = expanded
         self.node = node
@@ -133,7 +133,7 @@ class BlockView(ttk.Frame):
                 opt.trace_add("write", on_option_change)
 
                 combined = tk.Frame(self.expanded_frame)
-                combined.grid(row=1, column=0, padx=5, pady=5)
+                combined.grid(row=1, column=0, padx=5, pady=5, sticky="w")
 
                 label = tk.Label(combined, text="Operation mode: ")
                 label.pack(side="left")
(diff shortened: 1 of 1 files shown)


Yarikk26 (approved):

mojca:
Painting the full frame looks somewhat too intrusive to me, but I'm not a designer (and I believe this is a question for a designer rather than for a programmer). I see no reason for painting the frame green when things work fine. For warnings and errors it would make more sense to me to have either a filled rectangle or a filled circle drawn either on top or on the bottom (probably next to the error message).

aljazfrancic:
I changed to GUI to be less intrusive @mojca @Yarikk26 

image

image

image

image

adolenc:
These last pictures look fine, as long as the status text doesn't get pushed "out of view" for any reason (think window height or property count or anything like that), as this status is extremely important in figuring out what is wrong with your component.

Additionally, as requested in the ticket, you should somehow add the component statuses to the entries in the tree view on the left hand side, as you currently have no way to figure out that something is wrong with your entire setup unless you click on every single component.

aljazfrancic:
@adolenc The problem is that only way to add items to a Tkinter `Treeview` is via `insert`, which doesn't allow extra square/circle before text (image is already used for the icons). So without using an additional an library for custom TreeView entries, best I can do is change font color of TreeView items according to their Component status.

The result looks like this:
image

aljazfrancic:
@adolenc Also tested some long status text, and it appears to work:

image

adolenc:
> The result looks like this: image

In my opinion, make the entries with error state red, make the entries with warning state some readable version of orange that is still distinct enough from red, and leave the rest black.

aljazfrancic:
@adolenc I adjusted the color scheme, however we still color-code the TreeView items green, if their Component status is OK, to distinguish from Component status not being set (eg. BasicCsvRecorder). Also we still show the light blue square and "Status not set" for Components with unassigned status, while we match the color scheme for other "status squares" (OK, Warning, Error) from TreeView see:

image

adolenc:
> @adolenc I adjusted the color scheme, however we still color-code the TreeView items green, if their Component status is OK, to distinguish from Component status not being set (eg. BasicCsvRecorder). Also we still show the light blue square and "Status not set" for Components with unassigned status, while we match the color scheme for other "status squares" (OK, Warning, Error) from TreeView see:

For the end user, "not set" is the same as "no error/warning". Errors and warnings are something the user should be notified about because they probably have something misconfigured, and green for OK just adds unnecessary visual clutter.

Yellow on white is impossible to read that is why I am suggesting you go with darker orange.

mojca:
I'm with Andrej here. When status is OK, the text should remain at its default colour. Why do you need to distinguish "status ok" from "no status set" in the tree view?

examples/python/gui_demo.py
```diff
@@ -220,6 +220,9 @@ def tree_widget_create(self, parent_frame):
         scroll_bar.pack(fill=tk.Y, side=tk.RIGHT)
 
         parent_frame.add(frame)
+        tree.tag_configure('ok', foreground='olive drab')
+        tree.tag_configure('warning', foreground='orange')
+        tree.tag_configure('error', foreground='red')
```
mojca, line 225: You should define the colour palette at a single place and refer to those colours by a variable name. You are hardcoding "olive drab" several times in this single MR. Once the code starts growing and you'll want to change the colour, you'll be struggling to find all the zillion of places where some exotic colour has been specified, and there's a high probability that you'll miss some places. Or someone will want to use dark mode, the colour will be hardly readable on dark background, and it will be extremely painful to change it.

JakaMohorkoDS (approved):

PR #815, opened 2025-05-23
Remove dynamic casts in queryInterface and borrowInterface

dejancrn:
# Brief

# Description

- Rename a function
- ...

# Usage example

In C++ use:

```cpp
auto variable = "foo";
```

In terminal use:

```shell
cd folder
```

# API changes

> [!NOTE]

```diff
+ this
- that
```

- ...

Instead of:

```cpp
foo();
```

Do:

```cpp
bar();
```

# Required module changes

- ...

Instead of:

```cpp
foo();
```

Do:

```cpp
bar();
```


core/coretypes/include/coretypes/intfs.h
```diff
@@ -155,23 +157,33 @@ struct SupportsInterface
 {
     using Interface = typename TArgs::Head;
 
-    static bool Check(const IntfID& id, void** intf, void* object, bool addRef)
+    template <class TMainInterface, bool AddRef, class This>
```
Unsttopabull, line 160: Why move `AddRef` to template parameter? This will somewhat unnecessarily increase template instantiations significantly (and as such compile-time and compiled size)

Unsttopabull (approved):

core/coretypes/include/coretypes/intfs.h
dejancrn: That's a fair point. However, I measured compile time, and there's no significant difference—at least not in the MSVC compiler.

In addition, the template parameter removes unnecessary branching instructions at run time, which can pollute the branch predictor.

JakaMohorkoDS (approved):

PR #819, opened 2025-05-27
Ensure property order is preserved from server to client via the native protocol.

denise-opendaq:
# Brief

# Description

- Rename a function
- ...

# Usage example

In C++ use:

```cpp
auto variable = "foo";
```

In terminal use:

```shell
cd folder
```

# API changes

> [!NOTE]

```diff
+ this
- that
```

- ...

Instead of:

```cpp
foo();
```

Do:

```cpp
bar();
```

# Required module changes

- ...

Instead of:

```cpp
foo();
```

Do:

```cpp
bar();
```

### DIFF
--- a/core/coreobjects/include/coreobjects/core_event_args.h
+++ b/core/coreobjects/include/coreobjects/core_event_args.h
@@ -392,6 +392,21 @@ OPENDAQ_DECLARE_CLASS_FACTORY_WITH_INTERFACE(
     LIBRARY_FACTORY, CoreEventArgsTypeRemoved, ICoreEventArgs,
     IString*, typeName
 )
+
+/*!
+ * @brief Creates Core event args that property order of a component is changed.
+ * @param propOwner The property object that owns the properties whose order was changed.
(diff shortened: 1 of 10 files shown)


core/coreobjects/include/coreobjects/core_event_args.h
```diff
@@ -392,6 +392,21 @@ OPENDAQ_DECLARE_CLASS_FACTORY_WITH_INTERFACE(
     LIBRARY_FACTORY, CoreEventArgsTypeRemoved, ICoreEventArgs,
     IString*, typeName
 )
+
+/*!
+ * @brief Creates Core event args that property order of a component is changed.
+ * @param propOwner The property object that owns the properties whose order was changed.
+ * @param propertyOrder The list of property names in the new order.
```
denise-opendaq, line 399: оо

JakaMohorkoDS (approved):
Lgtm, unsure however, if this change should also increase the protocol version.

PR #826, opened 2025-05-29
Add GIL-safe queued Python event handling to prevent deadlocks in pybind11 callbacks

denise-opendaq:
# Brief

Introduce a thread-safe callback queue and deferred dispatch mechanism to safely execute Python event handlers from C++ without risking GIL-related deadlocks.

# Description

This PR introduces a GIL-safe, asynchronous event dispatch system for openDAQ's Python bindings. It replaces direct Python callback invocations from C++ threads with a queue-based dispatch model, ensuring thread safety and eliminating potential GIL-related issues.

- Added `opendaq.event_queue`, a globally accessible Python object exposing a thread-safe event queue from C++.

  - event_queue exposes the following methods:
  
  - `enqueue_event(handler, sender, args)` – defers the event callback.
  
  - `process_events()` – processes and dispatches all queued events.
  
  - `clear(`) – clears any pending events without dispatching.

- Introduced `opendaq.QueuedEventHandler` to safely register Python event handlers that are deferred and executed from the main thread.

- Integrated periodic polling of `event_queue.process_events()` into the GUI demo to ensure timely event dispatch.

- Automatic cleanup of unprocessed events is handled on module unload via a C++ RAII object and weak pointer tracking.

- This design ensures that Python callbacks are never invoked directly from C++ threads, maintaining safety and stability across multi-threaded scenarios.

# Usage example

In Python use:

```py
import opendaq as daq
import time
from typing import Optional

def on_core_event(sender: Optional[daq.IComponent], args: daq.IEventArgs):
    print(args.event_name)

instance = daq.Instance()
instance.context.on_core_event + daq.QueuedEventHandler(on_core_event)

try:
    print("Polling openDAQ event queue (press Ctrl+C to stop)...")
    while True:
        daq.event_queue.process_events()
        time.sleep(0.05)
except KeyboardInterrupt:
    print("Exiting.")
```


Yarikk26 (approved):
lgtm. It would be good also to change test cases to eliminate the old bug prone event handling.

Yarikk26 (approved):

PR #833, opened 2025-06-04
Fix Invalid MultiReader in PowerReaderFb on Sample Rate Change

denise-opendaq:
# Brief

Fix Invalid MultiReader in PowerReaderFb on Sample Rate Change

# Description

Fixes an issue in `PowerReaderFb` where the `MultiReader` becomes invalid due to signal sample rate changes. A new `MultiReader` is created when the existing one is invalid to ensure consistent processing:


adolenc:
Hey @denise-opendaq, is "-" the final commit I should be trying out, or are you still working on this?

denise-opendaq (commented):
@adolenc I have not tested it properly, so its up to you

adolenc:
This is looking very good to me! Do I need to change the acquisition lock to recursive config lock or was that just a leftover?

adolenc:
After playing with it a bit it seems to work fine to me. We will test it more extensively in the next couple of days.
Thanks!

PR #837, opened 2025-06-11
Implement main worker in the scheduler

denise-opendaq:
# Brief

# Description

- Rename a function
- ...

# Usage example

In C++ use:

```cpp
auto variable = "foo";
```

In terminal use:

```shell
cd folder
```

# API changes

> [!NOTE]

```diff
+ this
- that
```

- ...

Instead of:

```cpp
foo();
```

Do:

```cpp
bar();
```

# Required module changes

- ...

Instead of:

```cpp
foo();
```

Do:

```cpp
bar();
```


core/opendaq/scheduler/src/scheduler_impl.cpp
```diff
@@ -28,6 +28,100 @@ class CustomWorkerInterface : public tf::WorkerInterface
...
+    {
+        cv.wait(lock, [this] { return !workQueue.empty() || !running; });
+        if (!running)
+            return;
+
+        lock.unlock();
+        runIteration();
+        lock.lock();
+    }
+}
```
JakaMohorkoDS, line 104: This should not spin freely. We should have an adjustable loop cycle and wait for the remaining duration once the work queue is emptied.

core/opendaq/scheduler/src/scheduler_impl.cpp
```diff
@@ -142,6 +240,30 @@ std::size_t SchedulerImpl::getWorkerCount() const
...
+{    
+    mainThreadWorker->stop();
+    return OPENDAQ_SUCCESS;
+}
+
+ErrCode SchedulerImpl::runMainLoopIteration()
+{
+    mainThreadWorker->runIteration();
+    return OPENDAQ_SUCCESS;
+}
```
JakaMohorkoDS, line 259: Does this make sense when the loop is running?

core/opendaq/scheduler/include/opendaq/scheduler.h
```diff
@@ -85,6 +85,44 @@ DECLARE_OPENDAQ_INTERFACE(IScheduler, IBaseObject)
...
+     * @brief Schedules a task to be executed by the main loop.
+     *
+     * The provided work object is queued for execution during a call to either
+     * @ref runMainLoop or @ref runMainLoopIteration. This mechanism is commonly used
+     * to marshal tasks from background threads to the main loop thread.
+     *
+     * @param work A lightweight, non-blocking task object to be scheduled.
+     */
+    virtual ErrCode INTERFACE_FUNC scheduleWorkOnMainLoop(IWork* work) = 0;
+
```
JakaMohorkoDS, line 125: An additional function would be great if we also add a loop time:

`scheduleWorkOnNextMainLoop` or something similarly named. It should post a task that's put at the tail of the work queue that'll be run on the next loop cycle. This would allow a function block/device to be polled once every loop cycle.

core/opendaq/scheduler/include/opendaq/work_repetitive.h
```diff
@@ -0,0 +1,45 @@
...
+/*!
+ * @brief The interface inherited from IWork. And used for repetitive work that needs to be executed in the main loop.
+ * The callback is expected to return a boolean value indicating whether the work should be repeated.
+ * If the callback does return nothing, it is assumed that the work should not be repeated.
+ *
+ * @note if the callback should not be repeated, the method execute() should return OPENDAQ_ERR_REPETITIVE_TASK_STOPPED.
+ */
+DECLARE_OPENDAQ_INTERFACE(IWorkRepetitive, IWork)
+{
+};
```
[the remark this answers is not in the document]
JakaMohorkoDS: This one is not yet addressed.

JakaMohorkoDS:
> Looking at the new changes, the loop time looks good. I have one thought about how function blocks might use this - basically, some function blocks might rely upon the existence of the main thread, but not all applications will have it available.
> 
> Would it make sense to still start a new thread in the scheduler regardless of if the main loop is called? The main loop duration would in this case need to be set up in the context (scheduler constructor), but would ensure that a thread is always available for function blocks.
> 
> Invoking the main loop method would simply join the previous loop and transfer ownership of the work queue to the main thread.

This can be kept for a different PR. We discussed adding an IScheduler method `IScheduler::usesMainLoop` or something similarly named.

PR #845, opened 2025-06-20
implemented and integrated a WAV file reader

JakobVokac:
# Brief

Added a WAV reader function block.


JakaMohorkoDS:
OpcuaDeviceModulesTest.DeviceDynamicFeatures Should have the expected function block count adjusted to 12

modules/audio_device_module/src/wav_reader_fb_impl.cpp
```diff
@@ -0,0 +1,285 @@
...
+            if (read)
+            {
+                if (decoderReading())
+                {
+                    return;
+                }
+                if (!decoderReady())
+                {
+                    args.setValue(false);
+                    setComponentStatusWithMessage(ComponentStatus::Error, "Cannot set Reading to true, no valid file!");
```
JakaMohorkoDS, line 273: This checks that the device state is "stopped", not necessarily that we do not have a valid file.

The ComponentStatus is a persistent state, not a 1-time event. As such, we can sadly not really provide logging information on why the setter has failed. If we want to convey this information, the function block should always be in a "Warning" state when the file path is not yet configured and set to "Ok" once that is done.

modules/audio_device_module/src/wav_reader_fb_impl.cpp
```diff
@@ -0,0 +1,285 @@
...
+void WAVReaderFbImpl::initProperties()
+{
+    objPtr.addProperty(StringProperty("FilePath", ""));
+    objPtr.getOnPropertyValueWrite("FilePath") += [this](PropertyObjectPtr& /*obj*/, PropertyValueEventArgsPtr& args)
+        {
+            objPtr.setPropertyValue("Reading", false);
+            if (!updateFilePath(static_cast<std::string>(args.getValue())))
+            {
+                args.setValue("");
+                setComponentStatusWithMessage(ComponentStatus::Error, "Invalid file path!");
```
JakaMohorkoDS, line 256: This overrides any status message set up by the `initalizeDecoder` method if it failed.

modules/audio_device_module/src/wav_reader_fb_impl.cpp
```diff
@@ -0,0 +1,285 @@
...
+    return true;
+}
+
+void WAVReaderFbImpl::initProperties()
+{
+    objPtr.addProperty(StringProperty("FilePath", ""));
+    objPtr.getOnPropertyValueWrite("FilePath") += [this](PropertyObjectPtr& /*obj*/, PropertyValueEventArgsPtr& args)
+        {
+            objPtr.setPropertyValue("Reading", false);
+            if (!updateFilePath(static_cast<std::string>(args.getValue())))
```
JakaMohorkoDS, line 253: `updateFilePath` is overloaded here.
The call stack is:
```
- `updateFilePath`
  - `initializeDecoder`
    - `initializeSignal`
```
And all those calls are sequential, making them hard to follow in code.

Having something similar to the following would make it clearer to the user what's going on. 

```cpp
if (!initializeDecorder(path))
{
  // Handle error
}

if (!initializeSignal())
{
  // Handle error and uninitialize decoder
}
```

modules/audio_device_module/src/wav_reader_fb_impl.cpp
```diff
@@ -0,0 +1,285 @@
...
+    {
+        setComponentStatusWithMessage(ComponentStatus::Error, "Failed to stop playback device.\n");
+        ma_device_uninit(&device);
+        ma_decoder_uninit(&decoder);
+    }
+}
+
+bool WAVReaderFbImpl::updateFilePath(const std::string& newPath)
+{
+    if (newPath == filePath || newPath == "")
```
JakaMohorkoDS, line 167: The equality check is handled by the PropertyObject internal code where we do not trigger the `onWrite` callback if the same value is written twice.

modules/audio_device_module/src/wav_reader_fb_impl.cpp
```diff
@@ -0,0 +1,285 @@
...
+        return false;
+    }
+}
+
+void WAVReaderFbImpl::setRead(bool read)
+{
+    if (reading == read)
+    {
+        return;
+    }
```
JakaMohorkoDS, line 190: Same as `FilePath`

modules/audio_device_module/src/wav_writer_fb_impl.cpp
```diff
@@ -94,101 +144,52 @@ void WAVWriterFbImpl::initProperties()
...
-    std::vector<float> inputData;
+    std::vector<short> inputData;
     inputData.reserve(std::max(availableData, static_cast<SizeT>(1)));
 
     const auto status = reader.read(inputData.data(), &availableData);
 
-    if (storing)
+    if (status.getReadStatus() == ReadStatus::Event && status.getEventPacket().getEventId() == event_packet_id::DATA_DESCRIPTOR_CHANGED)
+    {
+        stopRecording();
```
JakaMohorkoDS, line 187: I would stop recording on any event as of now - the two possible events are "Descriptor changed" and "Gap detected". I don't think we can currently handle any of them.

modules/audio_device_module/src/wav_writer_fb_impl.cpp
```diff
@@ -28,14 +27,65 @@ FunctionBlockTypePtr WAVWriterFbImpl::CreateType()
     );
 }
 
+ErrCode WAVWriterFbImpl::startRecording()
```
JakaMohorkoDS, line 30: The start/stop/getIsRecording should be under a lock.

JakaMohorkoDS (changes requested):
Lgtm! A couple minor comments here and there, but otherwise a nice addition to the module suite.

modules/audio_device_module/src/wav_reader_fb_impl.cpp
```diff
@@ -0,0 +1,284 @@
...
+{
+    objPtr.addProperty(StringProperty("FilePath", ""));
+    objPtr.getOnPropertyValueWrite("FilePath") += [this](PropertyObjectPtr& /*obj*/, PropertyValueEventArgsPtr& args)
+        {
+            objPtr.setPropertyValue("Reading", false);
+            std::string path = static_cast<std::string>(args.getValue());
+
+            if (!updateFilePath(path))
+            {
+                setComponentStatusWithMessage(ComponentStatus::Warning, "Invalid file path!");
```
JakaMohorkoDS, line 246: We should probably return early in this case, so we provide the correct status message.

modules/audio_device_module/src/wav_reader_fb_impl.cpp
```diff
@@ -0,0 +1,284 @@
...
+            }
+            if (!reinitializeDecoder())
+            {
+                args.setValue("");
```
JakaMohorkoDS, line 250: I'm not quite sure I'd override the written value - I'd leave it as is, but give a descriptive status indicating what's wrong with the path. Whether it's an invalid file path, invalid file type, corrupted file...

And again, we should likely exit early in case an error is encountered and not do further initialization.

modules/audio_device_module/src/wav_reader_fb_impl.cpp
```diff
@@ -0,0 +1,284 @@
...
+    {
+        setComponentStatusWithMessage(ComponentStatus::Warning, "Failed to initialize miniaudio playback device.");
+        ma_decoder_uninit(&decoder);
+        return false;
+    }
+
+    samplesCaptured = 0;
+    framesAvailable = true;
+    objPtr.asPtr<IPropertyObjectProtected>().setProtectedPropertyValue("EOF", false);
+    setComponentStatusWithMessage(ComponentStatus::Ok, "File ready.");
```
JakaMohorkoDS, line 113: I would just change this to "OK" on user input -> so when the user successfully changes the "FilePath" variable. That's the case when `updateFilePath` `reinitializeDecoder` and `initializeSignals` all succeed.

PR #850, opened 2025-07-01
Ignore server notifications unless native device is connected

NikolaiShipilov:
# Brief

The native configuration device implementation keeps notification packet handling disabled until the device connection process is completed.

# Description

Previously, the configuration protocol client attempted to process configuration changes as core event notifications from the server before the connection was fully established, which resulted in a crash.
The applied fix ensures that the client simply ignores any such notifications received prior to connection completion.

However, this solution does not eliminate all potential race conditions. In particular, there's still a risk that a core event triggered while the server is processing a `GetSerializedRootDevice` request - e.g. during the reply encoding after serialization - may be missed by the client, as it might still ignore the corresponding core event notification.

### DIFF
--- a/changelog/changelog_3.10-3.20.md
+++ b/changelog/changelog_3.10-3.20.md
@@ -40,7 +40,8 @@
 ## Bug fixes
 
 - [#850](https://github.com/openDAQ/openDAQ/pull/850) Ignore server notifications unless native device is connected.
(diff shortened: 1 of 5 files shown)


JakaMohorkoDS (approved):
Lgtm. Do we have a similar issue on reconnection? Also, would this need to be under a lock?

denise-opendaq (approved):

PR #854, opened 2025-07-03
Wrapped data packet

dejancrn:
# Brief

Wrapped data packets

# Description

Wrapped data packets are packets that wrap another data packet as a source and use their data buffers. A wrapped data packet can set a different data descriptor from the source. This is useful if a function block processes source packets in a way that leaves raw data untouched, and it only changes post-scaling factors. This avoids unnecessary memory copies of raw data.

Note that the current implementation of native and LT streaming does not recognize wrapped data packets and treats them as standard data packets. This means that if both source and wrapped data packets are transmitted, the underlying shared buffer will be transmitted twice. This represents a potential area for future optimization.

# Usage example

In C++ use:

```cpp
    const auto domainDescriptor = DataDescriptorBuilder()
        .setSampleType(SampleType::Int32)
        .setRule(LinearDataRule(1, 0))
        .setTickResolution(Ratio(1, 1000))
        .build();

    const auto sourceValueDescriptor = DataDescriptorBuilder()
        .setSampleType(SampleType::Float32)
        .setPostScaling(LinearScaling(1.0, 0.0, SampleType::Int32, ScaledSampleType::Float32))
        .build();

    constexpr size_t sampleCount = 4;

    auto sourcePacket = DataPacket(sourceValueDescriptor, sampleCount);
    auto sourceRawData = static_cast<int32_t*>(sourcePacket.getRawData());
    for (size_t i = 0; i < sampleCount; i++)
        *sourceRawData++ = static_cast<int32_t>(i);

    auto sourceData = static_cast<float*>(sourcePacket.getData());
    for (size_t i = 0; i < sampleCount; i++)
        assert(*sourceData++ == static_cast<float>(i));

    const auto wrappedValueDescriptor = DataDescriptorBuilder()
        .setSampleType(SampleType::Float32)
        .setPostScaling(LinearScaling(2.0, 1.0, SampleType::Int32, ScaledSampleType::Float32))
        .build();

    auto wrappedPacket = WrappedDataPacket(sourcePacket, wrappedValueDescriptor);

    auto wrappedData = static_cast<float*>(wrappedPacket.getData());
    for (size_t i = 0; i < sampleCount; i++)
        assert(*wrappedData++ == 2.0f * static_cast<float>(i) + 1.0f);
```

# API changes

```diff
+ [factory] DataPacketPtr WrappedDataPacket(const DataPacketPtr& sourcePacket, const DataDescriptorPtr& descriptor)
```

### DIFF
--- a/core/opendaq/signal/include/opendaq/data_packet_impl.h
+++ b/core/opendaq/signal/include/opendaq/data_packet_impl.h
@@ -212,13 +212,18 @@ inline BaseObjectPtr buildObjectFromDescriptor(void*& addr, const DataDescriptor
     }
     return buildFromDescriptor(rawAddr, descriptor, typeManager);
 }
+
+struct CreatePacketNoMemoryTag
+{
+};
+
 }
 
-template <typename TInterface = IDataPacket>
-class DataPacketImpl : public GenericDataPacketImpl<TInterface, IReusableDataPacket>
+template <typename TInterface = IDataPacket, typename ... TInterfaces>
+class DataPacketImpl : public GenericDataPacketImpl<TInterface, IReusableDataPacket, TInterfaces ...>
 {
 public:
-    using Super = GenericDataPacketImpl<TInterface, IReusableDataPacket>;
+    using Super = GenericDataPacketImpl<TInterface, IReusableDataPacket, TInterfaces ...>;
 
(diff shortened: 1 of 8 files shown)


dejancrn:
@NikolaiShipilov, I have added you as a reviewer to familiarize you with the new feature that may require optimization for streaming in the future.

PR #869, opened 2025-07-25
Implementing video player function block

denise-opendaq:
# Brief

Implementing the function block `RefFBModuleVideoPlayer`, which visualizes video frames and displays the timestamp in the top-left corner.

# Description

The `RefFBModuleVideoPlayer` function block is responsible for visualizing video frames. It uses the main thread worker and throws an exception if the main worker is not set in the scheduler.
The function block supports the following image formats: `jpeg`, `png`, `bmp`, `targa`, and `hdr`. Format validation is performed both when validating a new packet description and when reading the image using the SFML library.

If the format is invalid, a warning is logged, and the visualization window is closed. The window will reopen if a subsequent packet with a valid description is received.

If the window is not open, the function block will set its input port to an inactive state to prevent the internal cache from overflowing with frame packets.

# Usage example

In C++ use:

```cpp
// Setting the instance builder to use the main loop
instanceBuilder.setUsingSchedulerMainLoop(true);

// Creating video player function block and connecting the video signal
auto videoPlayerFb = instance.addFunctionBlock("RefFBModuleVideoPlayer");
videoPlayerFb.getInputPorts()[0].connect(videoSignal);

// Run the scheduler's main loop
auto scheduler = instance.getContext().getScheduler();
scheduler.runMainLoop();
```

# API changes

None

None

# Required module changes

Have updated reference function block module


modules/ref_fb_module/src/video_player_fb_impl.cpp
```diff
@@ -0,0 +1,207 @@
...
+{
+    return FunctionBlockType("RefFBModuleVideoPlayer",
+                             "Video Player",
+                             "Video playback and visualization"
+    );
+}
+
+void VideoPlayerFbImpl::initProperties()
+{
+    objPtr.addProperty(FunctionPropertyBuilder("OpenWindow", ProcedureInfo()).setReadOnly(true).build());
```
JakobVokac, line 44: You add a read only property here with a procedure call, that is not set anywhere inside the class. What exactly is the purpose of it or am I missing something here?

JakobVokac (commented):
Generally looks good. A couple of comments. Also, could you provide an example with the resource files so I can run it?

modules/ref_fb_module/src/video_player_fb_impl.cpp
denise-opendaq: This proporty was addee to make posible re open the window without recreating fb. The usage is shown on the video 
https://github.com/user-attachments/assets/f4813003-c19f-49e6-b14d-fbab976e6276

JakobVokac (approved):

PR #871, opened 2025-07-31
Parallelize method setOperationModeRecursive()

janm-dw:
# Brief

Parallelize method setOperationModeRecursive()

# Description

A method called setOperationModeRecursive() is used to set the operation mode for all subdevices. Since this operation can be slow, a change was made to update the operation mode in parallel for all subdevices at the same hierarchical level.

# API changes

none

none

# Required module changes

none

### DIFF
--- a/core/opendaq/device/include/opendaq/device_impl.h
+++ b/core/opendaq/device/include/opendaq/device_impl.h
@@ -1198,13 +1198,35 @@ ErrCode GenericDevice<TInterface, Interfaces...>::setOperationModeRecursive(Oper
 {
     ErrCode errCode = setOperationMode(modeType);
     OPENDAQ_RETURN_IF_FAILED(errCode);
-    
-    for (const DevicePtr & dev: this->devices.getItems())
+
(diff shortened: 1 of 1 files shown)


core/opendaq/device/include/opendaq/device_impl.h
```diff
@@ -1198,13 +1198,35 @@ ErrCode GenericDevice<TInterface, Interfaces...>::setOperationModeRecursive(Oper
...
     }
-    return OPENDAQ_SUCCESS;
+
+    try
+    {
+        for (auto& job : jobs)
+        {
+            const auto jobResult = (ErrCode) job.getResult();
+            if (errCode == OPENDAQ_SUCCESS && jobResult != OPENDAQ_SUCCESS)
+                errCode = jobResult;
```
denise-opendaq, line 1220: if the device is not supporting the selected operation mode, it would return OPENDAQ_IGNORED. So even though only one sub device did not set operation mode. i probably would return the error code which was returned by setOperationMode(modeType) if there was no errors from subdevices

PR #876, opened 2025-08-07
Fix advertising LT streaming caps when discovery disabled

NikolaiShipilov:
# Brief

Ensures proper advertisement of LT streaming capabilities when discovery is disabled within the new LT server.

# Description

- Implements missing logic in the new LT server to register LT streaming capability in the `DeviceInfo` object. This enables automatic streaming connection setup when adding the device via OPC UA or Native configuration protocols, even if mDNS-based discovery for LT streaming is turned off.
- Adds validation to prevent simultaneous adding of multiple LT server versions per single device, disallowing it to avoid any conflicts
- Both new and old LT servers register server capability under ID "OpenDAQLTStreaming" but new server sets version field to "OpenDAQNewLTStreaming" while old one leaves this field empty


denise-opendaq (approved):
looks good for me :)

PR #886, opened 2025-08-21
Simple signal generator/simulator example

JakaMohorko:
# Brief

Simple signal generator example based on reference device

### DIFF
--- a/examples/modules/simulator_device_module/src/signal_generator.cpp
+++ b/examples/modules/simulator_device_module/src/signal_generator.cpp
@@ -137,7 +137,7 @@ PacketPtr SignalGenerator::generateData(const DataPacketPtr& domainPacket)
     }
     
     samplesGenerated += newSampleCount;
-    return std::move(dataPacket);
+    return dataPacket;
 }
 
 DataDescriptorPtr SignalGenerator::buildDescriptor() const
@@ -172,7 +172,6 @@ void SignalGenerator::waveformChanged(PropertyObjectPtr&, PropertyValueEventArgs
         {"Unknown", WaveformProperty::Unknown},
     };
     
-    const auto prevWaveform = waveformType;
     auto propName = args.getProperty().getName();
     auto value = args.getValue();
 


examples/modules/ref_device_module/src/ref_channel_impl.cpp
```diff
@@ -386,7 +386,7 @@ std::tuple<PacketPtr, PacketPtr> RefChannelImpl::generateSamples(int64_t curTime
 
         if (!dataPacket.assigned())
         {
-            DAQ_THROW_EXCEPTION(InvalidParameterException, "We have created an empty packet, OH NO");
+            DAQ_THROW_EXCEPTION(InvalidParameterException, "Packet creation failed!");
```
denise-opendaq, line 389: I like the previous message more :)

examples/modules/simulator_device_module/src/simulator_device_impl.cpp
```diff
@@ -0,0 +1,334 @@
...
+    const uint64_t packetTime = samplesGenerated * deltaTicks + ticksSinceEpochToDeviceStart;
+    uint64_t sampleRemainder = newSampleCount % dividerLcm;
+    uint64_t actualPacketSize = newSampleCount - sampleRemainder;
+
+    if (actualPacketSize > 0)
+    {
+        timeSignal.sendPacket(std::move(DataPacket(timeSignal.getDescriptor(), actualPacketSize, packetTime)));
+    }
+
+    samplesGenerated += actualPacketSize;
```
denise-opendaq, line 270: I have a question why are we storing samples Generated but not last packet offset. I think if we would store packer offset we wont need to recalculate samplesGenerated in case of changing frequency

examples/modules/simulator_device_module/src/simulator_device_impl.cpp
```diff
@@ -0,0 +1,334 @@
...
+    auto lock = getUniqueLock();
+
+    while (!stopAcq)
+    {
+        const auto loopTime = milli(acqLoopTime);
+        const auto time = std::chrono::high_resolution_clock::now();
+        const auto loopDuration = std::chrono::duration_cast<milli>(time - prevLoopTime);
+        const auto waitTime = loopDuration.count() >= loopTime.count() ? milli(0) : milli(loopTime.count() - loopDuration.count());
+        
+        cv.wait_for(lock, waitTime);
```
denise-opendaq, line 302: you can just use cv.wait_for(lock, prevLoopTime + loopTime);

and after that use 
const auto time = std::chrono::high_resolution_clock::now();

denise-opendaq (approved):
time signal looks way simple which is good

PR #893, opened 2025-08-28
Rework disabled permission manager

NikolaiShipilov:
# Brief

Enable or disable access control without requiring module rebuilds.

# Description

Previously, toggling the CMake option `OPENDAQ_ENABLE_ACCESS_CONTROL` and rebuilding the openDAQ core libraries also required rebuilding device and function block modules. Without this, a mismatch occurred between the regular and disabled permission managers when inheriting from a parent to a child component. This led to `NotAssignedException` **" Parameter config must not be null in the function `extend`"**, since one manager allowed `nullptr` permission objects while the other disallowed them.

The introduced changes remove the disabled permission manager and provide a single permission manager factory. This factory conditionally instantiates either the regular or the former disabled permission manager, depending on the `OPENDAQ_ENABLE_ACCESS_CONTROL` setting.

This approach eliminates conditional `#ifdef` branching in the `PropertyObject` implementation, making module code independent of the `OPENDAQ_ENABLE_ACCESS_CONTROL` option and thus compatible with both variants of the openDAQ core libraries (built with the option ON or OFF).

# Required module changes

No changes to module source code are required. However, modules must be rebuilt using the patched openDAQ include files and updated core library binaries for the fix to take effect.

### DIFF
--- a/bindings/c/include/ccoreobjects/permission_manager.h
+++ b/bindings/c/include/ccoreobjects/permission_manager.h
@@ -5,7 +5,7 @@
 //     Changes to this file may cause incorrect behavior and will be lost if
 //     the code is regenerated.
 //
-//     RTGen (CGenerator v0.7.0) on 03.06.2025 22:05:13.
+//     RTGen (CGenerator v0.7.0) on 28.08.2025 07:28:15.
 // </auto-generated>
 //------------------------------------------------------------------------------
 
@@ -43,7 +43,6 @@ extern "C"
     daqErrCode EXPORTED daqPermissionManager_setPermissions(daqPermissionManager* self, daqPermissions* permissions);
     daqErrCode EXPORTED daqPermissionManager_isAuthorized(daqPermissionManager* self, daqUser* user, daqPermission permission, daqBool* authorizedOut);
     daqErrCode EXPORTED daqPermissionManager_createPermissionManager(daqPermissionManager** obj, daqPermissionManager* parent);
-    daqErrCode EXPORTED daqPermissionManager_createDisabledPermissionManager(daqPermissionManager** obj);
 
 #ifdef __cplusplus
 }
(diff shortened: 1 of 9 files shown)


NikolaiShipilov:
As an additional note - the following patch also fixes the `NotAssignedException` -  
**"Parameter config must not be null in the function `extend`"** -  
occured when modules and core libraries are built with different values of the `OPENDAQ_ENABLE_ACCESS_CONTROL` option :
```diff
--- a/core/coreobjects/src/permission_manager_impl.cpp
+++ b/core/coreobjects/src/permission_manager_impl.cpp
@@ -5,6 +5,7 @@
 #include <coretypes/validation.h>
 #include <coreobjects/permissions_builder_factory.h>
 #include <coreobjects/permission_manager_factory.h>
+#include "coreobjects/permission_mask_builder_factory.h"
 
 BEGIN_NAMESPACE_OPENDAQ
 
@@ -205,8 +206,13 @@ ErrCode DisabledPermissionManagerImpl::removeChildManager(IPermissionManager* /*
     return OPENDAQ_SUCCESS;
 }
 
-ErrCode DisabledPermissionManagerImpl::getPermissions(IPermissions** /*permisisonConfigOut*/)
+ErrCode DisabledPermissionManagerImpl::getPermissions(IPermissions** permisisonConfigOut)
 {
+    OPENDAQ_PARAM_NOT_NULL(permisisonConfigOut);
+
+    PermissionsPtr permissions =
+        PermissionsBuilder().assign("everyone", PermissionMaskBuilder().read().write().execute()).build();
+    *permisisonConfigOut = permissions.detach();
     return OPENDAQ_SUCCESS;
 }
```
@janm-dw , could you please give some clue on how both solutions affect the optimization introduced with `OPENDAQ_ENABLE_ACCESS_CONTROL`?  And do the PR changes align with this idea or break it?

janm-dw:
@NikolaiShipilov I would store default permissions in a static const variable since building them likely takes some time.

PR #908, opened 2025-09-06
Add support for Intel-LLVM compiler

NikolaiShipilov:
# Brief

Enable building openDAQ with the Intel oneAPI LLVM compiler

# Description

- Patch TaskFlow to resolve issues with Intel-LLVM compiler detection  
- Add CMake build presets for Intel-LLVM compiler  
- Provide necessary build flags for Intel oneAPI LLVM compiler  
- Remove unused class members in reference modules and address other warnings  
- Fix `cxxabi.h` include errors  
- Introduce Ubuntu and Windows CI jobs for Intel-LLVM compiler — verified as passing, then disabled

### Known issues

- `test_core_types` fails on Ubuntu 22.04 in `TEST_F(JsonDeserializerTest, floatMin)` during float minimum deserialization. This is possibly caused by a bug in the Intel math runtime library. However, the same test passes locally on Ubuntu 20.04.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -181,7 +181,7 @@ jobs:
           call "C:\Program Files (x86)\Intel\oneAPI\setvars.bat"
           mkdir build
           cd build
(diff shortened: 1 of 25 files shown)


NikolaiShipilov:
> LGTM, except I don't understand why "Windows Intel-LLVM Release" is commented out in the CI matrix. Is there a particular reason for this?

This job is disabled as the existing approach of installing the compiler on the Windows runner doesn’t seem very robust - it relies on a hardcoded download link for the installer, which could become invalid at any time, I guess. 
Also the download and installation process takes around 30 minutes.

PR #917, opened 2025-09-11
Serialization improvements: DeviceInfo

Yarikk26:
# Brief
`DeviceInfo` serialization for update improvements
# Description
- serialize custom properties only

Expected file size impact:
- ~1000/13900 = 7%
# Usage example
**None**
# API changes
**None**
**None**
# Required module changes
**None**

### DIFF
--- a/core/opendaq/device/include/opendaq/device_info_impl.h
+++ b/core/opendaq/device/include/opendaq/device_info_impl.h
@@ -144,7 +144,6 @@ public:
 
     // IUpdatable
     ErrCode INTERFACE_FUNC updateInternal(ISerializedObject* obj, IBaseObject* context) override;
-    ErrCode INTERFACE_FUNC serializeForUpdate(ISerializer* serializer) override;
 
 protected:
     ErrCode createAndSetStringProperty(const StringPtr& name, const StringPtr& value);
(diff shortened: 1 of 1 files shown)


core/opendaq/device/include/opendaq/device_info_impl.h
```diff
@@ -1296,6 +1298,82 @@ ErrCode DeviceInfoConfigImpl<TInterface, Interfaces...>::updateInternal(ISeriali
...
 }
 
+template <typename TInterface, typename... Interfaces>
+ErrCode DeviceInfoConfigImpl<TInterface, Interfaces...>::serializeForUpdate(ISerializer* serializer)
+{
+    OPENDAQ_PARAM_NOT_NULL(serializer);
+
+    serializer->startTaggedObject(this);
+
+    auto name = this->objPtr.getClassName();
```
denise-opendaq: not sure that device info will ever have class name

denise-opendaq (approved):
lg

core/opendaq/device/include/opendaq/device_info_impl.h
JakaMohorkoDS: Agreed. Also, we should never serialize that when serializing for update anyhow. When updating we really need to only restore the property values. In the case of device info, it could be argued that only user-changeable ones should be restored.

PR #932, opened 2025-09-24
added BlockSize edge case fix

JakobVokac:
# Brief

Added conditional statement for FFT FB BlockSize property, to check if < 1, which should never be the case.

### DIFF
--- a/examples/modules/ref_fb_module/src/fft_fb_impl.cpp
+++ b/examples/modules/ref_fb_module/src/fft_fb_impl.cpp
@@ -37,7 +37,7 @@ FFTFbImpl::~FFTFbImpl()
 
 void FFTFbImpl::initProperties()
 {
-    auto blockSizeProp = IntPropertyBuilder("BlockSize", defaultBlockSize).build();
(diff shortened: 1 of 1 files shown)


examples/modules/ref_fb_module/src/fft_fb_impl.cpp
```diff
@@ -55,6 +54,13 @@ void FFTFbImpl::propertyChanged(bool configure)
 
 void FFTFbImpl::readProperties()
 {
+    size_t newBlockSize = objPtr.getPropertyValue("BlockSize");
+    if (newBlockSize < 1)
+    {
+        LOG_W("Cannot set BlockSize < 1, setting back to previous value {}", blockSize/2);
+        objPtr.setPropertyValue("BlockSize", blockSize / 2);
+    }
+
```
JakaMohorkoDS, line 63: This should instead be done with:

```cpp
auto blockSizeProp = IntPropertyBuilder("BlockSize", defaultBlockSize).setMinValue(1).build();
```

JakaMohorkoDS (changes requested):

JakaMohorkoDS (approved):

PR #943, opened 2025-10-07
Multi reader: Fixes, Dynamic input port usage, SumReader FB

JakaMohorko:
# Brief
Set of multi reader fixes. Allow releasing input port ownership from multi reader. Adds SumReader FB that showcases dynamic input port usage with multi reader.

# Description

List of changes:

1. Fix multi reader crash on reading explicit-type domain signals
2. Add reader `dispose` override that releases ownership of input ports
3. Add function block example that sums equal-rate signals using the multi reader
4. Do not throw exceptions on multi reader creation when domains are not valid
5. Fix check for non-matching sampling rates when `allowDifferentRates` is set to `false`
6. Add `IReaderConfig` method that checks whether the reader is valid

### DIFF
--- a/core/coretypes/include/coretypes/ctutils.h
+++ b/core/coretypes/include/coretypes/ctutils.h
@@ -179,26 +179,30 @@ inline std::string ErrorCodeMessage(ErrCode errCode)
     return ss.str();
 }
 
-inline std::string getErrorInfoMessage(ErrCode errCode)
+inline std::string getErrorInfoMessage(ErrCode errCode, bool clearErrorInfo = false)
 {
     if (OPENDAQ_SUCCEEDED(errCode))
         return "";
 
     std::string message;
-    daq::IString* errorMessage = nullptr;
+    IString* errorMessage = nullptr;
     ErrCode err = daqGetErrorInfoMessage(&errorMessage);
     if (errorMessage)
     {
         if (err == errCode)
         {
-            daq::ConstCharPtr msgCharPtr = nullptr;
+            ConstCharPtr msgCharPtr = nullptr;
             errorMessage->getCharPtr(&msgCharPtr);
-            errorMessage->releaseRef();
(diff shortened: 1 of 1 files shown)


examples/modules/ref_fb_module/src/sum_reader_fb_impl.cpp
```diff
@@ -0,0 +1,291 @@
...
+    catch (const std::exception& e)
+    {
+        setComponentStatusWithMessage(ComponentStatus::Warning, fmt::format("Failed to set descriptor for power signal: {}", e.what()));
+        reader.setActive(False);
+    }
+}
+
+void SumReaderFbImpl::onConnected(const InputPortPtr& inputPort)
+{
+    auto lock = this->getAcquisitionLock2();
```
Andrej-Peplinski-HBK: It would be appreciated if there would be a comment explaining why we use the `getAcquisitionLock2` method here.

In https://github.com/openDAQ/openDAQ/pull/894 there is an explanation that talks about the need to use this method when we override the locking strategy using `objPtr.asPtr<IPropertyObjectInternal>().setLockingStrategy(LockingStrategy::InheritLock)`. In this pull request however, I don't see that we override the locking strategy - which confuses me.

core/coretypes/include/coretypes/ctutils.h
```diff
@@ -179,6 +179,28 @@ inline std::string ErrorCodeMessage(ErrCode errCode)
...
+            errorMessage->getCharPtr(&msgCharPtr);
+            errorMessage->releaseRef();
+            return msgCharPtr;
+        }
+    }
+
+    return "";
+}
+
 inline void checkErrorInfo(ErrCode errCode)
```
denise-opendaq: you can use getErrorInfoMessage in checkErrorInfo to remove duplication of the code

PR #953, opened 2025-10-15
Store version info in function blocks and devices in configuration

dejancrn:
# Brief

Store version info of function blocks and devices in the configuration.

# Description

Store module version info with each function block and device. This will enable us to maintain backward compatibility in the future. Reworked reference modules to return their function blocks' and devices' version properly. This rework is optional and maintains source and binary compatibility.

# Usage example

### DIFF
--- a/core/opendaq/component/include/opendaq/component_impl.h
+++ b/core/opendaq/component/include/opendaq/component_impl.h
@@ -197,6 +197,8 @@ protected:
 
     virtual void onOperationModeChanged(OperationModeType modeType);
 
+    static VersionInfoPtr parseVersionString(const std::string& input);
+
 private:
     EventEmitter<const ComponentPtr, const CoreEventArgsPtr> componentCoreEvent;
 };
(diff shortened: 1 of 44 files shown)


dejancrn:
I thought about serializing the complete DeviceType and FunctionBlockType. However, the common complaint is that the JSON configuration files are too large. While it makes sense to serialize ComponentType for native protocol completely, it's of no benefit for configuration files. It only makes files larger. The configuration files should only store what can be changed by the user. The version information is provided for the possibility of transforming older configurations of FBs and devices to newer ones.

NikolaiShipilov:
> I thought about serializing the complete DeviceType and FunctionBlockType. However, the common complaint is that the JSON configuration files are too large. While it makes sense to serialize ComponentType for native protocol completely, it's of no benefit for configuration files. It only makes files larger. The configuration files should only store what can be changed by the user. The version information is provided for the possibility of transforming older configurations of FBs and devices to newer ones.

Ok. Under this logic, the PR lgtm and can be merged. 
And the approach for "mirrored" modules info could be discussed later and then be addressed by another pull-request.

PR #955, opened 2025-10-15
Fix large log files download

NikolaiShipilov:
# Brief

Resolves issues with downloading large log files (greater than 16 MB) and implements graceful rejection of payloads exceeding native protocol limits (such as log files over 256 MB).

Related PR https://github.com/openDAQ/libNativeStreaming/pull/26

### DIFF
--- a/changelog/changelog_3.20-3.30.md
+++ b/changelog/changelog_3.20-3.30.md
@@ -27,6 +27,7 @@
 
 ## Bug fixes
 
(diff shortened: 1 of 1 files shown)


denise-opendaq (commented):
I had an idea to call on the client get file log infos and read the size of file to load. and if the size is more than protocol limitation - make a couple calls of chunks

JakaMohorkoDS (approved):

PR #958, opened 2025-10-17
Bug save load not showing connected input ports on client

denise-opendaq:
# Brief

Fix input port connections not restored on client side when loading configuration on gateway devices with signals from remote sub-devices

# Description

- Fix deserialization of input port connections on gateway devices when configuration is loaded from client
- Change configuration save/load strategy: client now requests configuration directly on remote device itself and puts it inside client config, instead of client doing save configuration of remote devices as part of itself
- Resolve signal path confusion caused by client-side serialization of remote device topologies

Fixing broken topology
```
Client
  └─ Gateway Device (remote)
       ├─ Function Block
       │    └─ Input Port → [BROKEN: unassigned on client side] 
       └─ Sub-device (remote)
            └─ AI Signal
```

# Usage example
see test test_device_modules.NativeDeviceModulesTest.SaveLoadGateway

# API changes

```diff
+ [func] MirroredDeviceConfig::serializeForUpdateLocally(ISerializer* serializer)
```

- works only with both new server and client
- config json should be regenerated (otherwise will be applied all strategy) 

# Required module changes
None

### DIFF
--- a/core/opendaq/device/include/opendaq/device_impl.h
+++ b/core/opendaq/device/include/opendaq/device_impl.h
@@ -202,7 +202,7 @@ protected:
                                   LockingStrategy lockingStrategy = LockingStrategy::OwnLock);
 
     void serializeCustomObjectValues(const SerializerPtr& serializer, bool forUpdate) override;
-    void serializeConnectionValues(const SerializerPtr& serializer) override;
+    void serializeComponentCreationValues(const SerializerPtr& serializer) override;
 
     void updateFunctionBlock(const std::string& fbId,
                              const SerializedObjectPtr& serializedFunctionBlock,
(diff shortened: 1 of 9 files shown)


core/opendaq/device/include/opendaq/device_impl.h
```diff
@@ -202,6 +202,8 @@ class GenericDevice : public SignalContainerImpl<TInterface, IDevicePrivate, IDe
                                   LockingStrategy lockingStrategy = LockingStrategy::OwnLock);
 
     void serializeCustomObjectValues(const SerializerPtr& serializer, bool forUpdate) override;
+    void serializeConnectionValues(const SerializerPtr& serializer) override;
```
JakaMohorkoDS: I think we should name this something similar to "serializeDeviceCreationValues".

PR #974, opened 2025-11-13
Audio device refactor

JakaMohorko:
# Brief

Reorganizes the audio device implementation for greater clarity. Adds automatic device/FB version info setting.

# Description

## Audio device

The audio device implementation has not been revisited since initial development. Since then, while functional, openDAQ best practices and development guidelines have evolved. This pull request refactors the device and channel implementation, removing obsolete parts and reorganizing configuration methods.

## Version info in device/FB type

Previously, all devices and function block type's version info field had to be manually set. This logic has now been moved to the module manager and automated.

### DIFF
--- a/changelog/changelog_3.30-3.40.md
+++ b/changelog/changelog_3.30-3.40.md
@@ -11,6 +11,8 @@
 
 ## Misc
 
+- [#974](https://github.com/openDAQ/openDAQ/pull/974) Reorganizes the audio device implementation for greater clarity. Adds automatic device/FB version info setting.
+
 ## Required application changes
 
 ## Required module changes


core/opendaq/modulemanager/src/module_manager_impl.cpp
```diff
@@ -1060,7 +1072,22 @@ ErrCode ModuleManagerImpl::createFunctionBlock(IFunctionBlock** functionBlock, I
...
+            ModuleInfoPtr moduleInfo;
+            err = library.module->getModuleInfo(&moduleInfo);
+            OPENDAQ_RETURN_IF_FAILED(err);
+
+            auto fbType = fbPtr.getFunctionBlockType();
+            if (fbType.assigned())
+                fbType.asPtr<IComponentTypePrivate>().setModuleInfo(moduleInfo);
+        }
+
+        return OPENDAQ_SUCCESS;
```
dejancrn: How is this going to work for native module function blocks? Current implementation deserializes the function block and it uses serialized module version to construct ModuleInfo with "__unknown" id and correct version. It looks like this will override the constructed module info and will use client side native module info.

JakaMohorkoDS: This has been discussed in the past and I believe we should have the option of getting remote module information separately. 

A device added via the native configuration client should have the module info of the native client used. The device type of said device is "Native device".

I'm however not entirely certain how to handle child objects that were created on the server-side. In general, only top-level objects created by modules have this information at the moment. Signals, Input ports, Channels... Lack this information. Extending this to native would imply that the child objects of the device have no module information.

I think this should really be looked at from the user perspective and how they might interact with the API to retrieve remote module info.

denise-opendaq (approved):

PR #980, opened 2025-11-19
[TBBAS-2776] Feature/python add device with config rework

tomaz-cvetko:
# Brief

Rework for the Add device with config dialog in GUI demo (also known as `py -m opendaq`). The new UI displays configuration options that are relevant to the selected device.

# Description

- Filter the default add device config so that only relevant configuration options remain. This is achieved by removing default configurations of openDAQ modules that are not supported by the device.
- Added an example of filtering in `filter_add_device_configuration.py`
- It's now possible to right click to "Quick add" or "Add with config" in the "Add device" dialog
- Reworked "Add with config" dialog UI to now feature:
  - configuration protocol selection
  - streaming protocol (multi)selection
  - editing selected parts of configuration in tabs
  - status label that reports the validity of the selected options
- Improved "Add with config" dialog logic:
  - selected streaming protocols are now added to `AllowedStreamingProtocols`,
  - connection string is deduced from the selected protocols,
  - configuration editor displays relevant parts of configuration in tabs

Adding device with configuration includes generation of a default add device configuration that is a union of configurations for all available openDAQ protocols. Usually many of these protocols are not supported by the device that the user is trying to connect to. It is only sensible to display parts of configuration that affect the target device. Using device's server capabilities, the supported protocols are deduced and configurations of unsupported protocols are removed from the Device and Streaming sections of the default configuration. In General section some options also apply only to remote devices and devices supporting the openDAQ native protocol, which are removed if not applicable.

To enforce that the selected configuration protocol (or a streaming-only) is used, the connection string of the protocol is used instead of the smart openDAQ connection string.

The configuration of streaming protocols can only be edited in one place (in Streaming section) and is then synced to the Device section in case the streaming protocol was selected as the primary connection type.

# API changes

No API changes.

No required changes.

# Required module changes

No required changes.

### DIFF
--- a/examples/applications/python/GUI Application/gui_demo/components/add_config_dialog.py
+++ b/examples/applications/python/GUI Application/gui_demo/components/add_config_dialog.py
@@ -15,7 +15,7 @@ _STREAMING = "Streaming"
 _GENERAL = "General"
 _ASP = "AllowedStreamingProtocols"
 _ACS = "AutomaticallyConnectStreaming"
-_PAT = "PrimaryAdressType"
+_PAT = "PrimaryAddressType"
 
 class AddConfigDialog(Dialog):
     @classmethod


examples/applications/python/Discovery/filter_add_device_configuration.py
```diff
@@ -0,0 +1,158 @@
...
+    # Local devices have a single supported protocol that can be deduced from the devices connection string.
+    supported_prefix = device_info.connection_string.split("://")[0]
+
+    # Search available device types for protocol ID matching the connection string prefix
+    available_device_types = instance.available_device_types
+    for protocol_id, device_type in available_device_types.items():
+        dt_dict = device_type.as_dictionary
+
+        if dt_dict["Prefix"] == supported_prefix:
+            return [protocol_id]
```
JakaMohorkoDS: Shouldn't this return a list of protocol IDs, not only the one that matches the prefix? This would also not match in case of the "daq://" prefix.

examples/applications/python/Discovery/filter_add_device_configuration.py
```diff
@@ -0,0 +1,158 @@
...
+    instance_builder = daq.InstanceBuilder()
+    instance_builder.add_discovery_server("mdns")
+    instance_builder.set_root_device('daqref://device0', config)
+    instance_builder.module_path = daq.OPENDAQ_MODULES_DIR
+    server_instance = instance_builder.build()
+
+    server_config = daq.PropertyObject()
+    for protocol_name in protocols:
+        server_instance.add_server(protocol_name, server_config).enable_discovery()
+    return server_instance
```
JakaMohorkoDS: There is an utility python file available that already does something very similar. That could instead be slightly tweaked with additional parameters to facilitate this use case.

In general it would be good if we kept common methods in the utilities.

examples/applications/python/Discovery/filter_add_device_configuration.py
```diff
@@ -0,0 +1,158 @@
...
+                "AutomaticallyConnectStreaming"])
+
+def print_configuration(config, level=0):
+    _indent = level * "  "
+    for _prop in config.visible_properties:
+        if _prop.value_type == daq.CoreType.ctObject:
+            print(_indent, f"{_prop.name}:")
+            print_configuration(_prop.value, level + 1)
+        else:
+            print(_indent, f"{_prop.name} = '{_prop.value}'")
```
JakaMohorkoDS: Same, could be reused from utils.

examples/applications/python/GUI Application/gui_demo/components/add_config_dialog.py
```diff
@@ -1,37 +1,354 @@
...
+            if property.name not in supported_protocols:
+                device_section.remove_property(property.name)
+
+    def filter_streaming_section(self, config, supported_protocols):
+        streaming_section = config.get_property_value(_STREAMING)
+        for property in self.context.properties_of_component(streaming_section):
+            if property.name not in supported_protocols:
+                streaming_section.remove_property(property.name)
+
+    def filter_general_section(self, config, supported_protocols):
```
JakaMohorkoDS: As mentioned above, this likely must only hide the general properties. Otherwise a patch might be needed that makes sure not to use any of the props if they're not available.

Currently, we only check whether the config object is a `defaultAddDeviceConfig`. If yes, we expect those fields to be there.

JakaMohorkoDS (changes requested):
The implementation looks really good! I just added some minor comments. Still need to test out the application itself, as I did not review the UI code in detail.

I will submit another review once done with testing.

tomaz-cvetko:
>     * I would hide the `AllowedStreamingProtocols` field, as you are modifying it on-checkbox-click. Same goes for "AutomaticallyConnectStreaming".

Is it ok to edit the "AllowedStreamingProtocols" (empty when no streaming is selected in UI), but leave the "AutomaticallyConnectStreaming" as `True`?

examples/applications/python/Discovery/filter_add_device_configuration.py
```diff
@@ -0,0 +1,158 @@
...
+    for protocol_name in protocols:
+        server_instance.add_server(protocol_name, server_config).enable_discovery()
+    return server_instance
+
+def get_available_devices(parent):
+    try:
+        available_devices = parent.available_devices
+        return available_devices
+    except:
+        return []
```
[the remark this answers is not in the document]
tomaz-cvetko: Yes, removed it

examples/applications/python/Discovery/filter_add_device_configuration.py
tomaz-cvetko: Tweaked this with hiding mechanism.

examples/applications/python/GUI Application/gui_demo/utils.py
```diff
@@ -313,3 +313,35 @@ def is_device_connected(device: daq.IDevice):
...
+
+    for property in target.all_properties:
+            if not source.has_property(property.name):
+                print(f"WARNING: IPropertyObj mismatch, source does not contain property {property.name}")
+                continue
+
+            if property.value_type == daq.CoreType.ctObject:
+                update_properties(property.value, source.get_property_value(property.name))
+            elif property.value_type == daq.CoreType.ctStruct:
+                # Skip as structs are immutable
```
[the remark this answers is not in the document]
tomaz-cvetko: Replaced the function with `set_property_value` based approach.

examples/applications/python/GUI Application/gui_demo/utils.py
```diff
@@ -313,3 +313,35 @@ def is_device_connected(device: daq.IDevice):
...
+        for k in target_dict.keys:
+            if source_dict.has_key(k):
+                target_dict.set(k, source_dict.get(k))
+
+    def update_list(target_list, source_list):
+        target_list.clear()
+        for item in source_list:
+            target_list.push_back(item)
+
+    for property in target.all_properties:
```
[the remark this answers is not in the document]
tomaz-cvetko: Rewrote the suggested method in python.

examples/applications/python/GUI Application/gui_demo/utils.py
```diff
@@ -317,31 +317,21 @@ def update_properties(target: daq.IPropertyObject, source: daq.IPropertyObject):
...
-                update_dict(property.value, source.get_property_value(property.name))
-            else:
-                property.value = source.get_property_value(property.name)
+        if target_prop.read_only:
+            continue
 
+        if source_prop.value_type == daq.CoreType.ctObject:
+            update_properties(target_prop.value, source_prop.value)
+        else:
+            target.set_property_value(prop_name, source_prop.value)
```
tomaz-cvetko: Use `set_property_value` to handle all property types equally.

JakaMohorkoDS (approved):
Lgtm! 1 minor comment, but the UI looks really good now.

PR #985, opened 2025-12-02
Subtask: invalid multireader results in a warning state of the recorder

tomaz-cvetko:
# Brief

Add a test case with a half-rate signal to make sure recorder gets invalidated when different sample rate signals invalidate the multireader.

# Description

- Recorder detects Multireader invalid state (already implemented in sum_reader)
- Tests to confirm that.

### DIFF
--- a/examples/modules/basic_csv_recorder_module/src/multi_csv_recorder_impl.cpp
+++ b/examples/modules/basic_csv_recorder_module/src/multi_csv_recorder_impl.cpp
@@ -130,8 +130,14 @@ MultiCsvRecorderImpl::MultiCsvRecorderImpl(const ContextPtr& context,
 ErrCode MultiCsvRecorderImpl::startRecording()
 {
     auto lock = getRecursiveConfigLock();
-    recordingActive = true;
     reconfigureWriter();
+    if (!filePath.has_value() || !writer.has_value())
+    {
+        LOG_I("Start recording FAILED.")
+        return OPENDAQ_ERR_INVALIDSTATE;
+    }
+    LOG_I("Recording to: {}", writer.value().getFilename());
+    recordingActive = true;
 
     return OPENDAQ_SUCCESS;
 }
@@ -139,9 +145,7 @@ ErrCode MultiCsvRecorderImpl::startRecording()
 ErrCode MultiCsvRecorderImpl::stopRecording()
 {
     auto lock = getRecursiveConfigLock();
-    recordingActive = false;
-    // Close the file
-    writer = std::nullopt;
+    stopRecordingInternal();
 
     return OPENDAQ_SUCCESS;
 }
@@ -166,8 +170,8 @@ void MultiCsvRecorderImpl::initProperties()
 {
     this->tags.add(Tags::RECORDER);
 
-    objPtr.addProperty(StringProperty(Props::PATH, ""));
-    objPtr.getOnPropertyValueWrite(Props::PATH) += std::bind(&MultiCsvRecorderImpl::onPropertiesChanged, this);
+    objPtr.addProperty(StringProperty(Props::DIR, ""));
+    objPtr.getOnPropertyValueWrite(Props::DIR) += std::bind(&MultiCsvRecorderImpl::onPropertiesChanged, this);
 
     objPtr.addProperty(StringProperty(Props::BASENAME, "output"));
     objPtr.getOnPropertyValueWrite(Props::BASENAME) += std::bind(&MultiCsvRecorderImpl::onPropertiesChanged, this);
@@ -292,6 +296,10 @@ void MultiCsvRecorderImpl::configureWriter(const DataDescriptorPtr& domainDescri
         setComponentStatus(ComponentStatus::Ok);
         reader.setActive(True);
 
+        if (!filePath.has_value())
+        {
+            return;
+        }
         fs::path outputFile = getNextCsvFilename(filePath.value(), fileBasename, timestampEnabled);
 
         // Replace the csv writer (can it ever survive a reconfigure?)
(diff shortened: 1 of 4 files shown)


examples/modules/basic_csv_recorder_module/src/multi_csv_recorder_impl.cpp
```diff
@@ -290,61 +359,71 @@ void MultiCsvRecorderImpl::onDataReceived()
...
+        return;
+    }
 
-            bool domainChanged = false;
-            bool valueSigChanged = false;
+    const auto eventPackets = status.getEventPackets();
+    if (eventPackets.getCount() == 0)
+    {
+        return;
+    }
```
JakaMohorko: This check should be redundant.

examples/modules/basic_csv_recorder_module/src/multi_csv_recorder_impl.cpp
```diff
@@ -290,61 +359,71 @@ void MultiCsvRecorderImpl::onDataReceived()
...
-                reader = MultiReaderFromExisting(reader, SampleType::Float64, SampleType::Int64);
-                if (!reader.asPtr<IReaderConfig>().getIsValid())
-                {
-                    setComponentStatusWithMessage(ComponentStatus::Warning, "Reader failed to recover from invalid state!");
-                }
+                valueSigChanged = true;
+                valueDescriptors.pushBack(valueDescriptor);
+                cachedDescriptors[portGlobalId] = valueDescriptor;
+                SignalPtr signal = port.getSignal();
+                cachedSignalNames[portGlobalId] = signal.getName();
```
JakaMohorko: There's currently no really good approach to obtaining the signal name and it's something we need to explore. 

This approach, however, is dangerous, as we should try to avoid calling configuration methods in the data processing callbacks. This call acquires the signal's lock, which would lead to a deadlock if the input port notification would be set to `SameThread`.

I would suggest caching all signal names in the `onConnected` callback, or in start/stop recording.

examples/modules/basic_csv_recorder_module/src/multi_csv_writer.cpp
```diff
@@ -1,11 +1,85 @@
...
 #include <opendaq/custom_log.h>
+#include <boost/algorithm/string.hpp>
 
-#include <basic_csv_recorder_module/multi_csv_writer.h>
+#include <iostream>
 
 BEGIN_NAMESPACE_OPENDAQ_BASIC_CSV_RECORDER_MODULE
 
+namespace
+{
```
JakaMohorko: 👍for anonymous namespaces. Though it seems that these helpers are mainly unused as of now.

examples/modules/basic_csv_recorder_module/src/multi_csv_writer.cpp
```diff
@@ -1,11 +1,85 @@
...
+    return "";
+}
+
+std::string getDomainName(const DataDescriptorPtr& descriptor)
+{
+    if (!descriptor.assigned())
+        return "N/A";
+
+    std::ostringstream stream;
+    stream << descriptor.getName();
```
JakaMohorko: The name of the descriptor should not be used.

examples/modules/basic_csv_recorder_module/src/multi_csv_writer.cpp
```diff
@@ -76,4 +173,69 @@ void MultiCsvWriter::threadLoop()
...
+        {
+            outFile << ",";
+        }
+    }
+    outFile << "\n";
+}
+
+MultiCsvWriter::DomainMetadata MultiCsvWriter::getDomainMetadata(const DataDescriptorPtr& domainDescriptor)
+{
+    DomainMetadata metadata{StringPtr("N/A"), StringPtr("N/A"), Ratio(1, 1), 0, 1, 0, TimeProtocol::Unknown};
```
JakaMohorko: I would suggest using standard c++ types in the DomainMetadata object if it's mainly being used to write the headers. This avoids the need to create additional openDAQ objects that are not as efficient.

examples/modules/basic_csv_recorder_module/src/multi_csv_writer.cpp
tomaz-cvetko: Then let's just have "domain" as the first entry in the header row. Especially later on, it wouldn't make sense to get domain signal's name from the first signal (multi reader will effectively define its own domain that is the result of all input domains).

examples/modules/basic_csv_recorder_module/src/multi_csv_recorder_impl.cpp
tomaz-cvetko: Moved getting the signal to `updateInputPorts` method.

examples/modules/basic_csv_recorder_module/src/multi_csv_recorder_impl.cpp
```diff
@@ -215,47 +260,71 @@ void MultiCsvRecorderImpl::configure(const DataDescriptorPtr& domainDescriptor,
...
 {
     auto descriptorList = List<IDataDescriptor>();
+    auto signalNameList = List<IString>();
     for (const auto& descriptor : cachedDescriptors)
+    {
         descriptorList.pushBack(descriptor.second);
-    configure(sumDomainDataDescriptor, descriptorList);
+        signalNameList.pushBack(cachedSignalNames[descriptor.first]);
+    }
+    configure(recorderDomainDataDescriptor, descriptorList, signalNameList);
```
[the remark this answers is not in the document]
tomaz-cvetko: I now make sure that configure method is only entered with an assigned reader.

PR #986, opened 2025-12-02
Editable Structs in Python GUI application

aljazfrancic:
# Brief

Make Struct Properties editable in Python GUI demo application

# Description

Works for custom Structs

Has custom handling for:
* Range
* Unit

Displayed, but not editable:
* Ratio
* Complex Number


tomaz-cvetko (approved):
The changes I requested were implemented, I have no further complaints.

PR #991, opened 2025-12-08
Function Block configuration in Python GUI demo app

aljazfrancic:
# Brief

Add Function Block configuration to Python GUI demo app


tomaz-cvetko (approved):
Requested changes have been implemented.

PR #994, opened 2025-12-11
Update external lib mimalloc

aljazfrancic:
# Brief

Update external lib mimalloc from 2.1.1 to 3.0.11

# Description

* Update external lib mimalloc
* Prevent `Cannot find source file: test_mimalloc.cpp` (it's been moved elsewhere and is included in CMake)
* `ISignalDescriptor` to `IDataDescriptor` and correct imports

### DIFF
--- a/changelog/changelog_3.30-3.40.md
+++ b/changelog/changelog_3.30-3.40.md
@@ -16,6 +16,14 @@
 ## Misc
 
 - [#974](https://github.com/openDAQ/openDAQ/pull/974) Reorganizes the audio device implementation for greater clarity. Adds automatic device/FB version info setting.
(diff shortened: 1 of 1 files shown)


aljazfrancic:
I'm pretty sure this lib is not used anywhere, so it could be safely deleted as an alternative.

NikolaiShipilov (approved):
General question for all 3 PRs - should these changes have entry in the changelog ?

NikolaiShipilov:
> However, I can fix that. Let me know, if I should update the changelog and I'll add what's already been updated + any future external libs updates.

Yes, please add it to the changelog - I think such changes worth to be mentioned there.

PR #1001, opened 2025-12-18
Update external lib taskflow

aljazfrancic:
# Brief

Update external lib taskflow from 3.5.0 to 3.11.0

### DIFF
--- a/core/opendaq/scheduler/src/awaitable_impl.cpp
+++ b/core/opendaq/scheduler/src/awaitable_impl.cpp
@@ -50,7 +50,15 @@ ErrCode AwaitableImpl<TReturn>::getResult(daq::IBaseObject** result)
 
     if constexpr (std::is_void_v<TReturn>)
     {
-        future.get();
+        try
+        {
+            future.get();
+        }
+        catch (...)
+        {
+            // Mask exceptions from taskflow - don't rethrow
+            // This matches the behavior expected by ScheduleGraphMasksExceptions test
+        }
         *result = nullptr;
     }
     else


core/opendaq/scheduler/include/opendaq/awaitable.h
```diff
@@ -31,13 +31,6 @@ BEGIN_NAMESPACE_OPENDAQ
  */
 DECLARE_OPENDAQ_INTERFACE(IAwaitable, IBaseObject)
 {
-    /*!
-     * @brief Cancels the outstanding work if it has not already started.
-     * @param[out] canceled Is @c true if the execution was canceled
-     *                      or @c false if the execution has already completed
-     */
-    virtual ErrCode INTERFACE_FUNC cancel(Bool* canceled) = 0;
```
NikolaiShipilov: In this particular case removing method from iface is safe (method isn't used and `IAwaitable` isn't base for any other interface) but as better general practice I'd keep method anyway even if we have to disable `canceling` feature now and just return corresponding errCode from method impl
```
        return DAQ_MAKE_ERROR_INFO(
            OPENDAQ_ERR_NOT_SUPPORTED,
            "...smth..."
        );
```
this will also prevent modifying interface back and forth if we ever need to restore cancelling feature

aljazfrancic:
> Also newer version https://github.com/taskflow/taskflow/releases/tag/v4.0.0 of `Taskflow` was released few days ago - hope switching to it won't require any extra actions

Regarding Taskflow v4.0.0, from their release notes:
> Starting in v4, Taskflow will adopt the C++20 standard for better runtime performance and engineering efficiency. If your project does not support C++20, please use Taskflow v3.11.0, which can be downloaded [here](https://github.com/taskflow/taskflow/releases/tag/v3.11.0).

Since we support C++17 (GCC 7.3.1+, Clang 5+, VC++ (v14.1+)), unfortunately, I don't think we can not update beyond v3.11.0.

@NikolaiShipilov

PR #1007, opened 2025-12-23
Add an error popup when adding function block fails

tomaz-cvetko:
# Brief

Python GUI Demo: Add a popup with error description when adding a function block fails. Disable configuration button for function blocks with no properties in the configuration.

# Description

- Add an error popup to Python GUI Demo when adding a function block fails.
- Disable/enable the "add with config" button to match if any properties are available to be configured. Prevents the user from quick-adding a function block without configuration by clicking on the "Add with config" button.
- Minor refactor to enable the changes


aljazfrancic (changes requested):
Good improvement in general.

aljazfrancic (approved):

PR #1010, opened 2025-12-23
Use read callbacks for OPC UA property values and fix type checking on macOS

denise-opendaq:
# Brief

This PR refactors OPC UA client property value reading to use read callbacks via `getOnPropertyValueRead` instead of directly reading values through `getPropertyValue`. This enables proper property synchronization in gateway scenarios where OPC UA server values need to be propagated to native configuration clients.

Additionally, fixes type checking issues on macOS where pointer comparisons fail across different shared libraries.

# Usage example

See test [TEST_F(GatewayPropertySyncTest, PropertyChangePropagatesFromServerToClient) ](https://github.com/openDAQ/openDAQ/blob/b276a12d9054d2b397cfbbb7bec1b0b1ae7d690e/tests/integration/test_opendaq_device_modules/test_gateway_property_sync.cpp)

# API changes

None

None

# Required module changes

None

### DIFF
--- a/shared/libraries/opcua/opcuashared/include/opcuashared/opcuavariant.h
+++ b/shared/libraries/opcua/opcuashared/include/opcuashared/opcuavariant.h
@@ -17,7 +17,6 @@
 #pragma once
 
 #include "opcuacommon.h"
-#include <mutex>
 #include <cstring>
 #include <opcuashared/opcuanodeid.h>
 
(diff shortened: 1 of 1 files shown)


shared/libraries/opcuatms/opcuatms_client/src/objects/tms_client_property_object_impl.cpp
```diff
@@ -222,12 +161,6 @@ ErrCode INTERFACE_FUNC TmsClientPropertyObjectBaseImpl<Impl>::getOnPropertyValue
     return DAQ_MAKE_ERROR_INFO(OPENDAQ_ERR_OPCUA_CLIENT_CALL_NOT_AVAILABLE);
 }
 
-template <typename Impl>
-ErrCode TmsClientPropertyObjectBaseImpl<Impl>::getOnPropertyValueRead(IString* /*propertyName*/, IEvent** /*event*/)
-{
-    return DAQ_MAKE_ERROR_INFO(OPENDAQ_ERR_OPCUA_CLIENT_CALL_NOT_AVAILABLE);
-}
```
JakaMohorko: I'd leave this in, but use `Impl::geOnPropertyValueRead` below.

PR #1014, opened 2025-12-30
CI workflow: refactor build_macos job to use build-sdk action

alexadereyko:
# Brief

Refactor `build_macos` job to use `build-sdk` action

# Description

- Enable regular build process
- Run regular tests from run_tests preset
- Remove redundant steps
- Change runners from macos-13 to macos-15-intel and macos-latest

# Usage example

N/A

# API changes

N/A

N/A

# Required module changes

N/A

### DIFF
--- a/core/opendaq/reader/tests/test_stream_reader.cpp
+++ b/core/opendaq/reader/tests/test_stream_reader.cpp
@@ -1709,6 +1709,8 @@ TYPED_TEST(StreamReaderTest, SkipSamplesOnePacket)
 {
     this->signal.setDescriptor(setupDescriptor(SampleType::Float64));
 
+    double const multiplier = 2.5;
+
     auto reader = daq::StreamReader<TypeParam, ClockRange>(this->signal);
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -184,7 +190,6 @@ jobs:
             cmake-defines: >-
               -DCMAKE_C_COMPILER=clang
               -DCMAKE_CXX_COMPILER=clang++
-              -DOPENDAQ_VENV_PYTHON_EXECUTABLE_TESTS=ON
```
NikolaiShipilov: If I'm not mistaken this define is required for case when we use multiple py versions and since we've got rid of this in CI now we don't need it anymore. Correct ?

NikolaiShipilov (approved):

.github/workflows/ci.yml
alexadereyko: I removed that flag.

PR #1018, opened 2026-01-06
Multi reader sync optimizations for linear data rule

tomaz-cvetko:
# Brief

For linear data rule, calculate synchronization time point without iteration.

# Description

- Added some documentation to existing multi reader code Python GUI Demo functionality
- Added optimized synchronization for linear data rule domains


JakaMohorko (approved):
This looks good to me!
(but really, the multi reader code is overly complex for what it does...). 

The only thing to add would be some tests to really check that the linear offset calculation methods are used when reading linear data rule signals.

PR #1025, opened 2026-01-09
Update external lib miniaudio

aljazfrancic:
# Brief

Update external lib miniaudio from 0.11.11 to 0.11.23


aljazfrancic:
I've tested this update with mic + renderer and it works fine.

PR #1036, opened 2026-01-22
Multireader dynamic ports and unused ports

tomaz-cvetko:
# Brief

Multi reader is able to add/remove ports dynamically without being rebuilt. The ports may be marked as unused to enable the user to put in disconnected ports.

# Description

- Add/remove input ports on the same instance of multi reader.
- Mark ports as unused.
- Ignore unused ports when checking for sync failures, invalidity etc.

This solves:
- There is no need for recreating `MultiReader` instance whenever the input port is added/removed. This makes creation of a `MultiReader` with only one (empty) input port sensible, because `MultiReader` is no longer limited to the single input port it had on creation.
- `MultiReader` is able to read and synchronize data from all used ports, even if unused ports are incompatible. Unused ports will not cause the transition of `MultiReader` into an invalid state.

This doesn't solve:
- `MultiReader` still cannot transition from invalid back to a valid state. In this case creation of another instance is still the recommended way of attempting recovery.
- `MultiReader` still doesn't provide descriptive statuses which would enable the user to programmatically mark problematic input ports as unused.

Examples of how the new changes can be used are provided in `SumReaderFbImpl` and `MultiCsvRecorderImpl` function blocks.

# How to use

The new standard pattern in the function blocks using the multi reader should be:
1. Create a disconnected port in the function block.
2. Build multi reader with the disconnected port and mark it as unused.
3. When a signal is connected to the empty port, mark it as used and add a new empty one.
4. When a signal is disconnected from a port, remove it from multi reader.
5. The multi reader recovery from an invalid state still requires construction of a new instance "from old" multi reader.

Constructor:

```c++
SumReaderFbImpl::SumReaderFbImpl(const ContextPtr& ctx, const ComponentPtr& parent, const StringPtr& localId, const PropertyObjectPtr& config)
    : FunctionBlock(CreateType(), ctx, parent, localId)
{
    initComponentStatus();

    if (config.assigned())
        notificationMode = static_cast<PacketReadyNotification>(config.getPropertyValue("ReaderNotificationMode"));
    else
        notificationMode = PacketReadyNotification::Scheduler;

    createDisconnectedPort();
    createReader();
    createSignals();
}
```

NOTE: Make sure to initialize any members of the function block that influence the behavior of multi reader before the reader is created (e. g. `notificationMode`). Due to constant creations of multi reader instances, this was not important before.

The creation of multi reader:
```c++
void SumReaderFbImpl::createReader()
{
    if (!disconnectedPort.assigned())
        return;

    // Disposing the reader is necessary to release port ownership
    if (reader.assigned()
        reader.dispose();

    auto builder = MultiReaderBuilder()
                       .setDomainReadType(SampleType::Int64)
                       .setValueReadType(SampleType::Float64)
                       .setAllowDifferentSamplingRates(false)
                       .setInputPortNotificationMethod(notificationMode);

    builder.addInputPort(diconnectedPort);
    reader = builder.build();
    reader.setInputUnused(disconnectedPort.getGlobalId(), true);

    reader.setExternalListener(this->thisPtr<InputPortNotificationsPtr>());
    auto thisWeakRef = this->template getWeakRefInternal<IFunctionBlock>();
    reader.setOnDataAvailable(
        [this, thisWeakRef = std::move(thisWeakRef)]
        {
            const auto thisFb = thisWeakRef.getRef();
            if (thisFb.assigned())
                this->onDataReceived();
        });
}
```
In the function block, after a new disconnected input port is created, add it to the multi reader and mark it as unused:
```c++
reader.addInput(disconnectedPort);
reader.setInputUnused(disconnectedPort.getGlobalId(), true);
```
Then when a signal is connected to the port, mark it as used, so the multi reader will synchronize and read from it:
```c++
// Signal is connected into an empty port
auto newlyConectedPort = disconnectedPort;
// Mark as used (resume synchronization and reading)
reader.setInputUnused(newlyConnectedPort.getGlobalId(), false);
```
When the signal is disconnected from the port, we remove the port from the function block and also from the multi reader:
```c++
reader.removeInput(removedPort.getGlobalId());
```


core/opendaq/reader/src/multi_reader_impl.cpp
```diff
@@ -138,21 +154,30 @@ MultiReaderImpl::MultiReaderImpl(const MultiReaderBuilderPtr& builder)
...
         auto sourceComponents = builder.getSourceComponents();
-        checkListSizeAndCacheContext(sourceComponents);
 
-        if (notificationMethodsList.getCount() > 0 && sourceComponents.getCount() != notificationMethodsList.getCount())
-            DAQ_THROW_EXCEPTION(InvalidParameterException, "The list of source components is not of same size than the list of notification methods.");
+        if (builder.getContext().assigned())
+        {
+            // If context is provided in the builder, allow creation of empty multi reader.
+            this->context = builder.getContext();
+        }
```
JakaMohorko: Is this really necessary? We can probably just initialize the context when the 1st input is assigned. We don't have anything to do before that in either case.

core/opendaq/reader/include/opendaq/multi_reader_impl.h
```diff
@@ -194,12 +209,17 @@ class MultiReaderImpl : public ImplementationOfWeak<IMultiReader, IReaderConfig,
...
 
     SizeT minReadCount;
     PacketReadyNotification notificationMethod;
     ListPtr<PacketReadyNotification> notificationMethodsList;
+
+    const SampleType valueReadType;
+    const SampleType domainReadType;
+    const ReadMode readMode;
+
+    InputType typeOfInputs;
```
JakaMohorko: Good change 👍

JakaMohorko (changes requested):
Some general comments. Core tests are missing for the add/remove input functionality.

core/opendaq/reader/src/multi_reader_impl.cpp
```diff
@@ -346,70 +380,88 @@ ErrCode MultiReaderImpl::isDomainValid(const ListPtr<IInputPortConfig>& list) co
...
+    return std::find_if(signals.begin(),
+                        signals.end(),
+                        [&](auto& reader)
+                        {
+                            if (typeOfInputs == InputType::Ports)
+                                return id == reader.port.getGlobalId();
+                            else if (typeOfInputs == InputType::Signals)
+                                return id == reader.port.getSignal().getGlobalId();
+                            else
+                                return false;
```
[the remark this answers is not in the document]
tomaz-cvetko: I agree

PR #1042, opened 2026-01-26
Fixed externals directory for parallel config

dacuster:
# Brief

Parallel CMake configurations will fail due to FetchContent sharing a directory across configurations

# Description

 Updated the externals directory to be unique to the configuration in their respective build directories


NikolaiShipilov:
Hello @dacuster , the modification looks good to me and works fine within CI. I would just fold it into the larger refactoring made in https://github.com/openDAQ/openDAQ/pull/1049 and close this PR once #1049 is merged, rather than merging it separately now. The same code is already being touched in that bigger PR, and keeping this one open as a reminder reduces the risk of the change being lost due to merge conflicts.

Also, I'd like to clarify one thing for better understanding of issue origin: 
in the scenario where the issue occurred with the shared fetch dir, were two different conflicting CMake configuration processes triggered independently, or was it a single CMake invocation somehow initiating multiple configurations in parallel ?

PR #1043, opened 2026-01-26
RTGen enhancement for CMake configuration step

dacuster:
# Brief

Added RTGen generation at CMake configure/generation time

# Description

Added CMake option `OPENDAQ_RTGEN_ON_CMAKE_CONFIG` to allow users to use the RTGen tool to generate the smart pointer files at CMake configure time. This allows IDEs to have those files available for use during development without the need to build the project once. This doesn't change the behavior of the existing RTGen operation for build-time generation. RTGen.exe and mono programs are now found using `find_program` to make sure they exist since they are required to generate the smart pointers. Default behavior is that the CMake option is ON and doesn't break the configuration step if it fails to generate properly.


shared/tools/RTGen/bin/rtgen.exe
denise-opendaq: I see that there are some changing in the executable file? Did any changes been applied to the source code?

PR #1045, opened 2026-01-27
Implement ISynchronization and ISyncInterface

denise-opendaq:
# Brief

Implements ISynchronization/ISyncInterface — a new, typed synchronization subsystem (device clock source selection, PTP/GPS/internal interfaces, clock type reporting) — alongside a minor extension of IReferenceDomainInfo to carry a list of contributing reference domain IDs.

rtget supports @deprecated attribute

Note : this API should replace ISyncComponent 

# Description

# Synchronization interfaces

## Overview

`ISynchronization` is the interface that manages device synchronization. A device registers it internally via `setSynchronization(const SynchronizationPtr&)`. Instances are created with the `Synchronization(const TypeManagerPtr&)` factory method.

By default, a `Synchronization` object comes with a single `ClockSyncInterface` acting as the source, in `Input` mode. This means the device is not synchronized to any external time provider and simply uses its internal clock.

## Adding sync interfaces

A device module developer can register additional sync interfaces by casting the sync object to `ISynchronizationInternal` and calling `addInterface(ISyncInterface)`. This is only allowed before `setSynchronization(const SynchronizationPtr&)` is called — any attempt afterward returns an error. In other words, sync interfaces can only be configured during device initialization. Once initialization is done, the only supported operations are:

- set/get the source sync interface
- get the list of available sync interfaces
- get the list of reference domain IDs

The object created by the standard factory can also be cast to `IPropertyObject`, so clients can control `Synchronization` through the property object API as well.

## Source switching behavior

- Only one sync interface can be the source at a time. All other interfaces are either disabled or run in `Output` mode.
- When the source interface changes:
  - The previous source is notified that it is no longer the source, and is expected to switch to `Off` mode.
  - The new source switches to `Auto` mode, or to `Input` mode if `Auto` isn't supported.
  - If the selected interface supports neither mode, `Synchronization` returns an error and reverts the change.
  - Every other interface is also notified of the new source, so interfaces capable of tracking source accuracy can react accordingly (e.g. updating `clockClass` for ptp4l).

## ISyncInterface

`ISyncInterface` manipulates and tracks the state of a single synchronization interface. Each instance has a fully custom configuration; the only required parameter is `Mode`, whose available options depend on whether the interface is currently the source.

Each `ISyncInterface` also exposes two predefined statuses:

- **`SyncSourceStatus`** — the interface's synchronization state: `Off = 0`, `Listening`, `Calibrating`, `Synced`, `Error`, `Unknown`.
- **`SyncRoleStatus`** — the interface's actual role: `Off = 0`, `Input`, `Output`, `Unknown`. This is particularly useful when the interface is in `Auto` mode, since it reports the role that was actually assigned.

## PTP sync interface template

This PR also adds a template for a PTP sync interface, which pre-populates the configuration with PTP-related settings:

```cpp
// PTP configuration properties
constexpr const char* PtpConfigMode = "Mode";
constexpr const char* PtpConfigProfileOptions = "ProfileOptions";
constexpr const char* PtpConfigTransportProtocolOptions = "TransportProtocolOptions";
constexpr const char* PtpConfigProfile = "Profile";
constexpr const char* PtpConfigTwoStepFlag = "TwoStepFlag";
constexpr const char* PtpConfigDomainNumber = "DomainNumber";
constexpr const char* PtpConfigUtcOffset = "UtcOffset";
constexpr const char* PtpConfigPriority1 = "Priority1";
constexpr const char* PtpConfigPriority2 = "Priority2";
constexpr const char* PtpConfigTransportProtocol = "TransportProtocol";

// Port configuration properties
constexpr const char* PortConfiguration = "PortConfiguration";
constexpr const char* PortConfigModeOptions = "ModeOptions";
constexpr const char* PortConfigMode = "Mode";
constexpr const char* PortConfigDelayMechanismOptions = "DelayMechanismOptions";
constexpr const char* PortConfigDelayMechanism = "DelayMechanism";
constexpr const char* PortConfigLogSyncInterval = "LogSyncInterval";
```

This is not meant to be a final, fixed schema — other developers can derive from it and add further configuration settings. The list above is the minimum set expected to be implemented by devices that support PTP configuration.

`PtpSyncInterfaceBaseImpl` provides the internal logic and all the methods needed to track these settings, including the logic for becoming the source (`AsSource`) and for `setMode`.

`Profile` is a fully custom setting — `PtpSyncInterfaceBaseImpl` currently only defines `None`, but developers can add additional profiles (e.g. a vendor-specific default, or `802.1AS`). The same extensibility applies to the other option lists, for example:

```cpp
transportProtocolOptions = List<IString>("IEEE802_3", "UDP_IPV4", "UDP_IPV6");
portDelayMechanismOptions = List<IString>("E2E", "P2P");
modeOptions = Dict<IInteger, IString>({
    {static_cast<Int>(PortSyncMode::Off), "Off"},
    {static_cast<Int>(PortSyncMode::Output), "Output"},
    {static_cast<Int>(PortSyncMode::Auto), "Auto"}
});
```

These defaults can be reduced or expanded by derived implementations, using:

- `PtpSyncInterfaceBaseImpl::setProfileOptions(const ListPtr<IString>& options)`
- `PtpSyncInterfaceBaseImpl::setTransportProtocolOptions(const ListPtr<IString>& options)`
- `PtpSyncInterfaceBaseImpl::setPortDelayMechanismOptions(const ListPtr<IString>& options)`

Status reporting is also expanded with a per-port role status.
</content>

# Usage example

In C++ use:

```cpp
// add the new Synchronization in the device 
void initSynchronization()
{
     auto sync = Synchronization(this->context.getTypeManager());
     auto syncInternal = sync.asPtr<ISynchronizationInternal>(true);

     ntpSyncInterface = createWithImplementation<ISyncInterface, NtpSyncInterface>(context);
     ptpSyncInterface = createWithImplementation<ISyncInterface, PtpSyncInterface>(context);

     syncInternal.addInterface(ntpSyncInterface);
     syncInternal.addInterface(ptpSyncInterface);
     
     setSynchronization(sync);
}

// subscribe on reference domain id changings
void subscribeOnReferenceDomainIdsChangings()
{
        referenceDomainIds = List<IString>();
        referenceDomainIds.pushBack(ntpSyncInterface.getReferenceDomainId());
        ntpSyncInterface.asPtr<IPropertyObject>(true).getOnPropertyValueWrite("Status.ReferenceDomainId") +=  [this](PropertyObjectPtr&, PropertyValueEventArgsPtr& args)
        {
            auto lock = this->getRecursiveConfigLock2();
            referenceDomainIds.setItemAt(0, args.getValue());
            // updateReferenceDomainIds();
        };

        referenceDomainIds.pushBack(ptpSyncInterface.getReferenceDomainId());
        ptpSyncInterface.asPtr<IPropertyObject>(true).getOnPropertyValueWrite("Status.ReferenceDomainId") +=  [this](PropertyObjectPtr&, PropertyValueEventArgsPtr& args)
        {
            auto lock = this->getRecursiveConfigLock2();
            referenceDomainIds.setItemAt(1, args.getValue());
            // updateReferenceDomainIds();
        };
}

```

# API changes

```diff
+ [interface] ISyncInterface : public IBaseObject
+ [function] ISyncInterface::getId(IString** id)
+ [function] ISyncInterface::getSyncType(IString** syncType)
+ [function] ISyncInterface::getReferenceDomainId(IString** referenceDomainId)
+ [function] ISyncInterface::setMode(SyncMode mode)
+ [function] ISyncInterface::getMode(SyncMode* sourceMode)
+ [function] ISyncInterface::getAvailableModes(IDict** availableModes)
+ [function] ISyncInterface::canBeSource(Bool* canBeSource)
+ [function] ISyncInterface::getConfiguration(IPropertyObject** configuration)
+ [function] ISyncInterface::getStatusContainer(IComponentStatusContainer** syncStatus)

+ [interface] ISyncInterfaceInternal : public IBaseObject
+ [function] ISyncInterfaceInternal::setAsSource(Bool source)
+ [function] ISyncInterfaceInternal::sourceChanged(ISyncInterface* source)

+ [interface] ISynchronization : public IBaseObject
+ [function] ISynchronization::getSyncInterfaces(IDict** interfaces)
+ [function] ISynchronization::getAvailableSyncSources(IDict** sources)
+ [function] ISynchronization::setSource(IString* sourceName)
+ [function] ISynchronization::getSource(ISyncInterface** source)
+ [function] ISynchronization::getReferenceDomainIds(IList** ids)
+ [factory] SynchronizationPtr Synchronization(const TypeManagerPtr& manager)

+ [interface] ISynchronizationPrivate : public IBaseObject
+ [function] ISynchronizationPrivate::addInterface(ISyncInterface* syncInterface)

+ [function] IDevice::getSynchronization(ISynchronization** sync)

+ [function] IReferenceDomainInfo::getReferenceDomainIds(IList** referenceDomainIds)
+ [function] IReferenceDomainInfoBuilder::setReferenceDomainIds(IList* referenceDomainIds)
+ [function] IReferenceDomainInfoBuilder::getReferenceDomainIds(IList** referenceDomainIds)

-[function] IReferenceDomainInfo::getUsesOffset(UsesOffset * usesOffset)
-[function] IReferenceDomainInfoBuilder::setUsesOffset(UsesOffset usesOffset)
-[function] IReferenceDomainInfoBuilder::getUsesOffset(UsesOffset * usesOffset)

-[function] IDevice::getSyncComponent(ISyncComponent** sync)
+[deprecatedFunction] IDevice::getSyncComponent(ISyncComponent** sync)
```

ISyncComponent is deprecated. Use the new API instead 

# Required module changes

None


core/coretypes/include/coretypes/ctutils.h
```diff
@@ -276,19 +276,6 @@ inline void checkErrorGuard(IErrorGuard* errorGuard)
...
-void setErrorInfoWithSource(IBaseObject* source, const std::string& message, Params... params)
-{
-    IErrorInfo* errorInfo = nullptr;
-    const ErrCode err = createErrorInfoObjectWithSource(&errorInfo, source, message, std::forward<Params>(params)...);
-    if (OPENDAQ_FAILED(err))
-        return;
-    
-    daqSetErrorInfo(errorInfo);
-    errorInfo->releaseRef();
-}
```
JakaMohorko, line 290: This should be in a separate PR for the sake of transparency.

core/opendaq/synchronization/include/opendaq/synchronization.h
```diff
@@ -0,0 +1,80 @@
...
+    virtual ErrCode INTERFACE_FUNC getSyncInterfaces(IDict** interfaces) = 0;
+
+    /*!
+     * @brief Gets the registered synchronization interfaces that can currently be selected as the
+     * synchronization source.
+     * @param[out] sources A dictionary mapping interface names to the sync interfaces that can be
+     * selected as the synchronization source.
+     */
+    // [templateType(sources, IString, ISyncInterface)]
+    virtual ErrCode INTERFACE_FUNC getAvailableSyncSources(IDict** sources) = 0;
```
JakaMohorko: Would prefer just `getAvailableSources`.

core/opendaq/synchronization/include/opendaq/synchronization.h
```diff
@@ -0,0 +1,80 @@
...
+ * @brief Interface representing the Synchronization of a device in a Test & Measurement system.
+ */
+DECLARE_OPENDAQ_INTERFACE(ISynchronization, IBaseObject)
+{
+    /*!
+     * @brief Gets all synchronization interfaces registered with this synchronization.
+     * @param[out] interfaces A dictionary mapping interface names to the sync interfaces themselves.
+     */
+    // [templateType(interfaces, IString, ISyncInterface)]
+    virtual ErrCode INTERFACE_FUNC getSyncInterfaces(IDict** interfaces) = 0;
```
JakaMohorko: Would `getInterfaces` be sufficient here?

core/opendaq/synchronization/include/opendaq/sync_component.h
```diff
@@ -87,7 +87,7 @@ DECLARE_OPENDAQ_INTERFACE(ISyncComponent, IComponent)
 OPENDAQ_DECLARE_CLASS_FACTORY_WITH_INTERFACE(
     LIBRARY_FACTORY, SyncComponent, ISyncComponent,
```
JakaMohorko, line 88: We need to include a deprecation warning for this interface.

core/opendaq/signal/include/opendaq/reference_domain_info.h
```diff
@@ -102,6 +103,18 @@ DECLARE_OPENDAQ_INTERFACE(IReferenceDomainInfo, IBaseObject)
      * If False, a device will contain time jumps due to resync in the domain signal data.
      */
     virtual ErrCode INTERFACE_FUNC getUsesOffset(UsesOffset * usesOffset) = 0;
```
JakaMohorko: This should be removed.

core/opendaq/synchronization/src/synchronization_impl.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+    try
+    {
+        newSource.asPtr<ISyncInterfaceInternal>(true).setAsSource(True);
+        source = newSource.detach();
+    }
+    catch(...)
+    {
+        oldSource.asPtr<ISyncInterfaceInternal>(true).setAsSource(True);
+        oldSource.asPtr<IPropertyObject>(true).setPropertyValue("Mode", static_cast<Int>(oldSourceMode));
```
JakaMohorko, line 79: Mode is now in `Configuration.Mode`

core/opendaq/synchronization/tests/test_sync_component.cpp
```diff
@@ -169,5 +173,623 @@ TEST_F(SyncComponentTest, Serialization)
...
+    // The old source should have been demoted: Mode back to Off, and no longer selectable
+    // as an input/output source (ModeOptions switched back to the output-only set)
+    ASSERT_EQ(oldSource.getMode(), SyncMode::Off);
+
+    const auto oldSourceModes = oldSource.getAvailableModes();
+    ASSERT_TRUE(oldSourceModes.hasKey(static_cast<Int>(SyncMode::Off)));
+    ASSERT_FALSE(oldSourceModes.hasKey(static_cast<Int>(SyncMode::Input)));
+}
+
+TEST_F(SynchronizationTest, SetSourceRollsBackOnFailure)
```
JakaMohorko, line 345: This test doesn't check rollback. It fails before the source is changed at all and does not touch the rollback code.

core/opendaq/synchronization/include/opendaq/synchronization_impl.h
```diff
@@ -0,0 +1,176 @@
...
+{
+    OPENDAQ_PARAM_NOT_NULL(ids);
+    auto idList = List<IString>();
+
+    const PropertyObjectPtr interfacesProperty = this->objPtr.getPropertyValue("Interfaces");
+    for (const auto & intefaceProp : interfacesProperty.getAllProperties())
+    {
+        SyncInterfacePtr interface = intefaceProp.getValue();
+        const auto referenceDomainID = interface.getReferenceDomainId();
+        if (referenceDomainID.assigned() && referenceDomainID.getLength() != 0)
```
JakaMohorko, line 121: Should we check whether the interface is off here? Otherwise the requirement is that disabled interfaces have an empty ID.

core/opendaq/signal/include/opendaq/reference_domain_info.h
JakaMohorko, line 75: I think these comments might be slightly out-of-date given the new spec. We probably want to re-check for consistency.

core/opendaq/synchronization/src/sync_interface_base_impl.cpp
```diff
@@ -0,0 +1,202 @@
...
+    OPENDAQ_PARAM_NOT_NULL(id);
+    *id = this->id.addRefAndReturn();
+    return OPENDAQ_SUCCESS;
+}
+
+ErrCode SyncInterfaceBaseImpl::getSyncType(IString** syncType)
+{
+    OPENDAQ_PARAM_NOT_NULL(syncType);
+    *syncType = String("").detach();
+    return OPENDAQ_SUCCESS;
```
JakaMohorko: I think this should always be overridden by sync interfaces - I'd return a failing err code by default.

Alternatively, we could automate this:
- The `referenceDomainId` must always start with the `SyncType` -> all lowercase.
- The sync types "ptp", "irig", "ntp", "gps", "local" are standardized in the spec/core.
- `getSyncType` parses the reference domain ID up to the first colon.

core/opendaq/synchronization/include/opendaq/sync_interface.h
```diff
@@ -0,0 +1,129 @@
...
+     * @brief Gets the reference domain ID of the synchronization interface.
+     * @param[out] referenceDomainId The reference domain ID string.
+     */
+    virtual ErrCode INTERFACE_FUNC getReferenceDomainId(IString** referenceDomainId) = 0;
+
+    /*!
+     * @brief Gets whether the synchronization interface can be selected as the synchronization source.
+     * @param[out] canBeSource True if the interface can act as a synchronization source; False otherwise.
+     */
+    virtual ErrCode INTERFACE_FUNC canBeSource(Bool* canBeSource) = 0;
```
JakaMohorko: Our standard is for getters to start with `get`, otherwise RTgen has issues with binding generation (this is a method in Python instead of a property). We can do something like `getCanBeSource` or `getSourceSupported`. Neither are pretty, but they work?

In python it'd be `interface.can_be_source` in the former case.

bindings/c/include/copendaq/synchronization/sync_component.h
```diff
@@ -47,7 +47,7 @@ extern "C"
     daqErrCode EXPORTED daqSyncComponent_getSelectedSource(daqSyncComponent* self, daqInt* selectedSource);
     daqErrCode EXPORTED daqSyncComponent_setSelectedSource(daqSyncComponent* self, daqInt selectedSource);
     daqErrCode EXPORTED daqSyncComponent_getInterfaces(daqSyncComponent* self, daqDict** interfaces);
-    daqErrCode EXPORTED daqSyncComponent_createSyncComponent(daqSyncComponent** obj, daqContext* context, daqComponent* ParseFailedException, daqString* localId);
+    daqErrCode EXPORTED daqSyncComponent_createSyncComponent(daqSyncComponent** obj, daqContext* context, daqComponent* parent, daqString* localId);
```
JakaMohorko, line 50: Commenting here, but C is missing the new interfaces.

core/opendaq/synchronization/src/ptp_sync_interface_impl.cpp
```diff
@@ -0,0 +1,196 @@
+#include <opendaq/ptp_sync_interface_impl.h>
+#include <opendaq/component_status_container_private_ptr.h>
+
+BEGIN_NAMESPACE_OPENDAQ
+
+namespace PtpPropertyNames
+{
+    // Status properties
+    constexpr const char* StatusReferenceDomainId = "ReferenceDomainId";
+    constexpr const char* StatusSynchronized = "Synchronized";
```
JakaMohorko: Unused.

JakaMohorko (changes requested):
Lgtm, just some minor notes in the implementation and comments. This rework is a great step towards clearing up the openDAQ API usage.

PR #1046, opened 2026-01-27
Skip updating locked attributes

anzeskerjanc:
# Brief

Check on update if attribute is locked


JakaMohorko:
@anzeskerjanc Lgtm, just a changelog update would be great to add here.

PR #1048, opened 2026-01-29
Refactor package workflow: apply reusable actions

alexadereyko:
# Brief

Refactor package workflow to use reusable composite actions for build and packaging jobs

# Description

- Apply `build-sdk` action to `package_linux`, `package_linux_wheels`, `package_macos_wheels` jobs
- Apply `package-bindings-python` action for Python wheels generation
- Apply `package-bindings-csharp` action to `package_linux` for NuGet artifacts
- Add `skip-setup-python` and `skip-setup-ninja` inputs to `build-sdk` action for container environments (manylinux)
- Add `upload-artifact` and `package-artifact-name` inputs to `build-sdk` action for direct installer upload
- Add `build-doc` action to build, package and upload Doxygen API reference
- Move Linux dependency lists to `.github/packages/` (`apt.txt`, `apt-i386.txt`, `yum.txt`)
- Rename `build_macos` to `package_macos_wheels`, and add ARM64 variant
- Fix `package_linux` job to use `${{ matrix.runner }}` instead of hardcoded runner
- Installer artifact names now use CPack package name with format `opendaq-VERSION[-SHA]_OS_ARCH_COMPILER`
- Add package-version-detect action for centralized version detection
- Add package_macos_wheels_combined job to merge x86_64/arm64 wheels
- Remove beta- prefix from NuGet VersionSuffix
- Replace Doxygen version shipped with ubuntu-latest runner with 1.9.3 one

# Usage example

Using `build-sdk` action in a container environment with pre-installed Python and Ninja:

> [!NOTE]                                                                                                                                                                                                 
> Some GitHub Actions (e.g. actions/setup-python, seanmiddleditch/gha-setup-ninja) rely on runner system rather than container environment, which may cause issues when running inside containers.

```yaml
container:
  image: quay.io/pypa/manylinux_2_28_x86_64

steps:
  - name: Build SDK
    uses: ./.github/actions/build-sdk
    with:
      additional-dependencies: ninja-build           # install Ninja via system package manager
      cmake-generator: Ninja
      cmake-build-type: Release
      skip-setup-python: true                        # skip python installation (already installed in manylinux container)
      skip-setup-ninja: true                         # skip Ninja installation (installed via additional-dependencies)
```

Using `build-sdk` action with automatic installer upload:

```yaml
steps:
  - name: Build SDK
    uses: ./.github/actions/build-sdk
    with:
      cmake-generator: Ninja
      cmake-build-type: Release
      cpack-generator: DEB                           # required for artifact upload
      upload-artifact: true                          # upload installer as artifact
      package-artifact-name: MyInstaller             # optional: defaults to CPack package name
```

# API changes

N/A

N/A

# Required module changes

N/A

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -54,10 +54,7 @@ jobs:
 
     outputs:
       # trick for reusable workflow in dotnet_bindings because cannot use env context in "with:"
-      # test_path: ''
       windows_x64_artifact_name: ${{ env.windows_x64_artifact }}
-      # short_sha: ${{ env.short_sha }}
-      short_sha_dotnet: ${{ steps.package-csharp-bindings.outputs.short-sha-dotnet }}
 
     env:
       short_sha: 'package'  # for .NET Bindings build
@@ -81,15 +78,14 @@ jobs:
           generate-python-bindings: false
           upload-artifact: true
 
-      - name: Build Python Wheels
-        id: package-python-bindings
+      - name: Package Python Bindings
         if: matrix.generate-python-bindings == true
         uses: ./.github/actions/package-bindings-python
         with:
-          build-path: ${{ steps.build-sdk.outputs.build-path }}
           python-versions: ${{ env.python_versions }}
+          build-path: ${{ steps.build-sdk.outputs.build-path }}
           cmake-build-type: ${{ matrix.cmake-build-type }}
-          package-artifact-name: Wheels (${{ matrix.name }})
+          upload-artifact: true
 
       - name: Build C# Bindings
         id: package-csharp-bindings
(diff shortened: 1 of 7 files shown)


.github/workflows/package.yml
```diff
@@ -121,146 +121,97 @@ jobs:
...
-      - name: Upload API reference
-        if: ${{ matrix.name == 'Ubuntu 22.04 gcc-10 Release x86_64' }}
+      - name: Build Documentation
+        if: matrix.generate-doxygen-doc == true
+        shell: bash
+        run: |
+          cmake -B ${{ steps.build-sdk.outputs.build-path }} -DOPENDAQ_BUILD_DOCUMENTATION=ON
+          cmake --build ${{ steps.build-sdk.outputs.build-path }} --target doc_doxygen
+          cd ${{ steps.build-sdk.outputs.build-path }}/doc_doxygen/html
+          zip -r ../opendaq_cpp_api_reference.zip .
```
JakaMohorko: Same for this, shouldn't this also be part of the reusable workflow?

NikolaiShipilov: I've noticed that documentation archive became 3 Mb smaller after refactoring. So, I've double-checked the generated doc website from acrhive and it appears to not function as intended. My assumption is that building just target `--target doc_doxygen` without rebuilding other targets is somehow not sufficient to generate all required files.

PR #1050, opened 2026-02-02
Refactor property object implementation variable/function scope

JakaMohorko:
# Brief

Refactors property object implementation scoping, moving variables and methods to the private scope where appropriate.

# Next steps

- Remove friend classes from property object by exposing relevant fields in the protected scope
- Move private, static methods to `IPropertyObjectCore`

# Required module changes

If any modules used variables/functions that are now in the private scope, the private member scope should be re-evaluated. Feedback on this change is welcome.

### DIFF
--- a/core/opendaq/signal/include/opendaq/input_port_impl.h
+++ b/core/opendaq/signal/include/opendaq/input_port_impl.h
@@ -856,7 +856,7 @@ ErrCode GenericInputPortImpl<TInterface, Interfaces...>::getPublic(Bool* isPubli
 template <typename TInterface, typename... Interfaces>
 ErrCode GenericInputPortImpl<TInterface, Interfaces...>::setPublic(Bool isPublic)
 {
-    if (this->frozen)
+    if (this->isFrozen())
         return DAQ_MAKE_ERROR_INFO(OPENDAQ_ERR_FROZEN);
 
     {


core/coreobjects/include/coreobjects/property_object_core.h
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+#include <coreobjects/property_object_internal.h>
+
+BEGIN_NAMESPACE_OPENDAQ
+
+DECLARE_OPENDAQ_INTERFACE(IPropertyObjectCore, IBaseObject)
```
denise-opendaq: do we need extra interface for it? I would probably just create a basic class do not add more interfaces classes

JakaMohorko: Creating a basic class will likely not work due to the module structure. Last time I tried we had linker issues when creating a non-header only implementation. 

This approach allows us to move the core logic out of the main property object header (or other included headers).

Notably, this is an internal interface, we do not need to generate any ptrs, or make it public. We are also always able to modify (or remove) this interface.

core/coreobjects/include/coreobjects/property_object_impl.h
```diff
@@ -564,19 +393,19 @@ class GenericPropertyObjectImpl : public ImplementationOfWeak<PropObjInterface,
     // Gets the property value, if stored in local value dictionary (propValues)
     // Parses brackets, if the property is a list
     ErrCode readLocalValue(const StringPtr& name, BaseObjectPtr& value) const;
-    PropertyNameInfo getPropertyNameInfo(const StringPtr& name) const;
+    static PropertyNameInfo getPropertyNameInfo(const StringPtr& name);
 
     // Checks if the value is a container type, or base `IPropertyObject`. Only such values can be set in `setProperty`
-    ErrCode checkContainerType(const PropertyPtr& prop, const BaseObjectPtr& value);
+    static ErrCode checkContainerType(const PropertyPtr& prop, const BaseObjectPtr& value);
```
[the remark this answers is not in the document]
JakaMohorko: They can also be moved to the interface. If we can get it to compile without having a header-only PropertyObjectCore implementation, then I'd remove the interface.

PR #1056, opened 2026-02-06
implement attempting to reconnect with all available addresses

denise-opendaq:
# Brief

Pass available connection strings of the remote device to the native streaming client, enabling reconnection attempts through all available addresses. This improves reconnection reliability when a device loses one interface but remains accessible via others. 

Manually checked:
- Disabling the connection interface in both client and server
- The addressed of active clients are correct
- Primary connection was done with only ipv4, ipv6 or both

# API changes

None

None

# Required module changes

None


NikolaiShipilov (approved):

JakaMohorko (commented):
Upon review, I would suggest that this is reworked for native to create an internal cache of available addresses after device creation.

modules/native_streaming_client_module/src/native_device_impl.cpp
```diff
@@ -303,6 +308,59 @@ void NativeDeviceHelper::setupProtocolClients(const ContextPtr& context)
                                               transportConnectionStatusChangedCb);
 }
 
+void NativeDeviceHelper::setupAlternativeAddressesCallback()
```
JakaMohorko: I think we should just cache all available addresses in `onCompleteServerCapabilities`.

JakaMohorko (changes requested):

modules/native_streaming_client_module/src/native_device_impl.cpp
```diff
@@ -466,6 +467,63 @@ void NativeDeviceImpl::removed()
...
+                    primaryAddressType = primaryAddressTypeCandidate;
+            }
+        }
+    }
+
+    DeviceInfoPtr deviceInfo;
+    OPENDAQ_RETURN_IF_FAILED(getInfo(&deviceInfo));
+
+    if (deviceInfo == nullptr)
+       return OPENDAQ_IGNORED;
```
JakaMohorko: I would just return the `OPENDAQ_SUCCESS` error codes, as the call we're in is actually `setComponentConfig`.

modules/native_streaming_client_module/src/native_device_impl.cpp
```diff
@@ -466,6 +467,63 @@ void NativeDeviceImpl::removed()
...
+            continue;
+
+        if (addressInfo.getReachabilityStatus() == AddressReachabilityStatus::Unreachable)
+            continue;
+
+        alternativeAddresses.pushBack(addressInfo.getAddress());
+    }
+
+    if (!deviceHelper)
+        return OPENDAQ_IGNORED;
```
JakaMohorko: This can probably be checked at the top.

modules/native_streaming_client_module/src/native_device_impl.cpp
```diff
@@ -466,6 +467,63 @@ void NativeDeviceImpl::removed()
...
+                if (primaryAddressTypeCandidate.getLength())
+                    primaryAddressType = primaryAddressTypeCandidate;
+            }
+        }
+    }
+
+    DeviceInfoPtr deviceInfo;
+    OPENDAQ_RETURN_IF_FAILED(getInfo(&deviceInfo));
+
+    if (deviceInfo == nullptr)
```
JakaMohorko: Needs to check for `deviceInfo.assigned()`.

modules/native_streaming_client_module/src/native_streaming_device_impl.cpp
```diff
@@ -41,6 +41,60 @@ NativeStreamingDeviceImpl::NativeStreamingDeviceImpl(const ContextPtr& ctx,
     this->statusContainer.asPtr<IComponentStatusContainerPrivate>().addStatus("ConnectionStatus", connectionStatus);
 }
 
+ErrCode NativeStreamingDeviceImpl::setComponentConfig(IPropertyObject* config)
```
JakaMohorko, line 45: These calls are almost duplicate. Could we have them in a shared helper method? It could simply take as parameters: `config`, `deviceInfo`, `transportProtocolClient`.

JakaMohorko (dismissed):
Looks good in general, some refactoring suggestions.

This is missing test cases, but it likely can not be tested in the GitHub environment. Should be checked manually for edge cases.

shared/libraries/native_streaming_protocol/src/native_streaming_client_handler.cpp
```diff
@@ -276,7 +276,48 @@ void NativeStreamingClientImpl::tryReconnect()
...
+        }
+        catch (const std::exception& e)
+        {
+            LOG_W("Failed to get alternative addresses: {}", e.what());
+        }
+    }
+    else
+    {
+        LOG_D("No alternative addresses callback set for reconnection");
+    }
```
NikolaiShipilov: this section looks like smth from prev implementation based on callbacks.

NikolaiShipilov (commented):
One major thing (I've probalby missed while doing review last time as well) I've just realized - is that there likely isn't a mechanism for providing alternative addresses to established streaming connection & created streaming object (except streaming device use-case) - addresses are available only for device object via device info. Therefore, the streaming stays disconnected even if the configuration protocol successfully switched to alternative address.

NikolaiShipilov (approved):

PR #1067, opened 2026-02-12
Refactor deploy workflow: download package artifact and upload them to S3

alexadereyko:
# Brief

Refactor the `deploy` workflow to publish pre-built artifacts from the `package` workflow instead of building from source.

# Description

The `deploy` workflow trigger is changed from `push` to `workflow_run`, so it runs automatically right after `package` completes. This lets `deploy` download artifacts directly by the triggering run ID:

```yml
on:
  workflow_run:
    workflows: ["Build openDAQ packages"]
    types: [completed]
    branches: [main]

jobs:
  deploy_packages:
    steps:
      - uses: actions/download-artifact@v4
        with:
          github-token: ${{ secrets.GITHUB_TOKEN }}
          run-id: ${{ github.event.workflow_run.id }}
          path: ...
```

For `workflow_dispatch`, the run ID must be provided explicitly. It can be found in the URL of any Actions run page:
```
https://github.com/openDAQ/openDAQ/actions/runs/{run-id}
```
The `run-id` is the numeric identifier (e.g. `21960986807`). Enter it in the input field when triggering the deploy workflow manually.

When a run ID is provided, the commit SHA and branch name are resolved automatically via the GitHub API:

```bash
RUN_ID="${{ inputs.run-id }}"
RUN_INFO=$(gh api repos/${{ github.repository }}/actions/runs/${RUN_ID})
REF=$(echo "$RUN_INFO" | jq -r '.head_sha')
BRANCH=$(echo "$RUN_INFO" | jq -r '.head_branch')
```

The downloaded artifacts are uploaded to AWS S3 under `releases/${BRANCH}/`, where `BRANCH` is resolved either from the `workflow_run` event or from the API call above.

# Usage example

The run ID can also be obtained from the command line. Make sure `gh` is authenticated (or set `GH_TOKEN`), then list successful runs:

```bash
gh run list --workflow package.yml --status success --branch <branch> --limit 5
```

Copy the ID from the output and trigger the deploy:
```bash
gh workflow run deploy.yml --ref <branch> -f run-id=<ID>
```

# API changes

N/A

N/A

# Required module changes

N/A

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -2,7 +2,7 @@ name: Build openDAQ packages
 
 on:
   push:
-    branches: [release/**]
+    branches: [main, release/**]
 
   workflow_dispatch:
 


.github/workflows/deploy.yml
```diff
@@ -911,90 +80,166 @@ jobs:
...
-          || (export DAQ_DEV_WEBSITE=docs-dev; npx antora antora-playbook-dev.yml)
+        run: |
+          case "${{ needs.resolve.outputs.branch }}" in
+            release/*)
+              npx antora antora-playbook.yml
+              ;;
+            *)
+              npx antora antora-playbook-dev.yml
+              ;;
+          esac
```
alexadereyko: @JakaMohorko @NikolaiShipilov 
Previously it checked `github.ref` to decide which playbook to use, but `github.ref` always resolves to the default branch under `workflow_run`, so the condition didn't work correctly. I replaced it with a check on the resolved source branch: `release/*` uses `antora-playbook.yml`, everything else (including `main`) uses `antora-playbook-dev.yml`. Is this the right behavior?

.github/workflows/deploy.yml
```diff
@@ -911,90 +80,166 @@ jobs:
...
+            main)
+              wget ${{ secrets.DEPLOY_DEV_DOCUMENTATION_URL_AND_TOKEN }}
+              ;;
+            release/*)
+              wget ${{ secrets.DEPLOY_DOCUMENTATION_URL_AND_TOKEN }}
+              ;;
+            *)
+              echo "::warning::Branch '$BRANCH' is not a release or dev branch, skipping documentation deploy"
+              ;;
+          esac
```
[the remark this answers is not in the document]
JakaMohorko: That is correct. As the workflow is triggered manually, we can technically get away with not having pattern matching for the release branch.

.github/workflows/deploy.yml
```diff
@@ -1,900 +1,69 @@
...
+    workflows: ["Build openDAQ packages"]
+    types: [completed]
     branches: [main]
 
+  workflow_dispatch:
+    inputs:
+      run-id:
+        description: 'Package workflow run ID to deploy from'
+        required: true
+        type: string
```
JakaMohorko: I did not realize we can have inputs now. Can these be provided from the UI in the repository "actions" tab?

PR #1072, opened 2026-02-16
Reduce unnecessary RPC calls and signal updates by sending values only when the attribute, operation mode, or descriptor has actually changed.

janm-dw:
# Brief

- Send an RPC only if the value of any attribute (active, name, or description) has changed.
- Send an RPC only if the operation mode has changed.
- Update the signal descriptor only when its value has actually changed.

# API changes

none

### DIFF
--- a/shared/libraries/config_protocol/include/config_protocol/config_client_component_impl.h
+++ b/shared/libraries/config_protocol/include/config_protocol/config_client_component_impl.h
@@ -135,8 +135,7 @@ ErrCode ConfigClientComponentBaseImpl<Impl>::setDescription(IString* description
 {
     OPENDAQ_PARAM_NOT_NULL(description);
 
-    CharPtr descriptionStr = nullptr;
-    description->toString(&descriptionStr);
+    auto descriptionStr = StringPtr::Borrow(description);
 
     if (this->description == descriptionStr)
         return OPENDAQ_IGNORED;


tomaz-cvetko:
Checking for equality of descriptors means that Event Packets are not sent when the descriptor would be the same as last one. This caused a problem with a PowerReaderFB test. The FB in question always has the same descriptor (except for range, which doesn't change in this case).

The sequence of events was as follows:
1. voltage & current signals get connected, power FB is configured
2. voltage & current signals send data (for ticks 0-100)
3. voltage signal sets an invalid descriptor (unit="W")
4. voltage & current signals send data => descriptor change causes FB to go into warning state, multireader is set to inactive and data packets are ignored (power FB doesn't send samples for ticks 100-200)
5. voltage signals sets descriptor back to valid
6. voltage & current signals send data (for ticks 200-250) => descriptor change causes FB to recover to Ok state, multireader is set to active and power FB sends out data samples for ticks 200-250.

In step 6:
- BEFORE: if descriptor events can be duplicated, FB sets domain and value descriptors and 2 events must be read on stream reader before data can be read. There is **no gap detected, because event packets restart gap detection**
- AFTER THE CHANGE: setting descriptors doesn't trigger sending event packets, data is sent but **gap is detected (no event packet to restart gap detection)** after that the data can be read.

Additionally, this test doesn't utilize the `SameThread` notification mode and a thread sleep must be used to let the packet processing settle before the reader's status is checked. In my opinion, FB shouldn't be changed because of this (if anything `SameThread` mechanism should be added to it).

NikolaiShipilov (approved):

PR #1147, opened 2026-04-09
Add "Switch" Device update option

JakaMohorko:
# Brief

Adds "Switch" option to Device Update Options

# TODO
- Update changelog
- Add native client tests
- Provide PR description
- Check if Sig->IP connections must be remapped


tomaz-cvetko (approved):
Looks good and the test does check for the core functionality

PR #1148, opened 2026-04-09
Python GUI Demo: property widget fixes, auto-focus on add, and method invocation UX improvements

Maksim3l:
# Brief

Multiple UX and bug fixes for the Python GUI demo.

# Description

- Auto-focus the tree on newly added devices and function blocks using the return value of `self.context.add_device / self.parent_component.add_function_block`, with `self.tree.see()` to scroll into view on large trees
- Replace the `widget.pack_forget()` with `widget.destroy()` which significantly impacted the GUI resources.
- Removed the recursive drawout of parent and child items in `block_view.py`
     - Included refactoring the `block_view` to nolonger have a full scroll but two seperate columns that can be scrolled though
     - Refactored the handeling of the right column to be more freeform and able to display even with multiple banner items in it (example output signal, input signal and recoder). 
- Hide I/O signal ports and adjust row/column sizing to expand and fill the freed space properly
     - Always show output signals regardless of I/O port visibility,
     - Input now sorts suggestions based on unique global ids not names to prevent overriding same-named signals
- Fix Selection-type property widget to be non-editable (was accidentally made writable by focus management; `selection-type` was already intended to be read-only)
- Fix geometry bug in edit_container_property where DPI scale factor was not cast to int, causing container crashes and making ctList properties uneditable
- Add `EditContainerPropertyDialog` support for ctList properties including empty ones
- Make func/proc property names clickable to invoke them directly; display return value in the Value column; skip the popup dialog entirely when no input arguments are required and always give feedback to the user.
- Implemented an `Add Server` button which allows the user to add serverstreaming to their instance
- Fixed the cipping of checkboxes, comboboxes and buttons by hiding them based on their location
- Fixed visual glitch where the view button of string-type output signals got pushed out of view
- Added a domain view to `ISignals`

slika

slika

slika

slika

# Usage example

In terminal use:

```
python -m opendaq
```

# API changes

None

None

# Required module changes

None

### DIFF
--- a/examples/applications/python/GUI Application/gui_demo/components/add_server_dialog.py
+++ b/examples/applications/python/GUI Application/gui_demo/components/add_server_dialog.py
@@ -78,21 +78,17 @@ class AddServerDialog(Dialog):
 
             available_server_types = self.context.instance.available_server_types
             for server_type_id in available_server_types:
-                self.server_tree.insert('', tk.END, iid=server_type_id, values=(
-                    server_type_id, 
-                    daq.IServerType.cast_from(available_server_types[server_type_id]).name))
+                self.server_tree.insert('', tk.END, iid=server_type_id, 
+                                        values=(server_type_id, daq.IServerType.cast_from(available_server_types[server_type_id]).name))
 
     def handle_server_type_selected(self, e=None):
         can_config = False
         selected = utils.treeview_get_first_selection(self.server_tree)
+
         if selected:
-            try:
-                server_id = self.server_tree.item(selected)['values'][0]
-                server_type = self.context.instance.available_server_types[server_id]
-                cfg = daq.IComponentType.cast_from(server_type).create_default_config()
-                can_config = len(cfg.all_properties) > 0
-            except Exception:
-                pass
+            server_id = self.server_tree.item(selected)['values'][0]
+            can_config = self._is_server_configurable(server_id)
+
         self._server_config_btn.configure(state=tk.NORMAL if can_config else tk.DISABLED)
 
     def handle_right_click(self, event):
@@ -101,13 +97,8 @@ class AddServerDialog(Dialog):
         can_config = False
         selected = utils.treeview_get_first_selection(self.server_tree)
         if selected:
-            try:
-                server_id = self.server_tree.item(selected)['values'][0]
-                server_type = self.context.instance.available_server_types[server_id]
-                cfg = daq.IComponentType.cast_from(server_type).create_default_config()
-                can_config = len(cfg.all_properties) > 0
-            except Exception:
-                pass
+            server_id = self.server_tree.item(selected)['values'][0]
+            can_config = self._is_server_configurable(server_id)
 
         menu = tk.Menu(self, tearoff=0)
         menu.add_command(label='Add', command=lambda: self.handle_button(False))
@@ -115,6 +106,20 @@ class AddServerDialog(Dialog):
                          command=lambda: self.handle_button(True),
                          state=tk.NORMAL if can_config else tk.DISABLED)
         menu.tk_popup(event.x_root, event.y_root)
+        
+    def _is_server_configurable(self, server_id):
+        available_types = self.context.instance.available_server_types
+        if server_id not in available_types:
+            return False
+
+        server_type = available_types[server_id]
+
+        if daq.IComponentType.can_cast_from(server_type):
+            comp_type = daq.IComponentType.cast_from(server_type)
+            config = comp_type.create_default_config()
+            return len(config.all_properties) > 0
+
+        return False
 
     def handle_button(self, config: bool):
         parent_top = self.parent.winfo_toplevel()
@@ -210,8 +215,7 @@ class AddServerDialog(Dialog):
             except Exception as e:
                 # Server was added successfully but discovery failed -- warn, don't undo the add
                 parent_top.configure(cursor='')
-                utils.show_error('Warning',
-                                 f'Server added but enable_discovery failed: {str(e)}', self)
+                utils.show_error('Warning', f'Server added but enable_discovery failed: {str(e)}', self)
 
         parent_top.configure(cursor='')
         self.event_port.emit()
--- a/examples/applications/python/GUI Application/gui_demo/components/add_server_dialog.py
+++ b/examples/applications/python/GUI Application/gui_demo/components/add_server_dialog.py
@@ -1,0 +1,221 @@
+import tkinter as tk
+from tkinter import ttk
+
+import opendaq as daq
+from .properties_view import PropertiesView
+
+from .. import utils
+from ..app_context import AppContext
+from ..event_port import EventPort
+from .dialog import Dialog
+
+
+class AddServerDialog(Dialog):
+    def __init__(self, parent, context, selected_component=None, **kwargs):
+        Dialog.__init__(self, parent, 'Add server', context, **kwargs)
+        # send events to parent window
+        self.event_port = EventPort(self.parent)
+        self.parent_component = selected_component
+
+        self.geometry('{}x{}'.format(
+            int(900 * self.context.ui_scaling_factor * self.context.dpi_factor),
+            int(400 * self.context.ui_scaling_factor * self.context.dpi_factor)))
+
+        # child
+
+        tree_frame = ttk.Frame(self)
+        tree = ttk.Treeview(tree_frame, columns=('id', 'name'), displaycolumns=(
+            'id', 'name'), show='tree headings', selectmode=tk.BROWSE)
+        scroll_bar = ttk.Scrollbar(
+            tree_frame, orient=tk.VERTICAL, command=tree.yview)
+        tree.configure(yscrollcommand=scroll_bar.set)
+        scroll_bar.pack(side=tk.RIGHT, fill=tk.Y)
+
+        # define headings
+        tree.heading('id', text='TypeId', anchor=tk.W)
+        tree.heading('name', text='Name', anchor=tk.W)
+
+        tree.column('#0', width=0, stretch=tk.NO)
+        dpi = self.context.dpi_factor
+        tree.column('id', anchor=tk.W, minwidth=int(200 * dpi), width=int(300 *
+                    self.context.ui_scaling_factor * dpi), stretch=tk.NO)
+        tree.column('name', anchor=tk.W, minwidth=int(200 * dpi), width=int(300 *
+                    self.context.ui_scaling_factor * dpi))
+
+        # bind double-click and right-click
+        tree.bind('<Double-1>', self.handle_server_tree_double_click)
+        tree.bind('<Button-3>', self.handle_right_click)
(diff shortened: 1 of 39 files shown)


examples/applications/python/GUI Application/gui_demo/components/block_view.py
```diff
@@ -350,19 +356,19 @@ def on_expand(self):
             self.expanded_frame.grid_columnconfigure(
                 self.cols, weight=1, minsize=int(200 * self.context.dpi_factor), uniform='column')
             self.expanded_frame.grid_rowconfigure(self.rows, weight=1,
-                                                  minsize=int(300 * self.context.dpi_factor) if self.input_ports and self.output_signals
+                                                  minsize=int(600 * self.context.dpi_factor) if self.input_ports and self.output_signals
                                                                  or daq.IFolder.can_cast_from(self.node) and
                                                                  not daq.IDevice.can_cast_from(self.node) else int(600 * self.context.dpi_factor))
```
tomaz-cvetko: Let's write this inline if else to a local variable and then use it in the `minsize=int(var_name * self.context.dpi_factor)`.

examples/applications/python/GUI Application/gui_demo/components/generic_properties_treeview.py
```diff
@@ -500,18 +502,47 @@ def _on_mousewheel(e):
...
+
+            if has_args:
+                if _prop.value_type == daq.CoreType.ctFunc:
+                    dialog = FunctionDialog(self, _prop, daq.IFunction.cast_from(_prop.value), self.context)
+                    dialog.show()
+                    result = dialog.result
+                elif _prop.value_type == daq.CoreType.ctProc:
+                    dialog = FunctionDialog(self, _prop, daq.IProcedure.cast_from(_prop.value), self.context)
+                    dialog.show()
+                    result = dialog.result
```
tomaz-cvetko: This is duplication where the only difference is the class you are casting to `IFunction` vs `IProcedure` which python should be able to handle more cleanly, for example:
```python
def typed_cast(x, cls):
  return cls.cast_from(x)
method_class = daq.IFunction if _prop.value_type == daq.CoreType.ctFunc else daq.IProcedure
method = typed_cacst(_prop.value, method_class)
if has_args:
  dialog = FunctionDialog(self, _prop, method, self.context)
  # ...
else:
  try:
    method()
    # ...
```
Would this make the logic flow more readable?

examples/applications/python/GUI Application/gui_demo/components/generic_properties_treeview.py
```diff
@@ -662,12 +695,49 @@ def edit_value(self):
...
+        
+            if result is True or result is False:
+                result_str = 'OK' if result else 'Fail'
+            elif isinstance(result, Exception):
+                result_str = str(result)
+            else:
+                result_str = str(result)
+            self._last_method_results[prop.name] = result_str
+            if self.exists(selected_item_id):
+                self.set(selected_item_id, 'value', result_str)
```
tomaz-cvetko: Similar to above, a large chunk of code is duplicated but only one of them ever executes. Can we make it a single block without it being too inconvenient?

examples/applications/python/GUI Application/gui_demo/components/add_server_dialog.py
```diff
@@ -0,0 +1,221 @@
...
+        self.update_server_types()
+
+    def update_server_types(self):
+            self.server_tree.delete(*self.server_tree.get_children())
+
+            available_server_types = self.context.instance.available_server_types
+            for server_type_id in available_server_types:
+                self.server_tree.insert('', tk.END, iid=server_type_id, values=(
+                    server_type_id, 
+                    daq.IServerType.cast_from(available_server_types[server_type_id]).name))
```
tomaz-cvetko: Double indent - it works python wise but would be better to stay at standard one indent level

examples/applications/python/GUI Application/gui_demo/components/add_server_dialog.py
```diff
@@ -0,0 +1,221 @@
...
+        can_config = False
+        selected = utils.treeview_get_first_selection(self.server_tree)
+        if selected:
+            try:
+                server_id = self.server_tree.item(selected)['values'][0]
+                server_type = self.context.instance.available_server_types[server_id]
+                cfg = daq.IComponentType.cast_from(server_type).create_default_config()
+                can_config = len(cfg.all_properties) > 0
+            except Exception:
+                pass
```
tomaz-cvetko: What exactly are we try/excepting here? Because when we except anything, but don't actually handle it in a specific way, this only masks things not working properly and makes fixing things harder down the line due to not having exceptions thrown where something went wrong.

If cast fails -> use the `can_cast_from`

examples/applications/python/GUI Application/gui_demo/components/add_server_dialog.py
```diff
@@ -0,0 +1,221 @@
...
```
tomaz-cvetko: `can_cast_from` if you don't know the cast will succeed in advance

tomaz-cvetko (commented):
In general looks good, there are some improvements to be made that would benefit us in the long run

tomaz-cvetko (commented):
I tested the GUI and it looks good, all UI elements are visible, buttons work, no crashes. I think the code is ok now. The only suggestion I have is for the `input_port_row_view.py` to have combobox that can be clicked anywhere to trigger dropdown and not just the arrow. The combobox can be marked as `readonly` to disable typing in it (from add device with config: `cbox = ttk.Combobox(sub_frame, state="readonly", textvariable=var)`). Additionally, as an improvement, a button to disconnect instead of selecting none in the dropdown would be more intuitive. Both of these could be done as a separate non-urgent improvements at a later point.

examples/applications/python/GUI Application/gui_demo/components/block_view.py
```diff
@@ -184,6 +276,86 @@ def select():
...
+                self._make_banner(self.right_stack, 'Domain signal')
+                if self.node.domain_signal is not None:
+                    try:
+                        domain_row = OutputSignalRow(
+                            self.right_stack, self.node.domain_signal, self.context)
+                        domain_row.pack(fill=tk.X)
+                        self._signal_rows.append(domain_row)
+                    except RuntimeError:
+                        ttk.Label(self.right_stack, text='N/A',
+                                  padding=(10, 0)).pack(anchor=tk.W)
```
JakaMohorko: How come this try/catch is required here, but not for the `SignalInfo` row?

examples/applications/python/GUI Application/gui_demo/components/block_view.py
```diff
@@ -184,6 +276,86 @@ def select():
...
+
+                # Related Signals section
+                self._make_banner(self.right_stack, 'Related signals')
+                try:
+                    related = self.node.related_signals
+                    if related and len(related) > 0:
+                        for sig in related:
+                            row = OutputSignalRow(self.right_stack, sig, self.context)
+                            row.pack(fill=tk.X)
+                            self._signal_rows.append(row)
```
JakaMohorko: Same as above.

examples/applications/python/GUI Application/gui_demo/components/block_view.py
```diff
@@ -184,6 +276,86 @@ def select():
...
+                        if hasattr(self, '_signal_row') and self._signal_row.winfo_exists():
+                            self._signal_row.refresh()
+                            
+                        for signal_row in self._signal_rows:
+                            try:
+                                if signal_row.winfo_exists():
+                                    signal_row.refresh()
+                            except Exception:
+                                pass
+                    self._signal_refresh_job = self.after(200, _poll_signal_info)
```
JakaMohorko: Could this be a utility method? It's used in the output_signals_view as well.

examples/applications/python/GUI Application/gui_demo/components/block_view.py
```diff
@@ -184,6 +276,86 @@ def select():
...
+                self._signal_row = OutputSignalRow(
+                    self.right_stack, 
+                    output_signal=self.node, 
+                    context=getattr(self, 'context', None), 
+                )
+                self._signal_row.pack(fill=tk.X)
+
+                self._signal_rows = []
+
+                # Domain Signal
```
JakaMohorko: The Domain Signal, Related Signals, and Signal Info could all use the OutputSignalsView instead that would be extended to handle `ISignal`. That would keep the logic in one location.

examples/applications/python/GUI Application/gui_demo/components/block_view.py
```diff
@@ -395,3 +588,13 @@ def handle_active_toggle(self):
...
+    
+    def _make_banner(self, parent, text):
+        _banner_bg = '#afafaf'
+        _banner_fg = 'white'
+        bar = tk.Frame(parent, bg=_banner_bg, bd=0, highlightthickness=0)
+        bar.pack(fill=tk.X, pady=(0, 8))
+        tk.Label(bar, text=text, bg=_banner_bg, fg=_banner_fg,
+                 font=('TkDefaultFont', 10, 'bold')).pack(
+            side=tk.LEFT, padx=6, pady=2)
+        return bar
```
JakaMohorko: This could be in the utils.py, and reused for input ports and output signal views.

examples/applications/python/GUI Application/gui_demo/utils.py
```diff
@@ -186,6 +186,19 @@ def parse_iso_string(date_string: str) -> datetime:
...
+    """Extract the Reference Domain Offset (in ticks) from a descriptor's
+    ReferenceDomainInfo, if present."""
+    try:
+        ref_info = desc.reference_domain_info
+        if ref_info is not None and ref_info.reference_domain_offset is not None:
+            return int(ref_info.reference_domain_offset)
+    except (RuntimeError, AttributeError):
+        # Not every descriptor carries ReferenceDomainInfo -- that is
+        # perfectly normal and just means no offset correction is needed.
+        pass
```
JakaMohorko: This should not require a try/except. The `is not none` check should be sufficient.

examples/applications/python/GUI Application/gui_demo/utils.py
```diff
@@ -200,10 +213,19 @@ def get_last_value_for_signal(output_signal):
...
-                except ValueError as e:
+                except ValueError:
                     origin = parse_iso_string(origin_str)
                 if last_value is not None:
-                    last_value_in_seconds = int(last_value) * desc.tick_resolution.numerator / desc.tick_resolution.denominator
+                    tick_value = int(last_value)
+                    # The Reference Domain Offset must be summed into the
+                    # tick count *before* scaling -- omitting it produces a
+                    # timestamp that is off by the device's sync correction.
+                    offset_ticks = _get_reference_domain_offset(desc)
```
JakaMohorko: The `last_value` method of the signal already adds the domain offset.

JakaMohorko (changes requested):
Looks good. Mostly some refactoring changes requested.

JakaMohorko (approved):

PR #1153, opened 2026-04-13
Enable and disable discovery for openDAQ Server via native

NikolaiShipilov:
# Brief

Adds disabling discovery feature for openDAQ servers (native, OpcUa, LT, ...) and enables discovery enabling & disabling via native protocol

# Description

- Adds `"EnableDiscovery"` and `"DisableDiscovery"` RPCs
- Increases latest native protocol version to 24

# Usage example

In C++ use:

```cpp
// enable / disable discovery locally
auto serverInstance = InstanceBuilder().addDiscoveryServer("mdns").build();
auto nativeServer = serverInstance.addServer("OpenDAQNativeStreaming", nullptr);
nativeServer.enableDiscovery();
nativeServer.disableDiscovery();

// or via native protocol
auto connectedClient = Instance();
auto device = connectedClient.addDevice("daq.nd://127.0.0.1");
auto mirroredServer = device.getServers()[0];
mirroredServer.enableDiscovery();
mirroredServer.disableDiscovery();
```

# API changes

```diff
+ [function] IServer::disableDiscovery()
```

### DIFF
--- a/shared/libraries/config_protocol/include/config_protocol/config_client_daqserver_component_impl.h
+++ b/shared/libraries/config_protocol/include/config_protocol/config_client_daqserver_component_impl.h
@@ -21,12 +21,12 @@
 namespace daq::config_protocol
 {
 
-class ConfigClientServerImpl : public ConfigClientComponentBaseImpl<ServerImpl<IServer, IConfigClientObject>>
+class ConfigClientDaqServerComponentImpl : public ConfigClientComponentBaseImpl<ServerImpl<IServer, IConfigClientObject>>
 {
 public:
     using Super = ConfigClientComponentBaseImpl<ServerImpl<IServer, IConfigClientObject>>;
 
-    ConfigClientServerImpl(const ConfigProtocolClientCommPtr& configProtocolClientComm,
+    ConfigClientDaqServerComponentImpl(const ConfigProtocolClientCommPtr& configProtocolClientComm,
                            const std::string& remoteGlobalId,
                            const StringPtr& id,
                            const DevicePtr& parentDevice,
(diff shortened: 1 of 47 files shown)


shared/libraries/config_protocol/include/config_protocol/config_client_daqserver_component_impl.h
```diff
@@ -21,12 +21,12 @@
 namespace daq::config_protocol
 {
 
-class ConfigClientServerImpl : public ConfigClientComponentBaseImpl<ServerImpl<IServer, IConfigClientObject>>
+class ConfigClientDaqServerComponentImpl : public ConfigClientComponentBaseImpl<ServerImpl<IServer, IConfigClientObject>>
```
JakaMohorko: I realize the naming is a bit awkward here, but the pattern we use here should still work:
- ConfigClientServer
- ConfigServerServer

JakaMohorko (approved):
Looks good, but I think we should stick to the current naming pattern for config client/server objects.

tomaz-cvetko (approved):
Looks good to me

shared/libraries/config_protocol/include/config_protocol/config_client_daqserver_component_impl.h
NikolaiShipilov: What I've tried with the new naming is to make it a bit more clear to avoid misinterpretation in the end - but new naming appears to be just longer and not really clearer. So the old pattern is not any worse - I'll revert renaming changes. 
The question is should we also stick to original pattern in OpcUa impl https://github.com/openDAQ/OpcUaModules/pull/20 ?

PR #1155, opened 2026-04-13
Add GCC-7 CI build using Ubuntu 20.04 container

alexadereyko:
# Brief

Replace GCC-9 CI job with GCC-7 running in an Ubuntu 20.04 container on ARM runner

# Description

- Replace GCC-9 build matrix entry with GCC-7 in Ubuntu 20.04 container
- Add container bootstrap step to install git and ca-certificates
- Install python3-pip via additional-packages for cmake/ninja pip installation
- Install cmake and ninja via pip instead of external GitHub Actions

### DIFF
--- a/.github/actions/build-sdk/action.yml
+++ b/.github/actions/build-sdk/action.yml
@@ -122,9 +122,6 @@ runs:
         python-version: '3.14'
         architecture: ${{ inputs.enable-32bit == 'true' && 'x86' || 'x64' }}
 
-    - name: Setup cmake
-      uses: jwlawson/actions-setup-cmake@v2
-
     - name: Install Windows dependencies
       if: runner.os == 'Windows' && inputs.additional-dependencies != ''
       shell: pwsh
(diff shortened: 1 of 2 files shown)


.github/actions/build-sdk/action.yml
```diff
@@ -122,9 +122,6 @@ runs:
         python-version: '3.14'
         architecture: ${{ inputs.enable-32bit == 'true' && 'x86' || 'x64' }}
 
-    - name: Setup cmake
-      uses: jwlawson/actions-setup-cmake@v2
-
```
NikolaiShipilov, line 127: Does the cmake version installed via this action differs from the one installed via `pip` ? or we just cannot run the action inside the container and therefore now have to use `pip` for installation ?

alexadereyko: I should not remove the actions for `cmake` and `ninja`. They work in container. The assumption that the actions don't work was wrong.

alexadereyko:
Covered by PR #1172

PR #1157, opened 2026-04-14
Feature/default to safe operation

tomaz-cvetko:
# Brief

Change default for `OperationModeType` provided in the base device class to `SafeOperation`. Add `GenericDevice::onGetDefaultOperationMode` protected getter for device implementers to override (defaults to `SafeOperation`). When added to a folder, device will enter its default operation mode.

# Description

- Default set of available operation modes now includes only `OperationModeType::SafeOperation` (can be overriden)
- Add method `GenericDevice::onGetDefaultOperationMode` with default return value `OperationModeType::SafeOperation` (can be overriden)
- `GenericDevice::setAsRoot` now uses the `onGetDefaultOperationMode` method to change operation mode of the root device
- `GenericDevice` now starts in `OperationModeType::Unknown` during constructor. It is up to the derived classes/users to change it to a valid value.
- When added to a folder, device will enter its default operation mode (`updateOperationMode` now uses the return value of `onGetDefaultOperationMode` instead of hardcoded value).

# API changes

None

None

# Required module changes

To restore previous behavior of not overriden functions implement:
```c++
DeviceDerivedClass::onGetDefaultOperationMode() override
{
  return OperationModeType::Operation;
}

DeviceDerivedClass::onGetAvailableOperationModes() override
{
  return {OperationModeType::Operation};
}
```
Or use a custom set of values that describe your device.

### DIFF
--- a/examples/modules/ref_device_module/src/ref_device_impl.cpp
+++ b/examples/modules/ref_device_module/src/ref_device_impl.cpp
@@ -590,11 +590,6 @@ std::set<OperationModeType> RefDeviceImpl::onGetAvailableOperationModes()
     return {OperationModeType::Idle, OperationModeType::Operation, OperationModeType::SafeOperation};
 }
 
-OperationModeType RefDeviceImpl::onGetDefaultOperationMode()
-{
-    return OperationModeType::Operation;
-}
-
 void RefDeviceImpl::onOperationModeChanged(OperationModeType modeType)
 {
     bool active = modeType != OperationModeType::Idle;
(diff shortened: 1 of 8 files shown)


examples/modules/ref_device_module/src/ref_device_impl.cpp
```diff
@@ -590,6 +590,11 @@ std::set<OperationModeType> RefDeviceImpl::onGetAvailableOperationModes()
     return {OperationModeType::Idle, OperationModeType::Operation, OperationModeType::SafeOperation};
 }
 
+OperationModeType RefDeviceImpl::onGetDefaultOperationMode()
+{
+    return OperationModeType::Operation;
```
JakaMohorko: I would change just not add this override, as the device should be in `SafeOperation` by default.

JakaMohorko (approved):

examples/modules/ref_device_module/src/ref_device_impl.cpp
tomaz-cvetko: Removed

PR #1163, opened 2026-04-16
Python GUI: Added a device domain view and removed dialog box overlap

Maksim3l:
# Brief

Added a device domain outout with some ui tweeks

# Description

## Device domain
- is now nolonger collapsable
- shows only the last value in the row, other info is in the attributes dialog
- use signal output view for it
- uses the name of the domain signal

## Property treeview tweeks
- Now mark selection properties with only a single selectable value as read only (they should be grayed out) 
- Now hides the "Unit" column by default in properties view and show it as a suffix of the displayed Value. This is done for all types of properties, including selection.

## Dialog overlap
- the dialog box nolonger overlaps on all other windows on desktop

# Usage example

```sh
py -m opendaq
```

slika


tomaz-cvetko (approved):
Code looks good, I tried the functionality too

JakaMohorko (approved):

PR #1164, opened 2026-04-16
Do not call begin/endUpdate on sub-components

denise-opendaq:
# Brief

Previously, calling `beginUpdate/endUpdate` propagated down to child components. This PR changes the behaviour so that beginUpdate/endUpdate apply only to the component itself and to any nested property objects within it.

### DIFF
--- a/core/opendaq/component/include/opendaq/folder_impl.h
+++ b/core/opendaq/component/include/opendaq/folder_impl.h
@@ -83,8 +83,6 @@ protected:
                                            const BaseObjectPtr& context,
                                            const FunctionPtr& factoryCallback);
 
-    void callBeginUpdateOnChildren() override;
-    void callEndUpdateOnChildren() override;
     void onUpdatableUpdateEnd(const BaseObjectPtr& context) override;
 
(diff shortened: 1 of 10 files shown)


core/opendaq/component/include/opendaq/folder_impl.h
```diff
@@ -513,36 +511,6 @@ void FolderImpl<Intf, Intfs...>::deserializeCustomObjectValues(
     }
 }
 
-template <class Intf, class... Intfs>
-void FolderImpl<Intf, Intfs...>::callBeginUpdateOnChildren()
```
JakaMohorko, line 517: This change has major side effects. `begin/endUpdate` should be a recursive call when called from the `IPropertyObject` API, but it should not recurse down the whole openDAQ tree whenever any individual component calls `beginUpdate` during `IUpdadaple::Update`.

Basically, we need to have an internal, non-recursive option that only calls `beginUpdate` on nested property objects.

JakaMohorko (changes requested):

PR #1166, opened 2026-04-17
Add option to remove devices not in the loaded json configuration

tomaz-cvetko:
# Brief

Add a load configuration strategy setting that enables removal of existing devices that are not present in the loaded configuration

# Description

- Introduce a boolean flag to remove devices not mentioned in the json
- Add a getter and a setter for the flag to the `UpdateParameters`
- Fix a bug in the remapping of devices where you couldn't correctly remap to existing sibling device and remap two devices to each other.

# API changes

```diff
+ ErrCode IUpdateParameters::getRemoveUnusedDevices(Bool* remove);
+ ErrCode IUpdateParameters::setRemoveUnusedDevices(Bool remove);
```

To achieve the same behavior as before this change (do not remove not mentioned devices) use

```cpp
updateParameters.setConfigurationLoadMode(ConfigurationLoadMode::Merge);
```


NikolaiShipilov (approved):
LGTM. Just one failing test (due to changed default mode) to be fixed

core/opendaq/component/include/opendaq/update_parameters.h
```diff
@@ -44,12 +53,24 @@ DECLARE_OPENDAQ_INTERFACE(IUpdateParameters, IPropertyObject)
...
      * @brief Sets the device update options object that allows for specifying how a device and its subdevices are to be updated.
      * @param options The device update options object.
      */
     virtual ErrCode INTERFACE_FUNC setDeviceUpdateOptions(IDeviceUpdateOptions* options) = 0;
+
+    /*!
+     * @brief Gets the strategy with which the configuration will be loaded.
+     * @param mode The configuration load mode.
+     */
+    virtual ErrCode INTERFACE_FUNC getConfigurationLoadMode(ConfigurationLoadMode* mode) = 0;
```
JakaMohorko: This feels like too broad of a name for such a setting. Do we define loading or not loading devices as merge vs. exact? There are other things that are loaded that someone might think that fit together here (eg. do we load operation modes in both options, does it impact their loading)?

I'd change the method to be a bit more narrow, eg. `removeOldDevices` or similar.

core/opendaq/utility/tests/test_device_update_options.cpp
```diff
@@ -540,3 +540,71 @@ TEST_F(DeviceUpdateOptionsTest, SerializeDeserialize)
 
     ASSERT_EQ(options, optionsDeserialized);
 }
+
+TEST_F(DeviceUpdateOptionsTest, MergeKeepsExistingDeviceNotInConfiguration)
```
JakaMohorko: Tests for interaction with remapping should be present. Also tests using the native configuration protocol with gateway devices.

JakaMohorko (changes requested):

core/opendaq/component/include/opendaq/update_parameters.h
tomaz-cvetko: I agree that the setting name is broad, especially if its only purpose ever is just to remove devices not mentioned in the json. I thought I was building in some "soft exensibility" by keeping it a bit vague - in my mind, it was likely that someone who wants "Merge" may also want other settings to be kept as they are (and that we could decide to support that at one point). The plan for the future was to keep the "Exact" to always mean "exactly apply the json config" and then maybe have "KeepDeviceTree" or "KeepExistingSettings" and `ConfigurationLoadMode` could be taken into acount elsewhere to define the loading behavior instead of introducing a new boolean flag with the get/set pair for every aspect of the behavior. But that may be a bad thing as well and complicating things in advance for no reason

core/opendaq/device/include/opendaq/device_impl.h
```diff
@@ -2358,11 +2364,37 @@ void GenericDevice<TInterface, Interfaces...>::updateObject(const SerializedObje
...
+                               updateDevice(localId, obj, context);
+                               toRemove.erase(localId);
+                               std::cout << "Don't remove " << localId << std::endl;
+                           });
+
+        Bool removeOldDevices = contextPtr.getUpdateParameters().getRemoveOldDevices();
+        if (removeOldDevices){
+            auto remapping = contextPtr.getInternalState().get("DeviceMapping").asPtrOrNull<IDict, DictPtr<IString, IString>>();
+
+            for (const auto& [oldId, newId] : remapping)
```
denise-opendaq: why remove is after update?

tomaz-cvetko: Before update remapping is not updated yet and we can't use it to get the necessary references.

core/opendaq/component/src/update_parameters_impl.cpp
```diff
@@ -5,6 +5,7 @@ BEGIN_NAMESPACE_OPENDAQ
 
 UpdateParametersImpl::UpdateParametersImpl()
     : Super()
+    , removeOldDevices(False)
 {
     Super::addProperty(BoolProperty("RemoteUpdate", false));
```
denise-opendaq, line 9:
```suggestion
    Super::addProperty(BoolProperty("RemoteUpdate", false));
    Super::addProperty(BoolProperty("RemoveOldDevices", false));
```

core/opendaq/component/src/update_parameters_impl.cpp
```diff
@@ -29,6 +30,20 @@ ErrCode UpdateParametersImpl::setDeviceUpdateOptions(IDeviceUpdateOptions* optio
     return OPENDAQ_SUCCESS;
 }
 
+ErrCode INTERFACE_FUNC UpdateParametersImpl::getRemoveOldDevices(Bool* remove)
+{
+    OPENDAQ_PARAM_NOT_NULL(remove);
+
+    *remove = removeOldDevices;
```
denise-opendaq:
```suggestion
    *remove = getTypedProperty<Boolean>("RemoveOldDevices");
```

core/opendaq/component/src/update_parameters_impl.cpp
```diff
@@ -29,6 +30,20 @@ ErrCode UpdateParametersImpl::setDeviceUpdateOptions(IDeviceUpdateOptions* optio
...
+    return OPENDAQ_SUCCESS;
+}
+
+ErrCode INTERFACE_FUNC UpdateParametersImpl::setRemoveOldDevices(Bool remove)
+{
+    removeOldDevices = remove;
```
denise-opendaq:
```suggestion
    OPENDAQ_RETURN_IF_FAILED(setPropertyValue(String("RemoveOldDevices"), BaseObject(remove));
```

core/opendaq/utility/tests/test_device_update_options.cpp
```diff
@@ -540,3 +540,130 @@ TEST_F(DeviceUpdateOptionsTest, SerializeDeserialize)
...
+{
+    auto serializer = JsonSerializer();
```
denise-opendaq:
```suggestion
    auto str = instance.saveConfiguration();
```

core/opendaq/device/include/opendaq/device_impl.h
```diff
@@ -2356,13 +2442,39 @@ void GenericDevice<TInterface, Interfaces...>::updateObject(const SerializedObje
...
+
+                               // Device is referenced directly (has its own configuration)
+                               devicesWithoutReference.erase(localId);
+
+                               // Another device replaced the one with localId via remapping
+                               if (remapping.hasKey(localId))
+                                   devicesWithoutReference.erase(remapping.get(localId));
+                           });
+
+        Bool removeOldDevices = contextPtr.getUpdateParameters().getRemoveOldDevices();
```
JakaMohorko:
```suggestion
        // Remove devices that not in the setup and are not targets of remapping
        Bool removeOldDevices = contextPtr.getUpdateParameters().getRemoveOldDevices();
```

JakaMohorko (approved):
Lgtm!

PR #1167, opened 2026-04-17
Python GUI: Guarding against empty e arguments and additional gui tweeks

Maksim3l:
# Brief

Add server support, refactor tree columns, improve property clear actions, fix Py callback crashes

# Description

- Refactored `add_server_dialog.py` columns to define all three (`name`, `description`, `id`), populate values in that order, and read the id back from `values[2]` instead of `values[0]`
- Applied the same column refactor to `add_function_block_dialog.py`
- Added `clear_property_values` to the tree right-click context menu, calling the native SDK `clear_property_values()` on the selected node
- Split the properties panel context menu clear action into two: "Clear property value" (singular, for normal properties via SDK `clear_property_value(name)`) and "Clear property values" (plural, for containers via SDK `clear_property_values()`)
- Removed the manual `_clear_object_properties` and `_clear_single_property` helpers, replaced by native SDK calls
- Added server support to the tree: `IServer` added to `find_fb_device_folder`, excluded from the folder-toggle logic in `handle_tree_select`, and given its own right-click context menu with "Enable discovery"
- Added unit suffixes to selection and suggested combobox overlays in the properties panel, with strip-on-save for editable suggested values
- Fixed orphaned `<<DialogReady>>` toplevel bindings that persisted after widget destruction, causing `lambda() missing 1 required positional argument: 'e'` errors when operation modes changed.
- Made event lambda parameters optional (`e=None`) as a defensive measure against Python 3.14 Tkinter callback dispatch calling handlers with zero arguments.
- Added `<Destroy>` cleanup handlers that unbind toplevel-level event subscriptions when `PropertiesTreeview` or `LoadInstanceConfigDialog` instances are destroyed.
- Guarded `_set_node_update_status_recursive` against `None` components to prevent the cascading `'NoneType' object has no attribute 'updating'` error during tree refresh after operation mode changes.

# Usage example

## In terminal use

```sh
py -m opendaq
```

slika

slika

slika

slika

### DIFF
--- a/examples/applications/python/GUI Application/gui_demo/components/add_function_block_dialog.py
+++ b/examples/applications/python/GUI Application/gui_demo/components/add_function_block_dialog.py
@@ -18,7 +18,7 @@ class AddFunctionBlockDialog(Dialog):
         self.parent_component = selected_component
 
         self.geometry('{}x{}'.format(
-            int(900 * self.context.ui_scaling_factor * self.context.dpi_factor),
+            int(1400 * self.context.ui_scaling_factor * self.context.dpi_factor),
             int(400 * self.context.ui_scaling_factor * self.context.dpi_factor)))
 
         # parent
@@ -63,7 +63,7 @@ class AddFunctionBlockDialog(Dialog):
         dpi = self.context.dpi_factor
         tree.column('name', anchor=tk.W, minwidth=int(200 * dpi), width=int(300 *
                     self.context.ui_scaling_factor * dpi), stretch=tk.NO)
-        tree.column('description', anchor=tk.W, minwidth=int(200 * dpi), width=int(300 *
+        tree.column('description', anchor=tk.W, minwidth=int(400 * dpi), width=int(300 *
                     self.context.ui_scaling_factor * dpi))
         tree.column('id', anchor=tk.W, minwidth=int(200 * dpi), width=int(300 *
                     self.context.ui_scaling_factor * dpi))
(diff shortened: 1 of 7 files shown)


examples/applications/python/GUI Application/gui_demo/components/add_server_dialog.py
```diff
@@ -24,23 +24,26 @@ def __init__(self, parent, context, selected_component=None, **kwargs):
...
         tree.column('#0', width=0, stretch=tk.NO)
         dpi = self.context.dpi_factor
-        tree.column('id', anchor=tk.W, minwidth=int(300 * dpi), width=int(400 *
+        tree.column('name', anchor=tk.W, minwidth=int(200 * dpi), width=int(300 *
                     self.context.ui_scaling_factor * dpi), stretch=tk.NO)
-        tree.column('name', anchor=tk.W, minwidth=int(300 * dpi), width=int(400 *
+        tree.column('description', anchor=tk.W, minwidth=int(200 * dpi), width=int(300 *
                     self.context.ui_scaling_factor * dpi))
+        tree.column('id', anchor=tk.W, minwidth=int(200 * dpi), width=int(300 *
+                    self.context.ui_scaling_factor * dpi), stretch=tk.NO)
```
JakaMohorko: Full widget width should be adjusted to accommodate all columns. Same applies for the add fb dialog.

examples/applications/python/GUI Application/gui_demo/components/add_function_block_dialog.py
```diff
@@ -115,14 +118,14 @@ def traverse_devices_recursive(tree, parent_id, component):
...
-                tree.insert(parent_id, tk.END, text=device.name,
+                tree.insert(parent_id, tk.END, text=device.description,
                             iid=device.global_id, open=tk.TRUE)
                 parent_id = device.global_id
 
             if daq.IFunctionBlock.can_cast_from(component):
                 function_block = daq.IFunctionBlock.cast_from(component)
                 if function_block.available_function_block_types:
-                    tree.insert(parent_id, tk.END, text=function_block.name,
+                    tree.insert(parent_id, tk.END, text=function_block.description,
```
JakaMohorko: Typo when replacing name -> description

JakaMohorko (changes requested):

JakaMohorko (approved):

PR #1171, opened 2026-04-21
MDNS discovery ratelimiting

NikolaiShipilov:
# Brief

Introduces a token-bucket-based rate limiting mechanism for openDAQ mDNS server to mitigate excessive incoming query traffic and reduce the impact of abusive or malfunctioning queriers. 
Improves thread safety of connected-client list retrieval by operating on a cloned nested property object instead of the original instance.

# Description

Incoming queries are grouped into fixed-capacity "querier buckets" table-like container.
Each bucket represents a unique querier identified by:
- source IP address
- source port
- address family

Each querier is mapped into a bucket table using a 32-bit xxHash-based fingerprint and a 32-bit-computation-friendly hash-to-index transformation.

Every bucket maintains:
- current token count
- maximum burst capacity
- token refill rate
- last query timestamp

For every incoming query:
1. The querier identity is hashed.
2. The hash is transformed into a bucket table index.
3. If the bucket at the index belongs to the same querier:
    * tokens are refilled according to elapsed time
    * one token is consumed
    * the query is allowed only if tokens are available
4. If the bucket belongs to another querier:
    * stale buckets may be replaced
    * otherwise the querier shares the existing bucket limits

The design allows bucket sharing under hash collisions in order to maintain predictable limited memory usage. Increasing the configured maximum number of querier buckets reduces the probability of hash collisions and therefore improves fairness of rate-limit distribution between independent queriers. 

# Usage example

The mechanism is configurable through the introduced new group of instance config provider parameters under the key `"MdnsDiscoveryServer"`:
- `"EnableDiscoveryRateLimit"` - boolean - enables or disables rate limiting - default value is `true`
- `"SingleQuerierRateLimitPerSecond"` - integer number - per-querier refill rate - tokens per second - default  value is `25`,
- `"MaxActiveQueriers"` - integer (32bit) number - maximum tracked queriers - the number of querier buckets in the table - default value is `150`,
- `"MaxQueryCountPerSecond"` - integer number - maximum burst size - allows short-term spikes of mDNS queries from a querier - default value is `75`.

### DIFF
--- a/core/opendaq/modulemanager/include/opendaq/discovery_server.h
+++ b/core/opendaq/modulemanager/include/opendaq/discovery_server.h
@@ -44,6 +44,7 @@ DECLARE_OPENDAQ_INTERFACE(IDiscoveryServer, IBaseObject)
 
 OPENDAQ_DECLARE_CLASS_FACTORY_WITH_INTERFACE(LIBRARY_FACTORY, 
     MdnsDiscoveryServer, IDiscoveryServer,
-    ILogger*, logger)
+    ILogger*, logger,
+    IDict*, options)
 
 END_NAMESPACE_OPENDAQ
--- a/core/opendaq/modulemanager/include/opendaq/discovery_server.h
+++ b/core/opendaq/modulemanager/include/opendaq/discovery_server.h
@@ -42,8 +42,12 @@ DECLARE_OPENDAQ_INTERFACE(IDiscoveryServer, IBaseObject)
 };
 /*!@}*/
 
-OPENDAQ_DECLARE_CLASS_FACTORY_WITH_INTERFACE(LIBRARY_FACTORY, 
+OPENDAQ_DECLARE_CLASS_FACTORY_WITH_INTERFACE(LIBRARY_FACTORY,
     MdnsDiscoveryServer, IDiscoveryServer,
+    ILogger*, logger)
+
+OPENDAQ_DECLARE_CLASS_FACTORY_WITH_INTERFACE(LIBRARY_FACTORY,
+    MdnsDiscoveryServerWithOptions, IDiscoveryServer,
     ILogger*, logger,
     IDict*, options)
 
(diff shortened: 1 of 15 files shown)


core/opendaq/modulemanager/include/opendaq/discovery_server.h
```diff
@@ -44,6 +44,7 @@ DECLARE_OPENDAQ_INTERFACE(IDiscoveryServer, IBaseObject)
 
 OPENDAQ_DECLARE_CLASS_FACTORY_WITH_INTERFACE(LIBRARY_FACTORY, 
     MdnsDiscoveryServer, IDiscoveryServer,
-    ILogger*, logger)
+    ILogger*, logger,
+    IDict*, options)
```
JakaMohorko: We should not modify the factories at this point. I'd suggest adding an additional one with options and leaving the old one in tact - that one assumes default settings.

shared/libraries/discovery_server/src/mdnsdiscovery_server.cpp
```diff
@@ -135,7 +187,12 @@ std::string MDNSDiscoveryServer::getHostname()
     return hostname;
 }
     
-MDNSDiscoveryServer::MDNSDiscoveryServer(void)
+MDNSDiscoveryServer::MDNSDiscoveryServer(const DictPtr<IString, IBaseObject>& options)
+    : options(options)
+    , perQuerierBucketLimit(options.hasKey("RateLimitPerQuerier") ? static_cast<size_t>(options.get("RateLimitPerQuerier")) : 10)
+    , maxQueriers(options.hasKey("MaxQueriersLimit") ? static_cast<uint32_t>(options.get("MaxQueriersLimit")) : 100)
+    , maxBurst(std::max(options.hasKey("MaxBurst") ? static_cast<size_t>(options.get("MaxBurst")) : 50, perQuerierBucketLimit))
```
JakaMohorko: Not sure if these names can be intuitively understood, the concern is mainly the unit here is not clear - we are using number of queries per second as a unit. Not sure if these are better, would be good to get another opinion:

- SingleQuerierRateLimitPerSecond
- MaxQueryCountPerSecond

For the querier limit, I'd suggest:
- MaxActiveQueriers

JakaMohorko (commented):
Lgtm with some minor comments!

- This feature needs a usage example in Python on how the instance should be modified.
- I'm not sure how to validate that the default rate limiting settings are good. I'd suggest we add a way to turn the rate limiting off in case users prefer the previous behaviour.

denise-opendaq (approved):
апрув

PR #1172, opened 2026-04-21
Switch CI to unified reusable workflow

alexadereyko:
# Brief

Replace the project-local `ci.yml` build matrices with the unified reusable workflow from `openDAQ/openDAQ-CI`, and drive it via `packages` / `cmake-presets` inputs with per-job `cache-variables` overrides.

# Description

- Replace `build_windows` / `build_linux` / `build_macos` jobs in `.github/workflows/ci.yml` with a single ci job that delegates to `openDAQ/openDAQ-CI/.github/workflows/reusable.yml`
- Drive the reusable workflow via `packages` (apt/pip install, Python version per matrix row) and `cmake-presets` with per-job `cache-variables` overrides via `match-jobs` patterns:
  - default — `OPENDAQ_PYTHON_VERSION=3.14` (matches the Python set up by the reusable workflow)
  - `windows-*-msvs-*` — `OPENDAQ_MSVC_SINGLE_PROCESS_BUILD=ON`
  - `macos-*` — `CMAKE_OSX_DEPLOYMENT_TARGET=10.15` + `run-tests-priv`
  - `ubuntu-20.04-*-gcc-7-*` / `*-clang-9-*` — disables SFML renderer and CSV recorder, pins `OPENDAQ_PYTHON_VERSION=3.8`
  - `ubuntu-20.04-x86-*` — additionally `OPENDAQ_FORCE_COMPILE_32BIT=ON`, disables Python/C# bindings, suppresses shift-count-overflow error
  - `*-intel-cc-*` — disables Python/C# bindings for the Intel oneAPI DPC++/C++ (icx) job
- Suppress `-Wunknown-warning-option` globally for Clang/AppleClang < 10 in `CMakeLists.txt` so the legacy-compiler matrix rows compile cleanly with flags introduced in newer Clang versions
- Update supported compilers table in `README.md`


.github/workflows/ci.yml
```diff
@@ -5,329 +5,45 @@ on:
...
-  #           -DANDROID_ABI=${{ matrix.android_abi }} \
-  #           -DANDROID_NDK=${{ env.android_ndk_path }} \
-  #           -DANDROID_PLATFORM=android-${{ matrix.android_version }} \
-  #           -DCI_GIT_BRANCH=${{ github.head_ref }} \
-  #           ..
-  #     - name: Build
-  #       working-directory: ./build
-  #       run: cmake --build .
+    name: openDAQ
+    uses: openDAQ/openDAQ-CI/.github/workflows/reusable.yml@jira/TBBAS-3212-support-opendaq-ci
```
alexadereyko, line 12: Update ref before merging

.github/workflows/unstable_tests.yml
```diff
@@ -6,9 +6,44 @@ on:
   workflow_dispatch:
 
 jobs:
-  run-ci-jobs:
+  ci:
     if: ${{ github.event_name == 'workflow_dispatch' || github.ref == 'refs/heads/main' }}
-    uses: ./.github/workflows/ci.yml
+    name: openDAQ (unstable)
+    uses: openDAQ/openDAQ-CI/.github/workflows/reusable.yml@jira/TBBAS-3212-support-opendaq-ci
```
alexadereyko, line 12: Update `@ref` before merge

tests/integration/test_opendaq_device_modules/test_native_device_modules.cpp
```diff
@@ -3809,19 +3827,21 @@ TEST_F(NativeDeviceModulesTest, SettingOperationModeWithPermissionsNestedDevice)
...
+        ASSERT_NO_THROW(server.setOperationModeRecursive(OMT::Operation));
         std::this_thread::sleep_for(std::chrono::milliseconds(100));
         test_helpers::checkDeviceOperationMode(server, OMT::Operation, true);
         test_helpers::checkDeviceOperationMode(server.getDevices()[0], OMT::Operation, true);
         test_helpers::checkDeviceOperationMode(server.getDevices()[0].getDevices()[0], OMT::Operation, true);
 
-        test_helpers::checkDeviceOperationMode(client, OMT::Operation);
-        test_helpers::checkDeviceOperationMode(client.getDevices()[0], OMT::Operation);
-        test_helpers::checkDeviceOperationMode(client.getDevices()[0].getDevices()[0], OMT::Operation);
-        test_helpers::checkDeviceOperationMode(client.getDevices()[0].getDevices()[0].getDevices()[0], OMT::Operation);
```
NikolaiShipilov: If I'm not mistaken these checks were simply verifying that operation mode applied recursively. I guess this is the import of https://github.com/openDAQ/openDAQ/pull/1180 to the RC branch - @viacheslauK could you please comment on this and provide some context.

.github/workflows/ci.yml
```diff
@@ -5,329 +5,68 @@ on:
...
+    with:
+      timeout: '[{"timeout-minutes": 270}]'
+      include-jobs: >
+        [
```
NikolaiShipilov, line 30: Is it intentional that 32 bit build is not in the list anymore ?

.github/workflows/unstable_tests.yml
NikolaiShipilov, line 1: The file's content looks exactly the same as regular ci workflow except run triggers. It should at least use different presets, I guess. 
Ideally, this file should somehow reuse the regular CI file instead of copying it; so any general purpose change or fix will be required only in a single file without being applied to the same code twice.

.github/workflows/ci.yml
alexadereyko: Get 32 bit linux job back against GCC-7

.github/workflows/unstable_tests.yml
alexadereyko: Rolled back to previous approach when `unstable_tests.ym` calls `ci.yml` passing preset names.

NikolaiShipilov (approved):

JakaMohorko (approved):

.github/workflows/unstable_tests.yml
alexadereyko: irrelevante

.github/workflows/ci.yml
alexadereyko: Updated to `@v2`

tests/integration/test_opendaq_device_modules/test_native_device_modules.cpp
alexadereyko: Commit #1180 came accidentally and excluded from PR

PR #1174, opened 2026-04-22
Support creating package for macos

denise-opendaq:
# Brief

ci generated package for Mac OS (Intel and arm)

### DIFF
--- a/.github/actions/build-sdk/action.yml
+++ b/.github/actions/build-sdk/action.yml
@@ -47,7 +47,7 @@ inputs:
     required: false
     default: 'ci'
   cpack-generator:
-    description: 'CPack generator (NSIS, DEB, RPM, ZIP, TGZ). If set, Build Installer step will run'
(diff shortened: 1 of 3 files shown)


.github/actions/build-sdk/action.yml
```diff
@@ -355,13 +355,20 @@ runs:
           TGZ) PACKAGE_EXT="tar.gz" ;;
           DEB) PACKAGE_EXT="deb" ;;
           RPM) PACKAGE_EXT="rpm" ;;
+          PKG) PACKAGE_EXT="pkg" ;;
```
alexadereyko: Could we just pass `productbuild` directly via `input.cpack-generator` instead of introducing a `PKG` alias? All other accepted values (NSIS, ZIP, TGZ, DEB, RPM) are real CPack generator names, so the alias breaks that convention and requires an extra mapping.

alexadereyko (approved):
Looks good to me

PR #1185, opened 2026-04-29
Fix re-entrant property update crash in endApplyUpdate

denise-opendaq:
# Brief

This PR fixes a re-entrancy issue in endApplyUpdate().
# Description

`endApplyUpdate()` replays buffered property writes stored in `updatingPropsAndValues`. Applying these writes may trigger `onPropertyValueWrite events`. If an event handler starts another `beginUpdate() / endUpdate()` cycle on the same property object, or on a parent object that propagates updates to it, `endApplyUpdate()` can be called recursively.
Previously, the recursive call could clear `updatingPropsAndValues` while the outer call was still iterating over it. This invalidated the outer iteration and could result in a null value being passed to `setPropertyValueInternal`, causing an `ArgumentNullException`.
This PR fixes the issue by moving `updatingPropsAndValues` into a local variable before replaying the buffered writes. As a result, re-entrant calls see an empty update buffer and cannot invalidate the active iteration.

# API changes

None

# Required module changes

None


core/coreobjects/include/coreobjects/property_object_impl.h
```diff
@@ -2748,8 +2748,9 @@ void GenericPropertyObjectImpl<PropObjInterface, Interfaces...>::beginApplyUpdat
 template <typename PropObjInterface, typename... Interfaces>
 void GenericPropertyObjectImpl<PropObjInterface, Interfaces...>::endApplyUpdate()
 {
+    UpdatingActions localUpdates = std::move(updatingPropsAndValues);
```
JakaMohorko, line 2752: What should be the result of a property write during `endApplyUpdate`? In this case, `endUpdate` will need to be called twice to actually apply the changes if anything changed during `endApplyUpdate`.

Should we either:
- Not allow changes?
- Immediately apply them?

denise-opendaq: you have 
```cpp
obj.getOnPropertyValueWrite("prop1") += [](obj, event)
{
    obj.beginUpdate();
    //...
    obj.endUpdate();
}

obj.getOnPropertyValueWrite("prop2") += [](obj, event)
{
    obj.beginUpdate();
    //...
    obj.setPropertyValue("prop1", newValue);
    obj.endUpdate();
}
```

then somebody is triggering `prop2` callback does batching data. ending update(counter is decremented). and apply update, clear `updatingPropsAndValues`

while apply update, triggers event for prop1, which again does batching,  blabla...  clear **The same** `updatingPropsAndValues` which break  clear `updatingPropsAndValues` on the first layer. 
the pr just fixes this cases 

yes, end update calls twice in that case, but at least it does not corrupt the vector

JakaMohorko (approved):

PR #1188, opened 2026-05-05
Python GUI: Signals preview rewrite & UI Refactoring

Maksim3l:
# Brief
Rework the Python GUI signal preview panel: switch reader-event handling to the in-band reader-status path, redo the controls layout as a header table, and cover several visual and behavioral gaps.

# Description
- Replace the `on_component_core_event` subscription with reader-status event handling. `StreamReader` is now constructed with `skip_events=False`, and `_drain_reader` reads with `return_status=True` and dispatches `DATA_DESCRIPTOR_CHANGED` events inline. Descriptors come from the event packet's params dict, never from `signal.descriptor`, so samples produced under one descriptor can no longer be associated with a later descriptor's metadata.
- Honor the three-way assigned/null/unassigned semantics on each side of `DATA_DESCRIPTOR_CHANGED`: unassigned param means leave the cache alone, null descriptor means clear that side, real descriptor means apply.
- Cast `params['DataDescriptor']` and `params['DomainDataDescriptor']` to `IDataDescriptor` before reading attributes. The dict returns base-interface pointers and the Python attribute bindings live on the typed wrapper.
- Drop the `_descriptor_fingerprint` polling and the parent-walk event handler ancestry chain.
- Replace the single packed controls row with a four-column grid table: `Signal | Duration | Scale | Last value` with headers above their controls.
- Make the Duration combobox writable. Free-form input parses with a regex, clamps to `[0.01, 10]` s, and auto-appends the `s` suffix on commit (Enter, focus loss, or preset selection). Field rewrites to canonical form so `5` becomes `5s`.
- Pin X-axis grid labels to fixed positions (`-1s`, `-2s`, ...) instead of letting them slide leftward with the data. Edge labels read `0s` on the right and `-{window}s` on the left, with precision tracking the grid interval.
- Replace the `Logarithm` checkbox with a `Scale` dropdown (`Linear` / `Log`), shown only for 2D vector signals.
- Add a `Last value` toggle that hides the badge.
- Drop the collapsible-panel chevron and click handler; the panel is now permanently expanded.
- Strip the empty-state placeholder text from `No data` to `None`, centered in the plot area.
- Trim the left axis margin from 65 px to 38 px, sized to the actual Y-label widths.
- Hide the signal preview banner when the node has no output signals; hide input ports view when the node has no inputs.
- Guard `_ingest` against an empty descriptor cache so reads that arrive before the first descriptor event drop their samples instead of crashing.
- Reimplemented hiding inputs if they are empty
- Implemented hiding signal preview if there are no signals
- Added dynamic placement for the Y axis and it's labels

# Usage example
```
py -m opendaq
```

https://github.com/user-attachments/assets/e1e5540b-1f00-4723-a163-74bee165f5a2

### DIFF
--- a/examples/applications/python/GUI Application/gui_demo/components/signal_preview_panel.py
+++ b/examples/applications/python/GUI Application/gui_demo/components/signal_preview_panel.py
@@ -32,7 +32,6 @@ class SignalPreviewPanel(ttk.Frame):
         self.node = node
         self.context = context
 
-        self._collapsed = False
         self._reader = None
         self._selected_signal = None
         self._poll_job = None
@@ -71,19 +70,13 @@ class SignalPreviewPanel(ttk.Frame):
         self.bind('<Destroy>', self._on_destroy)
 
     def _build_toggle_bar(self):
-        self._toggle_frame = utils.make_banner(self, '\u25BC Signal Preview')
-        self._toggle_frame.configure(cursor='hand2')
-
-        self._arrow_label = self._toggle_frame.winfo_children()[0]
-        self._arrow_label.configure(cursor='hand2')
-
-        for w in (self._toggle_frame, self._arrow_label):
-            w.bind('<Button-1>', lambda _e: self._toggle())
+        self._toggle_frame = utils.make_banner(self, 'Signal Preview')
+        self._toggle_frame.pack_configure(padx=(0, 17))
 
     def _build_content(self):
         self._content_frame = ttk.Frame(self)
         dd_row = ttk.Frame(self._content_frame)
-        dd_row.pack(fill=tk.X, padx=4, pady=(4, 2))
+        dd_row.pack(fill=tk.X, padx=(4,17), pady=(4, 2))
 
         ttk.Label(dd_row, text='Signal:').pack(side=tk.LEFT, padx=(0, 4))
 
@@ -112,7 +105,7 @@ class SignalPreviewPanel(ttk.Frame):
         self._chart = tk.Canvas(
             self._content_frame, bg=self._BG,
             height=self.CANVAS_HEIGHT, highlightthickness=0)
-        self._chart.pack(fill=tk.BOTH, expand=True, padx=4, pady=(2, 4))
+        self._chart.pack(fill=tk.BOTH, expand=True, padx=(4,17), pady=(2, 4))
 
         self._chart.bind('<Configure>', lambda _e: self._invalidate_chart())
         self._block_mousewheel_recursive(self)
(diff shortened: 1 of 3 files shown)


tomaz-cvetko:
If I connect to SIRIUS-X and look at the system monitor function block, there are a couple of signals (asynchronous). The preview works in terms of showing last value, but no line is drawn. Is that expected?

I did experience some freezes that seemed correlated with late responses to reading last value, but as far as I can tell they may not be a direct consequence of this PR.

examples/applications/python/GUI Application/gui_demo/components/signal_preview_panel.py
```diff
@@ -368,28 +325,83 @@ def _poll_tick(self):
...
-            self._recreate_reader()
+            if status.read_status == daq.ReadStatus.Event:
+                self._handle_event_packet(status.event_packet)
+                continue
 
-        self._poll_job = self.after(self.POLL_MS, self._poll_tick)
+            if len(values) > 0:
+                self._ingest(values, domain_ticks)
+
+            return
```
tomaz-cvetko: This returns as soon as non-zero number of samples have been read, which may be insufficient to drain the packet queue.

There is also no guarantee that in the 33ms that you are waiting the queue won't look like `10samples - Event - 10samples - Event`. In this case, the code would process 10samples, then return. In the next run it would process 1 event and the next 10samples, but wouldn't even catch the second event from the first 33ms.

This should be a "do-while" style loop where you try it once and only exit the loop when both `read_status == ReadStatus.Ok` and `len(values) == 0`.

examples/applications/python/GUI Application/gui_demo/components/signal_preview_panel.py
```diff
@@ -368,28 +325,83 @@ def _poll_tick(self):
...
+            
+    def _apply_descriptors(self, data_descriptor, domain_descriptor):
+        if data_descriptor is not None:
+            unit = getattr(data_descriptor, 'unit', None)
+            symbol = getattr(unit, 'symbol', None) if unit is not None else None
+            self._unit_str = str(symbol) if symbol is not None else ''
+
+        if domain_descriptor is not None:
+            res = getattr(domain_descriptor, 'tick_resolution', None)
+            if res is None:
```
tomaz-cvetko: This is probably clearing the cache in the case of `NullDataDescriptor` and it probably works like it should for all cases. The most obvious characteristic of the `NullDataDescriptor` is its sample type being `SampleType::Null` (it's how it's defined in the core).

tomaz-cvetko:
> If I connect to SIRIUS-X and look at the system monitor function block, there are a couple of signals (asynchronous). The preview works in terms of showing last value, but no line is drawn. Is that expected?

I realized later that extending the time window resolves the issue - this happens only when there are fewer than 2 data points inside the time window.

tomaz-cvetko (changes requested):
I think we should change the reader draining loop. Otherwise the PR looks good

tomaz-cvetko (approved):
I approve, the requested changes have been made.

PR #1189, opened 2026-05-05
Fix open load module window on mac os gui

denise-opendaq:
# Brief

This PR fixes that GUI issues: 
- The button `load module` didn’t open the window because the extension filter wasn’t properly set on macOS. 
- Module prefix was showing N/A


Maksim3l (approved):
Looks good

PR #1200, opened 2026-05-14
Fix binary data packets not being recreated on client side

tomaz-cvetko:
# Brief

When binary data packets are received by the streaming client, they have to be recreated back into `BinaryDataPacketImpl` objects so that they behave the same way both on client and server side.

# Description

Binary data packets (that includes string packets) assume a single sample and the size of the buffer is determined by the sample size. The `BinaryDataPacketImpl` correctly handles this cases, but the existing implementation created regular data packets from headers and buffers sent for binary packets. This PR restores `BinaryDataPacket -> (header, buffer) -> BinaryDataPacket` the expected "transport" of binary data.

- Create a test where a binary data packet of sample type string is sent via native streaming protocol
- Test checks the `DataPacket::lastValue` method on the client-side packet to differentiate if the correct implementation is used (regular `DataPacketImpl` fails for string packets)
- In native streaming client create a `BinaryDataPacket` if sample type is either `String` or `Binary`

# Usage example

Restore expected behavior - use received string (binary) packets 

# API changes

No API changes.


shared/libraries/native_streaming_protocol/tests/test_streaming_protocol.cpp
```diff
@@ -947,6 +947,80 @@ TEST_P(StreamingProtocolTest, AddNotPublicSignalInConstructor)
     }
 }
 
+TEST_P(StreamingProtocolTest, SendStringDataPacketGetLastValue)
```
NikolaiShipilov: It feels like testing of such specifics would fit better into `shared/libraries/packet_streaming/tests/test_packet_streaming.cpp`

NikolaiShipilov (approved):
Lgtm, though the testing would make more sense if placed within `packet_streaming` library's internal tests collection

shared/libraries/native_streaming_protocol/tests/test_streaming_protocol.cpp
tomaz-cvetko: Thank you for the suggestion, I moved the test to the packet streaming tests.

denise-opendaq (approved):
looks good

PR #1201, opened 2026-05-14
Fix CMake configure failures with parquet recorder enabled

alexadereyko:
# Brief

Fix CMake configure failures when `DAQMODULES_PARQUET_RECORDER_MODULE=ON` by patching Arrow.

# Description

- Patch Arrow's `THIRDPARTY_DIR` to use `CMAKE_CURRENT_LIST_DIR/../thirdparty`
  instead of `arrow_SOURCE_DIR/thirdparty`. `arrow_SOURCE_DIR` is unreliable
  with FetchContent + SOURCE_SUBDIR on CMake <3.30 (points at Arrow root
  instead of `cpp/`); the file-relative path always resolves correctly.
- Patch Arrow's Apple libtool regex from `cctools-` to `cctools[-_]` so it
  accepts the `cctools_ld-NNN` format from recent Xcode.

> [!NOTE]
> This PR only fixes the misleading Arrow / Thrift errors that appeared first. 
> Parquet recorder still requires CMake 4.0.0 or higher. 
> Users on older CMake now get the proper message instead:
> `Parquet recorder module requires CMake 4.0.0 or higher`

### DIFF
--- a/external/arrow/CMakeLists.txt
+++ b/external/arrow/CMakeLists.txt
@@ -47,8 +47,10 @@ opendaq_dependency(
     ADD_FETCH_ALIAS     Arrow::arrow_static=arrow_static
     EXPECT_TARGET       Arrow::arrow_static
     SOURCE_SUBDIR       cpp
-    PATCH_FILES        
+    PATCH_FILES
                         "${CMAKE_CURRENT_SOURCE_DIR}/boost_linking.patch"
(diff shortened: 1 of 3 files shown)


NikolaiShipilov (approved):
Have checked this branch on Ubuntu 20.04 and it shows `Parquet recorder module requires CMake 4.0.0 or higher.` as expected, in comparison with RC branch where it fails at `Arrow` configuration step. 👍

PR #1204, opened 2026-05-15
Python GUI: Embed signal preview charts into output signal rows

Maksim3l:
# Brief
Move signal preview from a floating block-level panel into per-row expandable inline charts inside output signal rows, with per-signal duration controls and various UX fixes.

# Description
- Replace the block-wide `SignalPreviewPanel` overlay (attached via `place()` at the bottom of `BlockView`) with a new `OutputSignlGraf` component that lives inside each `OutputSignalRow`. Chartable signals get an expand/collapse arrow; clicking it creates or destroys the chart for that specific signal.
- Duration combobox with presets appears next to the signal name when the row is expanded. The chart's bottom-left axis label is also double-click editable. Both share the same `StringVar` so they stay in sync.
- The chart's bottom-left duration label switches to `xterm` cursor and gains an underline + color shift on hover to signal that it's editable.
- Each output signal row has a thin gray bottom border. The last row in a section hides it.
- The first chartable signal in a section auto-expands when a component is selected, so there's always a live preview visible.
- `BlockView` now shows `icons['channel']` for `IChannel` nodes instead of falling through to the `IFunctionBlock` icon, matching the tree view.
- Cogwheel padding in the block header defaults to `(6, 27)` and overrides to `(6, 10)` for `ISyncComponent` to align with the missing scrollbar.
- `OutputSignlGraf` tracks whether the domain rule is `Explicit` and uses `reader.read()` instead of `reader.read_with_domain()` for those signals, synthesizing a sample-index time axis.
- Guard `utils.yes_no[value]` against `None` values from disconnected OpcUA properties.
- Guard `device.operation_mode` access with a `None` check so disconnected devices don't crash the tree view.
- Change `lambda event=None` to `lambda event` in `generic_properties_treeview.py` context menu binding + add `None` guard in `show_menu`.

# Usage example
https://github.com/user-attachments/assets/1c00b5e5-c6b3-401f-8737-505f4fef7718


examples/applications/python/GUI Application/gui_demo/components/output_signal_graph.py
tomaz-cvetko, line 1: Change "graf" to "graph"

examples/applications/python/GUI Application/gui_demo/components/output_signal_graph.py
```diff
@@ -10,7 +10,7 @@
 
 from .. import utils
 
-class SignalPreviewPanel(ttk.Frame):
+class OutputSignlGraf(ttk.Frame):
```
tomaz-cvetko: -> Signal (missing A)
-> Graph (graf is Slovenian term)

examples/applications/python/GUI Application/gui_demo/components/output_signal_row.py
```diff
@@ -7,6 +7,7 @@
 from .. import utils
 from .attributes_dialog import AttributesDialog
 from .dialog import Dialog
+from .output_signal_graf import OutputSignlGraf
```
tomaz-cvetko: Spelling will be corrected after definition is changed

tomaz-cvetko (changes requested):
There is a spelling mistake in one of the classes. I think we should rename the `output_signal_graf.py` file to `..._graph.py`. Otherwise it looks good.

tomaz-cvetko (approved):
Looks good to me

JakaMohorko (approved):

PR #1205, opened 2026-05-15
Fix pinging on Linux as non root causing an access exception (raw sockets)

Unsttopabull:
## Brief
Fix raw socket access exception on Linux/macOS for non-root users

## Desciption 
`IcmpPing` uses raw ICMP sockets which require elevated privileges. On Linux and macOS, running without root (`sudo`) caused an access exception when checking device address reachability. The fix wraps the ICMP ping block in a `geteuid() == 0` check, skipping it for non-root users. Devices are still discovered via mDNS, but addresses are not marked as reachable via ICMP when running unprivileged.


denise-opendaq (approved):

core/opendaq/modulemanager/src/module_manager_impl.cpp
```diff
@@ -493,6 +493,9 @@ void ModuleManagerImpl::checkNetworkSettings(ListPtr<IDeviceInfo>& list)
         }
     }
 
+#if defined(__linux__)
+    if (geteuid() == 0)
```
denise-opendaq, line 497: the ipv4 addresses will be Unreachable. But they should be unknown I think

JakaMohorko (approved):

PR #1207, opened 2026-05-18
Fix set of tests; Disable remaining unstable tests

alexadereyko:
# Brief

Disable or fix flaky tests

**DISABLED**:
- `test_function_block`:
  - `FunctionBlockTest.SetDomainDescriptorUnderLock`
- `test_device_modules`:
  - `NativeDeviceModulesTest.GetConnectedClientsInfo`
  - `NativeDeviceModulesTest.SettingOperationMode`
  - `NativeDeviceModulesTest.SettingOperationModeWithoutPermissions`
  - `NativeDeviceModulesTest.SettingOperationModeWithPermissions`
  - `NativeDeviceModulesTest.SettingOperationModeWithPermissionsNestedDevice`
  - `NativeDeviceModulesTest.SettingOperationModeWithPermissionsForInvisibleDevice`
  - `OpcuaDeviceModulesTest.SettingOperationMode`
  - `NativeC2DStreamingTest.ClientLostConnection`
  - `StreamingTest.DataPackets` — `daq.lt://` parameter only
  - `StreamingTest.SignalDescriptorEvents` — `daq.lt://` parameter only
  - `StreamingTest.SetNullDescriptor` — `daq.lt://` parameter only
  - `StreamingTest.ChangedDataDescriptorBeforeSubscribe` — `daq.lt://` parameter only
  - `StreamingTest.LastValue` — `daq.opcua://[::1]/` parameter only (IPv6 OPC UA endpoint)

**FIXED**:
- `test_docs`:
  - `HowToAccessControl.ConnectWithUsername` raised ConnectionTimeout to 5000 ms via `Device.OpenDAQNativeConfiguration.TransportLayerConfig` — the 1 s default was too tight for the native config handshake on slow Debug CI
- `test_ref_fb_module`:
  - `StatisticsTest.StatisticsTestOverlapExplicitRangeRule` `memcpy` → `memmove` in `StatisticsFbImpl::copyRemainingCalcBuf` — in-place `calcBuf` compaction had overlapping `src`/`dst` -- Undefined Behaviour
- `test_audio_device_module`:
  - SEGFAULT fixed by switching `WAVWriterFbImpl::createInputPort()`'s `onDataAvailable` to a weak-ref capture — prevents use-after-free during destruction
- `test_device_modules`:
  - `SubDevicesTest.LeafStreamingToClient` added recursive config lock to `addProperty`, `enableCoreEventTrigger` and `disableCoreEventTrigger` in `GenericPropertyObjectImpl` — these touched `localProperties` without the lock that `removeProperty` and the property-value API already take; teardown raced `disableCoreEventTrigger` iterating the deque-backed `tsl::ordered_map` with the config-protocol thread inside `addProperty`, crashing on a value-initialized `deque` iterator
  - `StreamingTest.LastValue` (and other `StreamingTest.*` cases over `daq.opcua://`) bumped OpcUaModules to include the `TmsServer::stop` fix — `stop` nulled `OpcUaServer`'s `UA_Server*` before releasing the device/context, so a device worker thread emitting core events during teardown reached live `TmsServerSignal` listeners that called into the stopped server — use-after-free; the fix releases device/context (unsubscribing core-event listeners) before stopping the server
  - native-streaming family SEGFAULTs fixed by calling `stopReading()` before the `serverStopped` early-return in `NativeStreamingServerImpl::stopServerInternal()`

### DIFF
--- a/CMakeBasePresets.json
+++ b/CMakeBasePresets.json
@@ -338,7 +338,7 @@
                 "verbosity": "verbose"
             },
             "execution": {
-                "timeout": 1200
+                "timeout": 3600
             }
         },
         {
--- a/CMakeBasePresets.json
+++ b/CMakeBasePresets.json
@@ -334,7 +334,11 @@
             "description": "Run all tests excluding unstable ones",
             "configurePreset": "ci",
             "output": {
-                "outputOnFailure": true
+                "outputOnFailure": true,
+                "verbosity": "verbose"
+            },
+            "execution": {
+                "timeout": 1200
             }
         },
         {
--- a/CMakeBasePresets.json
+++ b/CMakeBasePresets.json
@@ -334,11 +334,7 @@
             "description": "Run all tests excluding unstable ones",
             "configurePreset": "ci",
             "output": {
-                "outputOnFailure": true,
-                "verbosity": "verbose"
-            },
-            "execution": {
-                "timeout": 3600
+                "outputOnFailure": true
             }
         },
         {
--- a/external_modules/CMakeLists.txt
+++ b/external_modules/CMakeLists.txt
@@ -68,7 +68,7 @@ if (OPENDAQ_ENABLE_OPCUA AND (DAQMODULES_OPENDAQ_CLIENT_MODULE OR DAQMODULES_OPE
     FetchContent_Declare(
         OpcUaModules
         GIT_REPOSITORY https://github.com/openDAQ/OpcUaModules.git
-        GIT_TAG        5bce4a3421e715354317f316947eb6a716aca01e
+        GIT_TAG        fc6af466eede67c4726910b8aa97be46ce697a3b
         GIT_PROGRESS   ON
         GIT_SHALLOW    OFF
         SOURCE_SUBDIR  "INVALID"
--- a/external_modules/CMakeLists.txt
+++ b/external_modules/CMakeLists.txt
@@ -68,7 +68,7 @@ if (OPENDAQ_ENABLE_OPCUA AND (DAQMODULES_OPENDAQ_CLIENT_MODULE OR DAQMODULES_OPE
     FetchContent_Declare(
         OpcUaModules
         GIT_REPOSITORY https://github.com/openDAQ/OpcUaModules.git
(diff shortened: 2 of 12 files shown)


CMakeBasePresets.json
```diff
@@ -334,7 +334,11 @@
             "description": "Run all tests excluding unstable ones",
             "configurePreset": "ci",
             "output": {
-                "outputOnFailure": true
+                "outputOnFailure": true,
+                "verbosity": "verbose"
+            },
+            "execution": {
+                "timeout": 1200
```
alexadereyko: Restore before commit

Maksim3l (approved):
Looks good.

external_modules/CMakeLists.txt
```diff
@@ -68,7 +68,7 @@ if (OPENDAQ_ENABLE_OPCUA AND (DAQMODULES_OPENDAQ_CLIENT_MODULE OR DAQMODULES_OPE
     FetchContent_Declare(
         OpcUaModules
         GIT_REPOSITORY https://github.com/openDAQ/OpcUaModules.git
-        GIT_TAG        5bce4a3421e715354317f316947eb6a716aca01e
+        GIT_TAG        fc6af466eede67c4726910b8aa97be46ce697a3b
```
alexadereyko: Replace hash after commit merge

NikolaiShipilov (approved):
👏 looks good. 
Considering the amount of tests turning into disabled (or partially disabled) with this PR it would be great to add some in-code comments explaining the reason for each case: 
- e.g. just mark the line in the test body where failure occurs, if applicable; 
- or provide some details right before the test definition - just a statement if test crashes with a segfault or freezes without a completion.

JakaMohorko (approved):
Looks really good, nice work!

external_modules/CMakeLists.txt
alexadereyko: Replaced

CMakeBasePresets.json
alexadereyko: Restored

PR #1206, opened 2026-05-18
setOperationModeRecursive() fix

viacheslauK:
# Brief
- now `setOperationModeRecursive()`  sets OM for all devices (not only for visible)

PR #1177, opened 2026-04-24
Reset missing attributes to default values in update.

dejancrn:
# Brief

Reset missing component attributes to default values in the configuration update.

PR #1209, opened 2026-05-20
Added API changelog to 3.40 changelog

Maksim3l:


changelog/changelog_3.30-3.40.md
```diff
@@ -102,3 +102,110 @@ On load configuration, all non-static function blocks will be removed and recrea
 
 ### [#1037](https://github.com/openDAQ/openDAQ/pull/1037) Mandatory device types
 To enable static components, devices must include the Device Type in their Device Info objects. Modules set the value within the onGetInfo overriding method by calling IDeviceInfoConfig::setDeviceType().
+
```
alexadereyko, line 118: The existing changelogs never list internal/protected API, so I don't think we should introduce them here either. Most of the remaining entries in the "API Surface Changes" section also just duplicate what's already described above.

The actually missing info is the set of bug fixes / changes that aren't in the file yet. I'd suggest dropping the API surface section and adding these lines to the appropriate existing sections instead:

## Python

- [#1142](https://github.com/openDAQ/openDAQ/pull/1142) Improve properties treeview UX in Python GUI demo app: float formatting, method Run buttons and cleaner object display.

## Bug fixes

- [#1206](https://github.com/openDAQ/openDAQ/pull/1206) Fix setting operation mode recursively to apply to hidden child devices, not only visible ones.
- [#1197](https://github.com/openDAQ/openDAQ/pull/1197) Propagate parent active state on configuration update.
- [#1177](https://github.com/openDAQ/openDAQ/pull/1177) Reset missing attributes to their default values during update.
- [#1169](https://github.com/openDAQ/openDAQ/pull/1169) Fix active not being set via native protocol when parent active is false.
- [#1160](https://github.com/openDAQ/openDAQ/pull/1160) Align behaviour between the ObjectProperty and ObjectPropertyBuilder factories.
- [#1111](https://github.com/openDAQ/openDAQ/pull/1111) Reset last value calculation values on unsubscribe completion.
- [#1040](https://github.com/openDAQ/openDAQ/pull/1040) Check for cyclic references when connecting input ports; cyclic connections are now rejected.

## Misc

- [#1120](https://github.com/openDAQ/openDAQ/pull/1120) Change reachability status to a sparse selection property type.

Maksim3l: Aha okay, I wasn't really sure what to do when it said to make a 'API changelog for 3.40'

alexadereyko (approved):
Looks good to me.
For future releases, when we settle on the diff API output format, we can move it to the `changelog/api/` directory

PR #1210, opened 2026-05-20
C++ iterator improvements

anzeskerjancDS:
# Description

I fixed and speedup comparing with end sentinel. In that case we don't compare by value.

The problem remains with comparing two non-end iterators. But luckly we don't have such use cases.

```
auto list = List<Int>(1, 2, 1, 4);

auto it1 = list.begin();
auto it2 = list.begin();

ASSERT_EQ(it1, it2);

++it2;
ASSERT_NE(it1, it2);

++it2;
ASSERT_NE(it1, it2);  // fails
```

### DIFF
--- a/core/coretypes/include/coretypes/objectptr.h
+++ b/core/coretypes/include/coretypes/objectptr.h
@@ -1717,7 +1717,6 @@ IIterator* ObjectPtr<T>::createStartIteratorInterface() const
     errCode = iterable->createStartIterator(&iterator);
     checkErrorInfo(errCode);
 
-    iterator->moveNext();
     return iterator;
 }
 
@@ -1742,7 +1741,6 @@ IIterator* ObjectPtr<T>::createEndIteratorInterface() const
     errCode = iterable->createEndIterator(&iterator);
     checkErrorInfo(errCode);
 
-    iterator->moveNext();
     return iterator;
 }
 
--- a/changelog/changelog_3.30-3.40.md
+++ b/changelog/changelog_3.30-3.40.md
@@ -38,6 +38,7 @@
 
 ## Bug fixes
 
+- [#1210](https://github.com/openDAQ/openDAQ/pull/1210) Fix and optimize comparing with end iterator (C++).
 - [#1200](https://github.com/openDAQ/openDAQ/pull/1200) Recreate binary data packets into `BinaryDataPacketImpl` objects on client side of native streaming protocol.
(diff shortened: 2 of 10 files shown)


core/coretypes/include/coretypes/objectptr.h
```diff
@@ -1717,7 +1717,6 @@ IIterator* ObjectPtr<T>::createStartIteratorInterface() const
     errCode = iterable->createStartIterator(&iterator);
     checkErrorInfo(errCode);
 
-    iterator->moveNext();
```
tomaz-cvetko, line 1720: This is potentially changing behavior and it is unrelated to the .end() check. Why do we need this?

tomaz-cvetko: I think this shouldn't be removed as `IteratorBaseImpl` seems to need this first call to `moveNext` to have its flag `started` set to true.

denise-opendaq: he move the logic to NativeIterator<U>::NativeIterator(IIterator*&& iterator)
but from my point of view its not really clear

anzeskerjancDS: I moved this call to NativeIterator. When we create NativeIterator, we need to check if this iterator is end iterator (empty container). If we do this here, we lost this information. 

The behavior change is within C++ code that is probably never explicitly called. Therefore, I doubt it will break anyone's existing code.

tomaz-cvetko: Ok, I see it now. I think this solution will probably work in terms of speeding up/improving the standard use case of the range iteration.

However, the issue is probably in the `iterator_helper.h` in the `compareIterators` method. The call to `obj2->equals(obj1, equal)` is not "the c++ way" for comparing iterators - this should be comparison of pointers (which with some additional details the iterators are) and not values... But we probably have some syntax that we mustn't break where comparing iterators is actually comparing their values?

anzeskerjancDS: Yes, you are right. The compareIterators method doesn't work. I just fixed and improved it for the standard/common use case.

Our interface API uses a 'Python' style iterator (Next, GetValue).
So I think future changes to the interface API should be:
- Remove IIterable::createEndIterator. Confusing because Python-style iterator doesn't have such a term. In this proposed change, I have removed the calls to this function from C++.
- Undefined behavior when comparing two non-end iterators (C++)

JakaMohorko:
Duplicated by #1216 with a different implementation approach.

PR #1212, opened 2026-05-22
fix for test_device_modules and test_audio_device_module SEGFAULTs

Maksim3l:
# Brief
Fix teardown race conditions causing flaky CI segfaults in test_audio_device_module and test_device_modules

# Description
- Fix use-after-free in `WAVWriterFbImpl::createInputPort()` by replacing raw `this` capture with weak reference pattern in the `onDataAvailable` callback, matching the existing pattern in `power_reader_fb_impl`, `sum_reader_fb_impl`, and `multi_csv_recorder_impl`
- Fix thread lifecycle in `NativeStreamingServerImpl::stopServerInternal()` by moving all IO thread stops (`stopReading`) before the `serverStopped` guard so they are always joined even on repeated calls

# Required module changes
Modules implementing function blocks with `onDataAvailable` callbacks should use the weak reference pattern to prevent use-after-free during destruction:

Instead of:
```cpp
reader.setOnDataAvailable([this] { processInputData(); });
```
Do:
```cpp
auto thisWeakRef = this->template getWeakRefInternal<IFunctionBlock>();
reader.setOnDataAvailable(
    [this, thisWeakRef = std::move(thisWeakRef)]
    {
        const auto thisFb = thisWeakRef.getRef();
        if (thisFb.assigned())
            processInputData();
    });
```

### DIFF
--- a/modules/native_streaming_server_module/src/native_streaming_server_impl.cpp
+++ b/modules/native_streaming_server_module/src/native_streaming_server_impl.cpp
@@ -272,6 +272,7 @@ void NativeStreamingServerImpl::stopProcessingOperations()
 
 void NativeStreamingServerImpl::stopServerInternal()
 {
+    stopReading();
     if (serverStopped)
         return;
 
@@ -296,7 +297,6 @@ void NativeStreamingServerImpl::stopServerInternal()
     registeredClientIds.clear();
     disconnectedClientIds.clear();
 
-    stopReading();
     serverHandler->stopServer();
     stopTransportOperations();
     stopProcessingOperations();
--- a/examples/modules/audio_device_module/src/wav_writer_fb_impl.cpp
+++ b/examples/modules/audio_device_module/src/wav_writer_fb_impl.cpp
@@ -185,7 +185,15 @@ void WAVWriterFbImpl::createInputPort()
 {
     inputPort = createAndAddInputPort("Input", PacketReadyNotification::Scheduler);
     reader = StreamReaderFromPort(inputPort, SampleType::Undefined, SampleType::Undefined);
-    reader.setOnDataAvailable([this] { processInputData();});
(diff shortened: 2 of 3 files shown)


modules/native_streaming_server_module/src/native_streaming_server_impl.cpp
```diff
@@ -296,7 +297,6 @@ void NativeStreamingServerImpl::stopServerInternal()
     registeredClientIds.clear();
     disconnectedClientIds.clear();
 
-    stopReading();
     serverHandler->stopServer();
     stopTransportOperations();
     stopProcessingOperations();
```
NikolaiShipilov, line 302: This a bit unfollows the statement from the PR description:
> - Fix thread lifecycle in `NativeStreamingServerImpl::stopServerInternal()` by moving all IO thread stops (`stopReading`, `stopTransportOperations`, `stopProcessingOperations`) before the `serverStopped` guard so they are always joined even on repeated calls

I guess the moving of the `stopReading();` line alone to the top fixes the issue, or ?

NikolaiShipilov (approved):

alexadereyko (approved):
Looks good to me

modules/native_streaming_server_module/src/native_streaming_server_impl.cpp
Maksim3l: Yes, moving just `stopReading()` before the guard is enough. That's the thread (`NatSrvStreamRead`) that was confirmed stuck in the debugger. The other two threads (`NatSrvStreamTrans`, `NatSrvProc`) were not observed as problematic and their stop order relative to `serverHandler->stopServer()`. Originally I had them all moved in front because I wasn't sure which one specifically caused the issue.

PR #1214, opened 2026-05-25
Fix packet enqueue order in `sendPackets`

JakaMohorko:
# Brief

Fixes the order in which packets are enqueued in `sendPackets`. They are now always correctly enqueued front-to-back.

# Required module changes

`ISignalConfig::sendPackets` had an issue where packets were sometimes enqueued front-to-back (first packet in list is enqueued first) and sometimes back-to-front. This inconsistency was addressed to always enqueue packets front-to-back.

Module implementations where `sendPackets` is used should check whether this change in behaviour reversed the order in which packets are enqueued and adjust the packet list order accordingly.

### DIFF
--- a/changelog/changelog_3.30-3.40.md
+++ b/changelog/changelog_3.30-3.40.md
@@ -105,5 +105,11 @@ On load configuration, all non-static function blocks will be removed and recrea
 
 ## Required module changes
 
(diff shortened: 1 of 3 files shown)


tomaz-cvetko (approved):
Looks good to me, front to back seems like the right choice.

PR #1216, opened 2026-05-27
List/Dict iteration performance improvement

JakaMohorko:
# Brief

Speeds up comparison between end sentinel when iterating over openDAQ list and dictionary objects.

### DIFF
--- a/core/coreobjects/src/property_object_class_builder_impl.cpp
+++ b/core/coreobjects/src/property_object_class_builder_impl.cpp
@@ -194,10 +194,8 @@ ListPtr<IProperty> PropertyObjectClassBuilderImpl::getProperties() const
         }
     }
 
-    for (const auto& prop : props)
-    {
-        properties.unsafePushBack(prop.second);
-    }
+    for (const auto& prop : props.getValues())
+        properties.unsafePushBack(prop);
 
     return properties;
 }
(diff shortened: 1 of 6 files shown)


core/coretypes/include/coretypes/iterator_support.h
```diff
@@ -87,6 +87,18 @@ NativeIterator<U>& NativeIterator<U>::operator++()
...
+    ErrCode err2 = other.iterator->getCurrent(nullptr);
+    
+    if (err1 == OPENDAQ_ERR_NOTASSIGNED && err2 == OPENDAQ_ERR_NOTASSIGNED)
+        return false;
+    
+    if (err1 == OPENDAQ_ERR_NOTASSIGNED || err2 == OPENDAQ_ERR_NOTASSIGNED)
+        return true;
+
+    checkErrorInfo(err1);
+    checkErrorInfo(err2);
```
tomaz-cvetko: In case neither is `it == end`, the error code won't have `OPENDAQ_ERR_NOTASSIGNED` but `OPENDAQ_ERR_ARGUMENT_NULL` so the checkErrorInfo will catch that - I think we shouldn't check for errors here as we're triggering them intentionally by passing in `nullptr` to the `getCurrent`.

core/coreobjects/src/property_object_class_builder_impl.cpp
```diff
@@ -194,10 +194,8 @@ ListPtr<IProperty> PropertyObjectClassBuilderImpl::getProperties() const
         }
     }
 
-    for (const auto& prop : props)
-    {
-        properties.unsafePushBack(prop.second);
-    }
+    for (const auto& prop : props.getValues())
+        properties.unsafePushBack(prop);
```
tomaz-cvetko, line 198: Why the change to `.getValues()`?

anzeskerjancDS: The key-value iterator is heavy because each iteration step allocates a new, two-element daq::List to return the key-value pair.

tomaz-cvetko (approved):
Looks good to me

PR #1220, opened 2026-06-01
getLastValueWithTimestamp() API

viacheslauK:
# Brief

Add `ISignal::getLastValueWithTimestamp` to retrieve a signal's last value together with its corresponding domain timestamp, normalized to UNIX time (**microseconds** since epoch).

# Description

  - Add `getLastValueWithTimestamp(IBaseObject** value, IBaseObject** timestamp)` to the `ISignal` interface
  - Implement it in `SignalBase`, computing the timestamp from the domain packet (origin + tick × resolution) and converting it to **microseconds** since the UNIX epoch
  - Cache the last raw timestamp and domain descriptor alongside the existing last-value caching, refreshed in `setLastValueFromPacket`
  - Add domain-descriptor validation for timestamp support (numeric sample type, no dimensions, unit symbol "s", assigned origin)
  - Refactor the existing `getLastValue` to share a common `getLastValueImpl` helper
  - Reset the new timestamp/domain state on unsubscribe in `MirroredSignalBase`
  - Add tests

# Usage example

In C++ use:
```
  BaseObjectPtr value;
  BaseObjectPtr timestamp;
  signal->getLastValueWithTimestamp(&value, &timestamp);

  if (value.assigned() && timestamp.assigned())
  {
      const Float lastValue = value;          // cast depends on the signal's sample type
      const Int unixTimeUs = timestamp;       // UNIX time since epoch, in microseconds
  }
```
```
  BaseObjectPtr value;
  BaseObjectPtr timestamp;
  timestamp = signal.getLastValueWithTimestamp(&value);

  if (value.assigned() && timestamp.assigned())
  {
      <...>
  }
```

# API changes

`ISignal` gains one new pure virtual function:
`virtual ErrCode INTERFACE_FUNC getLastValueWithTimestamp(IBaseObject** value, IBaseObject** timestamp) = 0;`

None. The change is purely additive — existing calls to `getLastValue` are unaffected. Applications may optionally adopt the new method to obtain a timestamp alongside the value.

# Required module changes

No source changes are required for modules that derive their signal implementations from SignalBase — they inherit the new method automatically.


core/opendaq/signal/include/opendaq/signal_impl.h
```diff
@@ -1329,21 +1341,182 @@ template <typename TInterface, typename... Interfaces>
...
+    else if (!origin.empty() && origin.back() == 'Z')
+        origin = origin.erase(origin.size() - 1) + "+00:00";
+    else if (origin.find('+') == std::string::npos)
+        origin += "+00:00";
+
+    std::chrono::system_clock::time_point signalEpoch;
+    std::istringstream ss(origin);
+    date::from_stream(ss, "%FT%T%z", signalEpoch);
+    if (ss.fail())
+        DAQ_THROW_EXCEPTION(InvalidParametersException, "Origin string is not a valid ISO 8601 date-time.");
```
tomaz-cvetko: Use `daq::reader::parseEpoch(origin)` helper to avoid duplication

core/opendaq/signal/include/opendaq/signal.h
```diff
@@ -143,6 +143,25 @@ DECLARE_OPENDAQ_INTERFACE(ISignal, IComponent)
...
+     * @param[out] value The IBaseObject value can be a nullptr if there is no value, or if the data type is not supported by the function.
+     * @param[out] timestamp The IBaseObject timestamp corresponding to the returned value. It can be a nullptr if there is no timestamp available.
+     *
+     * If a value is assigned, it can be cast based on the signal description to IFloat if the type is Float32 or Float64,
+     * to IInteger if the type is Int8 through Int64 or UInt8 through UInt64, to IComplexNumber if the type is ComplexFloat32 or ComplexFloat64,
+     * to IRange if the type is RangeInt64, to IString if the type is String, to IStruct if the type is Struct, and to IList of the aforementioned types if there is exactly
+     * one dimension.
+     *
+     * If a timestamp is assigned, it can typically be cast to IInteger and represents the UNIX time since epoch in microseconds
+     * corresponding to the returned signal value.
```
JakaMohorko, line 158: I would list the criteria for this function to work - unit is seconds, the origin is an ISO8601 date.

core/opendaq/signal/include/opendaq/signal_impl.h
```diff
@@ -1329,21 +1341,182 @@ template <typename TInterface, typename... Interfaces>
...
+
+    const auto originStr = lastDomainDescriptor.getOrigin();
+    if (!originStr.assigned())
+        DAQ_THROW_EXCEPTION(NotAssignedException, "Origin in domain descriptor is not assigned.");
+
+    auto resolution = lastDomainDescriptor.getTickResolution();
+    if (!resolution.assigned())
+        resolution = Ratio(1, 1);
+
+    resolution = (resolution / Ratio(1, 1000000)).simplify(); // Convert to microseconds
```
JakaMohorko: We should  cache this value when assigning the descriptor. openDAQ ratio calculations are expensive, as both the division and simplification create a new ratio object - openDAQ core types are immutable and we must create a new one whenever we modify the values.

This calculation here creates 3 ratio objects.

core/opendaq/signal/include/opendaq/signal_impl.h
JakaMohorko: Similarly, this could also be cached on descriptor change. However, that would place the a burden on devices that are sending  the data. We could have a flag `descriptorChanged` which is set during `sendPacket`. If `true`, the origin and resolution are re-calculated during `getLastValueWithTimestamp`, otherwise the cached values are used.

core/opendaq/signal/include/opendaq/signal_impl.h
```diff
@@ -1329,21 +1341,182 @@ template <typename TInterface, typename... Interfaces>
...
+            lastDomainDescriptor = nullptr;
+            return;
+        }
+
+        lastRawTimestamp.resize(lastDomainDescriptor.getSampleSize()); // 8 bytes for int64/uint64
+        void* rawValue = lastRawTimestamp.data();
+        const ErrCode errCode = domainPacket->getRawLastValue(&rawValue);
+        if (errCode != OPENDAQ_SUCCESS)
+            lastDomainDescriptor = nullptr;
     }
```
JakaMohorko: This operation needs to be really lightweight. Checking the descriptor on every `sendPacket` call is really expensive, resulting in a large overhead in calculation.

I would suggest checking whether the descriptor has changed since the last packet and validate only then. We shouldn't use the `==` operator though, as that does a value comparison that's expensive as well. We should check whether we are actually dealing with the same object by calling `descriptor.getObject()` and comparing the pointer addresses.

core/opendaq/streaming/include/opendaq/mirrored_signal_impl.h
```diff
@@ -383,16 +383,15 @@ ErrCode MirroredSignalBase<Interfaces...>::unsubscribeCompletedInternal(IString*
...
-    else
     {
+        LockGuardPtr lock;
+        if (syncLock)
+            lock = this->getRecursiveConfigLock2();
+
         this->lastDataValue = nullptr;
         this->lastDataDescriptor = nullptr;
+        this->lastTimestamp = nullptr;
+        this->lastDomainDescriptor = nullptr;
```
JakaMohorko: `getLastValueWithTimestamp` over native will work only when the signal is subscribed unless we add a corresponding RPC. When the signal is not subscribed, data is not flowing from the server, thus no packets/last values are cached.

We address this by invoking a RPC when the values are missing. This RPC should be added.

core/opendaq/signal/include/opendaq/last_value_cache.h
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    void calculateTimestamp(const BaseObjectPtr& timestamp)
+    {
+        if (!timestamp.assigned())
+            DAQ_THROW_EXCEPTION(NotAssignedException, "Timestamp value is not assigned.");
+
+        int64_t tick = 0;
+        if (const auto floatObj = timestamp.asPtrOrNull<IFloat>(); floatObj.assigned())
+            tick = static_cast<int64_t>(floatObj);
```
tomaz-cvetko: I don't think casting float domain ticks to int64 will produce the desired results - if someone were to use floats as the domain, they would probably have 42325.625s since origin that would get cast to 42325...

We don't expect floats as time domain though so this might be irrelevant.

core/opendaq/signal/include/opendaq/last_value_cache.h
```diff
@@ -0,0 +1,293 @@
...
+        if (packet.assigned())
+        {
+            if (const auto domainPacket = packet.getDomainPacket(); domainPacket.assigned())
+            {
+                if (auto packetDomainDescriptor = domainPacket.getDataDescriptor(); packetDomainDescriptor.assigned())
+                {
+                    if (reinterpret_cast<std::uintptr_t>(packetDomainDescriptor.getObject()) != reinterpret_cast<std::uintptr_t>(domainDescriptor.getObject()))
+                    {
+                        domainDescriptor = std::move(packetDomainDescriptor);
+                        internalCacheValid = false;
```
tomaz-cvetko: Please rename this variable to something more meaningful. If I understand correctly, if false, this means "resolution and offset must be recalculated". I would not call these two variables an "internal cache" to the LastValueCache (internal cache is to vague, since the cached value itself might be considered the internal cache). I suggest maybe `domainInfoValid`.

tomaz-cvetko (approved):
Lgtm

JakaMohorko (approved):

PR #1228, opened 2026-06-06
Copy docstrings to autogenerated C binding headers

adolenc:
# Brief

Add docstrings, type hints, and `DECLARE_OPENDAQ_INTERFACE` class hierarchy to the generated C bindings header files. This can be used to massively improve the bindings for languages that interface with openDAQ via C FFI.

Note that I have not committed the new `RTGen.C.dll`.


adolencds:
This MR really opens up a new set of possibilities 🙂

image

JakaMohorko (commented):
While I'm not entirely certain about copying the `//DECLARE_OPENDAQ_INTERFACE(daqArgumentInfo, daqBaseObject)` comments, I also don't think it's the worst approach. 

It does mean, however, that we can never really remove the comment again in the future. To provide options for the future, I would rather implement a `DAQ_EXTENDS_INTERFACE(base)` macro that currently evaluates to nothing.

This might allow us to add (for example) Clang annotations later on:

```C
typedef struct daqDevice daqDevice DAQ_EXTENDS_INTERFACE(daqFolder);
typedef struct daqDeviceInfo daqDeviceInfo;
typedef struct daqDeviceDomain daqDeviceDomain;
typedef struct daqFolder daqFolder;
...
```

adolenc:
I like the idea about using macros a lot actually, I would just slightly touch up your suggestion:

 - instead of using "context" code to figure out what the derived class is for a base class, I would use a stand-alone `DAQ_EXTENDS_INTERFACE(intf, base)`. This has a benefit that both types are passed into the macro so your macro code can be quite a bit more specific.
 - I would also encode the `//[ElementType...]` and `//[defaultValue...]` comments as _inline_ macros, making them even easier to parse; i.e.:
```
daqErrCode EXPORTED daqInstance_setRootDevice(daqInstance* self, daqString* connectionString, daqPropertyObject* config DAQ_DEFAULT_VALUE(nullptr));
daqErrCode EXPORTED daqInstance_addStandardServers(daqInstance* self, daqList** servers DAQ_LIST_ELEMENT_TYPE(daqServer));
```

As you suggested for now all of these can just expand to noop, but I agree that in this way we leave things a lot more open for adding extra functionality later. Let me try updating the generator like this and we see what it looks like.

JakaMohorko (approved):
This looks really solid to me now. Feels like it's also very flexible for usage in generating other bindings - macros can be also be re-defined with annotations depending on what's needed by other generators.

We can use this approach to port other RTGen tags to the C bindings as well when needed.

adolenc:
I rebased the pr to the top of main and cleaned things up a bit. 
I think in a separate pr I should also fix the c generator so it automatically outputs only the working c code, because right now it still outputs a bunch of noise like daqvoid and typedeffing floats that need to be manually cleaned up on every regeneration.

PR #1231, opened 2026-06-15
Fix ComparableValue value scaling with large domain resolution 

ankirn-ds:
# Brief

Fix crashing multireader with device domain resolution 15M and unix epoch as epoch of the domain

# Description

ComperableValue value calculation was adjusted to avoid issues with double precision causing the assertion error.

Added test for the case.

### DIFF
--- a/core/opendaq/reader/tests/test_multi_reader.cpp
+++ b/core/opendaq/reader/tests/test_multi_reader.cpp
@@ -955,6 +955,66 @@ TEST_F(MultiReaderTest, Clock10kHzDelta10)
     ASSERT_THAT(time[2], ElementsAreArray(time[0]));
 }
 
+TEST_F(MultiReaderTest, Clock15MHzFromEpoch)
+{
+    constexpr const auto NUM_SIGNALS = 3;
+
+    // prevent vector from re-allocating, so we have "stable" pointers
+    readSignals.reserve(3);
+
+    Int clock = 15000000;
+    Int startOffest = 1781269748ll * clock;
+    auto& sig0 = addSignal(startOffest, 6093750, createDomainSignal("1970-01-01T00:00:00+00:00", Ratio(1, clock)));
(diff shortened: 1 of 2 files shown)


core/opendaq/reader/tests/test_multi_reader.cpp
```diff
@@ -955,6 +955,66 @@ TEST_F(MultiReaderTest, Clock10kHzDelta10)
...
+
+    // prevent vector from re-allocating, so we have "stable" pointers
+    readSignals.reserve(3);
+
+    Int clock = 15000000;
+    Int startOffest = 1781269748ll * clock;
+    auto& sig0 = addSignal(startOffest, 6093750, createDomainSignal("1970-01-01T00:00:00+00:00", Ratio(1, clock)));
+    auto& sig1 = addSignal(startOffest, 2812500, createDomainSignal("1970-01-01T00:00:00+00:00", Ratio(1, clock)));
+    auto& sig2 = addSignal(startOffest, 3750000, createDomainSignal("1970-01-01T00:00:00+00:00", Ratio(1, clock)));
+
```
tomaz-cvetko, line 970: I would just like to understand this case better: what do these numbers achieve in terms of alignment via the comparable value? Do they all degenerate into the same value due to lacking double precision?

ankirn-ds: It is just real life use case we observed with our XHS device.

Start offset in 15M tick from linux epoch
packet size is just random number we typically see in packet size on 15M Samples/s acquisition.

Generally such big numbers results in the mantisa error causing the algoritm inside `typed_reader.cpp` fails - either `(target <= *comparablePreviousEndTick || target > *comparableEndingTick)` conditions is wrong or asserion error on `assert(index < packetSize && "Index out of bounds")` is thrown.

In release build the multireader stalls unsychronized or even throw the access violation exception.

tomaz-cvetko (approved):
Looks good, we noticed this as a potentially risky cast before.

PR #1233, opened 2026-06-18
ReadWithDomain should return Event status even in invalid

tomaz-cvetko:
# Brief

ReadWithDomain had a typo bug where invalid reader would not return events in the status even when the callback was triggered for this specific reason (overwriting a valid status with an empty one).

# Description

Refactor `read` and `readWithDomain` methods to delegate to a single `readInternal` where the logic is identical with exception of two operations that include domain buffer in `withDomain` version.

This does not change the usage of `Multireader` or its expected behavior.


JakaMohorko (approved):
Lgtm! Good catch.

PR #1236, opened 2026-06-19
Integration tests for Modern LT module

viacheslauK:
## Brief:
`getRecursiveConfigLock2()` to protect `items` field of the `FolderImpl<>` class;

Adds end-to-end integration test coverage for the Modern LT streaming module:
- New parameterized suites StreamingTestForModernLt and StreamingReconnectionTestForModernLt
- Test cases covering signal descriptor events, data packet delivery, multiple concurrent signals, getLastValue(), descriptor changes before subscribe, reconnection.
- New test helpers in test_helpers.h: packetsEqualWithComparators() and packetBehaviorComparison() for comparing packet streams with custom data/event/descriptor comparators.

### Notes
A few cases are guarded with GTEST_SKIP or DISABLED_. Read comments to find out the reasons.

### DIFF
--- a/core/opendaq/component/include/opendaq/folder_impl.h
+++ b/core/opendaq/component/include/opendaq/folder_impl.h
@@ -344,6 +344,12 @@ ErrCode INTERFACE_FUNC FolderImpl<Intf, Intfs...>::getSerializeId(ConstCharPtr* 
 template <class Intf, class ... Intfs>
 ErrCode FolderImpl<Intf, Intfs...>::enableCoreEventTrigger()
 {
+    // hold the recursive config lock while iterating "items"
+    // consistent with addItem()/removeItem()
+    // otherwise iteration can race with a concurrent structural change (e.g. a streaming client
+    // adding mirrored signals from a background thread), invalidating the container's iterators
+    // and corrupting the heap
+    auto lock = this->getRecursiveConfigLock2();
     for (const auto& [_, item] : items)
     {
         const ErrCode err = item.template asPtr<IPropertyObjectInternal>(true)->enableCoreEventTrigger();
@@ -356,6 +362,12 @@ ErrCode FolderImpl<Intf, Intfs...>::enableCoreEventTrigger()
 template <class Intf, class ... Intfs>
 ErrCode FolderImpl<Intf, Intfs...>::disableCoreEventTrigger()
 {
+    // hold the recursive config lock while iterating "items"
+    // consistent with addItem()/removeItem()
+    // otherwise iteration can race with a concurrent structural change (e.g. a streaming client
+    // adding mirrored signals from a background thread), invalidating the container's iterators
+    // and corrupting the heap
+    auto lock = this->getRecursiveConfigLock2();
     for (const auto& [_, item] : items)
     {
         const ErrCode err = item.template asPtr<IPropertyObjectInternal>(true)->disableCoreEventTrigger();
--- a/core/opendaq/component/include/opendaq/folder_impl.h
+++ b/core/opendaq/component/include/opendaq/folder_impl.h
@@ -175,6 +175,8 @@ ErrCode INTERFACE_FUNC FolderImpl<Intf, Intfs...>::isEmpty(Bool* empty)
 {
     OPENDAQ_PARAM_NOT_NULL(empty);
 
+    // hold the recursive config lock while accessing "items"
+    auto lock = this->getRecursiveConfigLock2();
     *empty = items.empty() ? True : False;
 
     return OPENDAQ_SUCCESS;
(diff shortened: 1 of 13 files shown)


tests/integration/test_opendaq_device_modules/test_streaming.cpp
```diff
@@ -839,3 +845,816 @@ TEST_F_UNSTABLE_SKIPPED(NativeDeviceStreamingTest, ChangedDataDescriptorBeforeSu
...
+        {
+            // LT streaming does not support post scaling, so it is expected to be default constructed on client side
+            // result &= transmitted.getPostScaling() == received.getPostScaling();
+            result &= !received.getPostScaling().assigned();
+        }
+        result &= transmitted.getStructFields() == received.getStructFields();
+        {
+            // LT streaming does not support metadata, so it is expected to be default constructed on client side
+            // result &= transmitted.getMetadata() == received.getMetadata();
+            result &= !received.getMetadata().assigned() || received.getMetadata().getCount() == 0u;
```
NikolaiShipilov, line 898: Just short observation regarding this - we use ws-streaming tag 3.0.7 but there is also 3.0.8 which has extra metadata support. There is not support for it in modules code, though.

tests/integration/test_opendaq_device_modules/test_streaming.cpp
```diff
@@ -839,3 +845,816 @@ TEST_F_UNSTABLE_SKIPPED(NativeDeviceStreamingTest, ChangedDataDescriptorBeforeSu
...
+                // LT streaming does not support post scaling, so it is expected to be default constructed on client side
+                // EXPECT_EQ(transmitted.getPostScaling(), received.getPostScaling());
+                EXPECT_FALSE(received.getPostScaling().assigned());
+            }
+            EXPECT_EQ(transmitted.getStructFields(), received.getStructFields());
+            {
+                // LT streaming does not support metadata, so it is expected to be default constructed on client side
+                // EXPECT_EQ(transmitted.getMetadata(), received.getMetadata());
+                EXPECT_TRUE(received.getMetadata().assigned() == false || received.getMetadata().getCount() == 0u);
+            }
```
NikolaiShipilov, line 931: Once modern implementation introduces support for extra metadata and postScaling the check will fail (+ the compare method above will return false) and tests will require modifications. I'm not sure if strict validation of these filds really makes much sense at this point - if just ignore them instead - the switching to upgraded modules versions will not break tests.

tests/integration/test_opendaq_device_modules/test_streaming.cpp
```diff
@@ -839,3 +845,816 @@ TEST_F_UNSTABLE_SKIPPED(NativeDeviceStreamingTest, ChangedDataDescriptorBeforeSu
...
+        if (!transmitted.assigned() && !received.assigned())
+            return true;
+
+        bool result = true;
+
+        auto eventPacketT = transmitted.asPtr<IEventPacket>(true);
+        auto eventPacketR = received.asPtr<IEventPacket>(true);
+
+        result &= eventPacketT.getEventId() == eventPacketR.getEventId();
+        result &= eventPacketT.getEventId() == event_packet_id::DATA_DESCRIPTOR_CHANGED;
```
NikolaiShipilov, line 954: If this condition is not met the following parameters' validation does not make sense and probably will throw an exception as parameters under such keys do not exist. If event id is not `DATA_DESCRIPTOR_CHANGED` then the function should just return false.
Though, the other types of event packet are not relevant for LT and should not appear.

tests/integration/test_opendaq_device_modules/test_streaming.cpp
viacheslauK: That is exactly the point. Once support for these fields is implemented, the test should fail, and that will give the developer a clear indication that the test needs to be adapted for the new implementation.
If we only validate the currently existing fields, a developer implementing new functionality may easily overlook these tests. In that case, the tests would only partially fulfill their purpose and could potentially mask an existing issue by creating the illusion of sufficient test coverage.

core/opendaq/component/include/opendaq/folder_impl.h
```diff
@@ -214,6 +216,12 @@ void FolderImpl<Intf, Intfs...>::syncComponentOperationMode(const ComponentPtr&
 template <class Intf, class ... Intfs>
 void FolderImpl<Intf, Intfs...>::notifyActiveChanged(bool onUpdate)
 {
+    // hold the recursive config lock while iterating "items"
+    // consistent with addItem()/removeItem()
+    // otherwise iteration can race with a concurrent structural change (e.g. a streaming client
+    // adding mirrored signals from a background thread), invalidating the container's iterators
+    // and corrupting the heap
```
JakaMohorko: These comments are repeated and too wordy for core implementation. Something like "Protect `items` from concurrent access." is sufficient.

JakaMohorko (approved):
Lgtm. Just a small request for the folder comments.

JakaMohorko (approved):

PR #1238, opened 2026-06-29
Try all streaming connection addresses if first fails (#1235)

NikolaiShipilov:
# Brief

Leaf device is connected to gateway device via IPv4 but streaming should go through IPv6 if gateway device is connected by IPv6

# Description

We have the following structure
 
 - GW device (opendaq device IPv4: 192.168.10.10)
 -- leaf device (opendaq device IPv4: 192.168.10.11)
 -- leaf device (opendaq device IPv4: 192.168.10.12)
 
 GW device add subdevice via IPv4 connection string.
 
 When client (PC) tries to connect to GW device and IPv4 is connection is not available (i.e. different subnet) the connection to the GW is established by IPv6. Leafs are running the discovery server so streaming does get the information about possible streaming connections but somehow only "deviceConnectionAddress" is used as possible - which in this case is IPv4 (since GW device is connected to the subdevice via IPv4) and of curse not possible.

### DIFF
--- a/tests/integration/test_opendaq_device_modules/test_native_device_modules.cpp
+++ b/tests/integration/test_opendaq_device_modules/test_native_device_modules.cpp
@@ -2832,6 +2832,225 @@ TEST_F(NativeDeviceModulesTest, TestAddressInfoGatewayDevice)
     ASSERT_EQ(ltAddressInfo.getReachabilityStatus(), AddressReachabilityStatus::Reachable);
 }
 
+TEST_F(NativeDeviceModulesTest, TestLeafDeviceUnreachableIPv4Streaming)
+{
+    SKIP_TEST_MAC_CI;
+    if (test_helpers::Ipv6IsDisabled())
+    {
+        GTEST_SKIP();
+    }
+
+    auto server = Instance("[[none]]");
+    {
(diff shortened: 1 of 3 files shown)


tests/integration/test_opendaq_device_modules/test_native_device_modules.cpp
```diff
@@ -2832,6 +2832,225 @@ TEST_F(NativeDeviceModulesTest, TestAddressInfoGatewayDevice)
...
+        serverConfig.setPropertyValue("NativeStreamingPort", 7421);
+
+        gateway.addServer("OpenDAQNativeStreaming", serverConfig);
+    }
+
+    auto client = Instance("[[none]]");
+    addNativeClientModule(client);
+
+    // register handler before adding the device so it will be called before the one in streaming manager
+    client.getContext().getOnCoreEvent() += [](const ComponentPtr& comp, const CoreEventArgsPtr& args) {
```
denise-opendaq, line 2976: maybe make a method replace server cap with fake address, to reduce the code duplication?

denise-opendaq (approved):

JakaMohorko (approved):

PR #1239, opened 2026-07-04
Document approaches for generating bindings for other languages

adolenc:
# Brief

Quick draft of the documentation page for what it takes to create language bindings for openDAQ, which mechanisms are available to do it, and what the bindings need to cover or be careful with. 

Full disclosure, this was created with the help of claude.


docs/Antora/modules/explanations/pages/language_bindings.adoc
```diff
@@ -0,0 +1,227 @@
...
+
+* *Reference counting.* Objects are created with a reference count of one, owned by the caller of the factory.
+  Out-parameters of getters likewise return a reference the caller owns and must eventually release. Plain
+  in-parameters are borrowed -- do not release them, and add-ref before storing them beyond the call. Garbage-collected
+  languages typically attach a finalizer that calls `releaseRef` exactly once per owned reference.
+* *Errors.* Every `ErrCode` return must be checked and mapped to the language's error mechanism. Extended error
+  information for the calling thread can be retrieved through the error-info API (`daqGetErrorInfoMessage` in the C
+  API) before it is overwritten by the next failing call.
+* *Strings.* `IString`'s character pointer (`getCharPtr`) borrows storage owned by the string object; copy it into a
+  native string before releasing the reference.
```
JakaMohorko: This is something common to all core type objects. These are immutable and should be copied and converted to a native type before release. Basic data types available on the C-ABI are usually exposed as such on the openDAQ interfaces, but additional wrappers exist that are mainly used in calls using polymorphic arguments:

The call 
```
setPropertyValue(IString* name, IBaseObject* value)
```
when changing a `Integer` property value, requires a basic integer type to be boxed into an `IInteger` value. The  correlated getter functions in a similar manner.

docs/Antora/modules/explanations/pages/language_bindings.adoc
```diff
@@ -0,0 +1,227 @@
...
+  propagate through the native frames is undefined behavior. Note also that openDAQ(TM) may invoke callbacks from
+  its own worker threads, so the binding's callback entry points must be safe to enter from threads the language
+  runtime did not create.
+
+Implementing a whole interface from the binding language is the same mechanism scaled up: the binding must construct
+a complete virtual table -- including `queryInterface`, `addRef`, and `releaseRef`, i.e. reference counting
+implemented on the binding side -- and hand out a pointer to a structure whose first word points to it. Few bindings
+need this; when only callables and event handlers are required, the function-pointer route above is far simpler.
+
+== Rules every binding must respect
```
JakaMohorko: This section seems to outline a set of openDAQ features that are arbitrarily chosen. Whereas the rest of the document is focused on extensive technical details of bindings and different approaches to interfacing with openDAQ, this section briefly touches upon a few select topics that are parts of larger sections.

Namely:
- The `Strings` section applies to all core types. In the context of binding creation, the base step of implementation is converting all openDAQ core types to ones native to the language (`IInteger`, `IString`, `IFloat`...)
- *Sample buffers* are another key part of writing bindings that differ between languages. Here it's critical to take an approach that's best suited for a language. Data packet buffers are the only object in openDAQ that's accessible via a raw `void*` pointer.
- Polymorphism should be mentioned here. Some openDAQ arguments can accept a range of different types of which validity is checked during runtime and depend on the current system state. As with buffers, the typing rules of the target binding language determine how polymorphic arguments should be handled.

JakaMohorko (commented):
Looks like a really good overview of the technicalities about binding generation! 

It's a bit technical and should likely be reworked in future documentation iterations for the article to have clear learning outcomes. As of now, it adds additional information and context to binding creation that was previously missing. Nevertheless, I believe it to be a good addition to the current openDAQ knowledge base.

I would, however, look to rename it to something similar to "Language binding internals" and provide a 2-3 bullet point summary of contents at the top of the document for readers to know whether this document actually contains information they require. As of now, openDAQ users who are only looking to use languages other than C++ might try to read this document, but they are not its target audience.

docs/Antora/modules/explanations/pages/language_bindings.adoc
adolencds: I am not sure I would call these features arbitrary, I think there needs to be a section dedicated to additional things that every language will need to handle in some way or another and didn't fit anywhere else. 

Could I please ask you clarify what you mean with the sample buffers comment, how does it differ from the final bullet point in this section? Readers and data packets will need special treatment to have void* work nicely.

Could you also please clarify about the polymorphism? From the point of view of the language bindings, the arguments should match what the interfaces expect, or am I wrong? Are you referring to the comment above about setPropertyValue?

JakaMohorko: Arbitrary was not the correct choice of words - English bindings are sometimes hard. 

These are all topics one must be aware of when generating openDAQ bindings, not specifically rules. Labelling the section as rule reads to me that this is a finite ruleset for openDAQ binding generation.

Regarding polymorphism - a set of function calls accept an `IBaseObject` or another abstract interface. It's usually not as generic to accept *any* interface; it's often limited to a few. Which ones are accepted can sometimes be deduced at compile time and sometimes at runtime. These calls include `IPropertyObject::get/setPropertyValue`, any method that returns `INumber`, `IProperty::getSelectionValues`, `IContext::getModuleOptions`... Depending on whether a language is dynamically/statically typed and whether we want to have strong typing for the input arguments of these methods, these functions need to be approached in different manners.

Re-reading the document, the sample buffers section contains the information required. My original comment did not convey my thoughts. It was mainly that I'm not sure that conversion to the language's natural array types is the main approach/wording to use here. I think all bindings should provide raw access to the data if allowed by the language, providing conversion options into natural array types as additional helpers. I'd say that packet buffers should ideally always allow for raw access, while reader wrappers should always have the option to convert into natural types.  

I think all of the topics in this part warrant their own section. That's not necessary at the moment, but I would potentially rename the section to something similar to "OpenDAQ characteristics relevant to binding generation".

PR #1240, opened 2026-07-06
Fix reading of multi-dimensional (vector/matrix) signals in typed reader

andrejsusnik-dewesoft:
#Brief

Fixes reading of multi-dimensional (vector/matrix) signals in TypedReader. Previously the reader only accounted for values-per-sample on single-dimension signals, so matrix signals (2+ dimensions) computed the wrong number of values per sample and advanced the output buffer pointer incorrectly during type-converting reads — corrupting data on multi-block reads. The reader now derives values-per-sample from the product of all dimension sizes and advances the output pointer accordingly.

#Description

- Compute values-per-sample across all dimensions (handleDescriptorChanged): The descriptor handler previously set valuesPerSample only when a signal had exactly one dimension, reading dimensions[0].getSize(). It now handles any signal with one or more dimensions by multiplying the sizes of all dimensions together, so matrix signals (e.g. an N×M sample) report the correct total values per sample.
- Fix output-buffer pointer advance in type-converting reads (readValues): After copying with a type conversion, the output buffer pointer was advanced by toRead (sample count) instead of toRead * valuesPerSample. For multi-dimensional signals this left the pointer short, so subsequent reads overwrote or misaligned previously read data. Corrected in both the generic TypedReader<TReadType>::readValues path and the TypedReader<ClockTick>::readValues<ClockRange> specialization.

### DIFF
--- a/core/opendaq/reader/src/typed_reader.cpp
+++ b/core/opendaq/reader/src/typed_reader.cpp
@@ -733,9 +733,15 @@ bool TypedReader<ReadType>::handleDescriptorChanged(DataDescriptorPtr& descripto
 
         rawSampleSize = descriptor.getRawSampleSize();
         auto dimensions = descriptor.getDimensions();
-        if (dimensions.assigned() && dimensions.getCount() == 1)
+        if (dimensions.assigned() && dimensions.getCount() >= 1)
         {
-            valuesPerSample = dimensions[0].getSize();
+            SizeT count = 1;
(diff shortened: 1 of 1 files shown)


tomaz-cvetko (approved):
Looks good to me, this generalises values per sample to tensor of arbitrary dimensionality as far as typed reading is concerned.

JakaMohorko (approved):

PR #1241, opened 2026-07-09
Implement IDevice::getDomainSignal()

denise-opendaq:
# Brief

Implement IDevice::getDomainSignal()

# Description

The device now considers the signal with tag `DeviceDomain` as a domain signal of this device.  at that case, the method Device::getTicksSinceOrigin() will automatically return the signal value by calling domainSignal.getLastValue();

# Usage example

In C++ use:

```cpp
class SomeDeviceImpl::Device
{
SomeDeviceImpl(const ContextPtr& ctx,
                          const ComponentPtr& parent,
                          const StringPtr& localId)
  : Device(ctx, parent, localId)
{
    createSignals();
}

void createSignals()
{
    auto timeSignal = createAndAddSignal("Time", nullptr, false);
    timeSignal.getTags().asPtr<ITagsPrivate>(true).add("DeviceDomain");
}
};

// outside from device
// get the domain signal
auto deviceDomainSignal = device.getDomainSignal();

// get the last domain signal value
auto lastValue = device.getTicksSinceOrigin();
// which is the same as 
auto lastValueAnotherWay = deviceDomainSignal.getLastValue();
```

# API changes

```diff
+ IDevice::getDomainSignal(ISignal** signal);
```

None
But if device has the domain signal with correct domain , the overridden method `onGetTicksSinceOrigin` can be removed inside of your Implementation

# Required module changes

None

### DIFF
--- a/core/opendaq/device/include/opendaq/device_impl.h
+++ b/core/opendaq/device/include/opendaq/device_impl.h
@@ -124,6 +124,8 @@ public:
     ErrCode INTERFACE_FUNC setOperationModeRecursive(OperationModeType modeType) override;
     ErrCode INTERFACE_FUNC getOperationMode(OperationModeType* modeType) override;
 
+    ErrCode INTERFACE_FUNC getDomainSignal(ISignal** signal) override;
+
     // IDevicePrivate
     ErrCode INTERFACE_FUNC setAsRoot() override;
     ErrCode INTERFACE_FUNC getDeviceConfig(IPropertyObject** config) override;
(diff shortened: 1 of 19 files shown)


core/opendaq/device/include/opendaq/device_impl.h
```diff
@@ -1259,6 +1269,20 @@ ErrCode GenericDevice<TInterface, Interfaces...>::getOperationMode(OperationMode
...
+    OPENDAQ_PARAM_NOT_NULL(signal);
+    *signal = this->domainSignal.addRefAndReturn();
+    return OPENDAQ_SUCCESS;
+}
+
+template <typename TInterface, typename... Interfaces>
+void GenericDevice<TInterface, Interfaces...>::setDomainSignal(const SignalPtr& signal)
+{
+    this->domainSignal = signal;
+}
```
JakaMohorko: I think we might be able to avoid requiring to set the domain signal. This could be designed in a manner where the getter finds a device signal with the tag `DeviceDomain`.

denise-opendaq: what if device for some reason have multiple signals with tag `DeviceDomain`

JakaMohorko: This should be a modelling rule that we provide as a standard for development. The `DeviceDomain` signal is a way to represent the device's clock. The only reason to have multiple is if the device had multiple clocks - our current API, however, does not support that. A device can only have 1 clock.

PR #1244, opened 2026-07-13
Implement static objects and object pool

dejancrn:
# Brief

Static objects and object pool/cache implementation

# Description

This PR features two distinct but related concepts:

1) Cached objects (Static objects)
----------------------------------------------
The name "Cached" is misleading. They are actually static objects created at startup. These are some bool (True, False), int (0, 1, 2, ...), float (0.0, 1.0) objects that are used very often. Instead of creating them every time, only the reference count is increased. It is worth considering adding float and int values that are used very often (10.0, 100.0, 0.1,...).

2.) Pooled objects
----------------------------------------------
Pooled (cached) objects are ints and floats that are created very often. Instead of creating them every time, when an object is being freed, it will be returned to the pool and reused the next time a new object is required. This optimization is more invasive than the first one. Therefore, it is disabled by default and must be enabled with OPENDAQ_ENABLE_OBJECT_POOLS. The pool size is initialized to 100 and will never shrink, but it may grow. 
Possible upgrades include optional runtime pool size management and support for other types of objects (e.g., packets?).

Both optimizations heavily decrease the number of memory allocations, especially on the configuration path. The second optimization uses a boost lock-free stack, and its performance gains depend on the application and platform. Therefore, it is recommended to thoroughly profile the app and SDK.


JakaMohorko (commented):
This looks good to me and a good addition. It can be extended in the future, but is already a large improvement as is.

This PR is however missing:
- A changelog entry
- The CMake flag OEPNDAQ_ENABLE_OBJECT_POOLS should be enabled in the CI preset.

JakaMohorko (approved):

PR #1245, opened 2026-07-13
Fix Linux mingw cross-build issues

lvanluik:
# Brief

Fixes cross-compilation for Windows on Linux/mingw.

# Description

We are cross-compiling OpenDAQ for Windows on our Linux build machine using mingw. These are the changes I had to make to get the build working, mainly due to case sensitivity on Linux and one CMake issue where the target platform was used instead of the host platform.


JakaMohorko (approved):
Lgtm, thank you for the patch!

PR #1255, opened 2026-07-28
Export generatePacketId() from the opendaq shared library

dejancrn:
# Brief

Rename generatePacketId() to daqGeneratePacketId() and make it accessible from outside opendaq.dll/.so

# Description

Resolves the issue when DataPacketImpl is used to implement packets outside the opendaq shared library.

generatePacketId() is reached from GenericDataPacketImpl's constructor and from DataPacketImpl::reuse(), both of which are defined in headers. The symbol itself was not exported, so on Windows no DataPacketImpl subclass outside the opendaq DLL could link, as PUBLIC_EXPORT resolves to dllimport for consumers.

Exporting the SDK's single monotonic id source keeps packet ids unique process-wide; a counter local to each consumer would hand out duplicates.

### DIFF
--- a/core/opendaq/signal/include/opendaq/data_packet_impl.h
+++ b/core/opendaq/signal/include/opendaq/data_packet_impl.h
@@ -808,7 +808,7 @@ ErrCode DataPacketImpl<TInterface, TInterfaces...>::reuse(IDataDescriptor* newDe
         this->callDestructCallbacks();
     }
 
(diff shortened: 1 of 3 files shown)


tomaz-cvetko (approved):
Looks good to me, fixes linkage outside the opendaq dll.

JakaMohorko (approved):

PR #1257, opened 2026-07-28
Fix stale computed-data cache in DataPacketImpl

dejancrn:
# Brief

Fix data packet scaling cache that references old data when the packet is reused.

# Description

getData() used the scaledData pointer itself as its cache-validity flag, so the buffer could not be kept without also being treated as current. Three defects followed from that:

- reuse() left scaledData in place on its non-reallocating path, so a recycled packet kept returning the values computed for its previous contents.

- reuse() never re-derived hasScalingCalc, hasDataRuleCalc, hasRawDataOnly or hasReferenceDomainOffset after replacing the descriptor, so those flags went on describing the old one. The stale buffer masked this by being returned without recomputation, which means fixing only the first defect would have turned stale data into a call to a calculator the new descriptor does not have.

- getData() returned through OPENDAQ_RETURN_IF_FAILED while still holding readLock, leaving the packet permanently locked after any calculator failure.

Validity is now tracked by a separate scaledDataValid flag, with scaledDataCapacity recording the size of the allocation, and the calculators are handed the buffer to write into instead of allocating their own. Recomputing a packet's data therefore no longer allocates. reuse() invalidates the cache and refreshes the derived flags through updateCalcFlags(), split out of initPacket(). getData() takes readLock with std::lock_guard.

The first two defects are covered by new tests that fail without this change. The lock leak is not covered: provoking it needs a calculator failure that cannot be produced through the public API.

### DIFF
--- a/core/opendaq/signal/include/opendaq/data_packet_impl.h
+++ b/core/opendaq/signal/include/opendaq/data_packet_impl.h
@@ -317,6 +317,7 @@ protected:
     void freeMemory();
     void freeScaledData();
     void initPacket();
+    void updateCalcFlags();
 
     DeleterPtr deleter;
     DataDescriptorPtr descriptor;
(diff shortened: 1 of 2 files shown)


tomaz-cvetko (approved):
Looks good, addresses the described issues and adds relevant tests.

PR #1258, opened 2026-07-29
Make device info the nested device property

denise-opendaq:
# Brief

`DeviceInfo` is now a core property of the device.

# Description

Implement the `PropertyObjectImpl::addCoreProperty` method. It behaves like the regular `addProperty` method, but stores the property's original default value instead of cloning it.
Core properties are also excluded from the result returned by `getAllProperties`.
`DeviceInfo` is now registered using the new `addCoreProperty` method instead of being stored as a device member.

# API changes

None

None

# Required module changes

None


core/opendaq/device/include/opendaq/device_info_impl.h
```diff
@@ -1220,6 +1246,38 @@ void DeviceInfoConfigImpl<TInterface, Interfaces...>::triggerCoreEventMethod(con
...
+
+    Super::InterfaceIds::AddInterfaceIds(*ids);
+    return OPENDAQ_SUCCESS;
+}
+
+template <typename TInterface, typename ... Interfaces>
+ErrCode DeviceInfoConfigImpl<TInterface, Interfaces...>::freeze()
+{
+    return OPENDAQ_IGNORED;
+}
```
JakaMohorko: Why are we not allowing freezing of the device info?

external_modules/CMakeLists.txt
```diff
@@ -68,7 +68,7 @@ if (OPENDAQ_ENABLE_OPCUA AND (DAQMODULES_OPENDAQ_CLIENT_MODULE OR DAQMODULES_OPE
     FetchContent_Declare(
         OpcUaModules
         GIT_REPOSITORY https://github.com/openDAQ/OpcUaModules.git
-        GIT_TAG        v4.0.1
```
JakaMohorko, line 71: Is this a breaking change? Is it possible to still compile the OPC UA modules without the newer core version, but an older OPC UA one?

core/coreobjects/include/coreobjects/property_object_impl.h
```diff
@@ -395,7 +399,7 @@ class GenericPropertyObjectImpl : public ImplementationOfWeak<PropObjInterface,
 
     // Gets the property, as well as its value. Gets the referenced property, if the property is a refProp
     ErrCode getPropertyAndValueInternal(const StringPtr& name, BaseObjectPtr& value, PropertyPtr& property, bool triggerEvent = true, bool retrieveUpdatingValue = false);
-    ErrCode getPropertiesInternal(Bool includeInvisible, Bool bind, IList** list);
+    ErrCode getPropertiesInternal(Bool includeInvisible, Bool bind, IList** list, Bool includeCoreProperties = true);
```
JakaMohorko: This should probably use `includeCoreProperties = false` to preserve previous behaviour. Currently, this changes the behaviour of `findProperties`.

core/coreobjects/include/coreobjects/property_object_impl.h
JakaMohorko, line 1826: This function probably needs to also clone `corePropertyNames`.

core/coreobjects/include/coreobjects/property_object_impl.h
```diff
@@ -2329,14 +2346,22 @@ ErrCode GenericPropertyObjectImpl<PropObjInterface, Interfaces...>::addProperty(
...
+                ErrCode err = cloneable->clone(&clone);
+                OPENDAQ_RETURN_IF_FAILED(err);
+
+                if (defaultValue.getObject() != clone.getObject())
+                    propInternalPtr.overrideDefaultValue(clone);
+            }
+            else 
+            {
+                corePropertyNames.insert(propName);
+            }
```
JakaMohorko: We only allow object-type core properties like this, but we can still call `addCoreProperty` with a different property type. We should either allow all, or return an early error if we try adding non `ctObject` props.

core/coreobjects/include/coreobjects/property_object_impl.h
```diff
@@ -2371,6 +2412,9 @@ ErrCode GenericPropertyObjectImpl<PropObjInterface, Interfaces...>::removeProper
         propValues.erase(propertyName);
     }
 
+    if (auto it = corePropertyNames.find(namePtr); it != corePropertyNames.end())
+        corePropertyNames.erase(it);
```
JakaMohorko, line 2421: Do we want to allow removing core properties from the outside?

core/coreobjects/include/coreobjects/property_object_impl.h
```diff
@@ -2090,6 +2099,11 @@ ErrCode GenericPropertyObjectImpl<PropObjInterface, Interfaces...>::clearPropert
             {
                 if (auto it = propValues.find(prop.getName()); it->second.assigned())
                 {
+                    // Match clearPropertyValuesInternal: leave frozen nested objects intact.
+                    if (const auto freezable = it->second.template asPtrOrNull<IFreezable>(true);
+                        freezable.assigned() && freezable.isFrozen())
+                        return OPENDAQ_IGNORED;
```
JakaMohorko, line 2111: Not sure if this is correct behaviour. For `clearPropertyValues` it's a bit different, as we iterate through the list of nested properties and we'll anyhow end up in a dirty state if we return an error. In that case we should ideally log all objects that were frozen and skip them.

Here I think it's more valid to return an error.

core/opendaq/opendaq/src/client_impl.cpp
```diff
@@ -14,7 +14,11 @@ ClientImpl::ClientImpl(const ContextPtr ctx, const StringPtr& localId, const Dev
                           : throw ArgumentNullException("Logger must not be null"))
 {
     this->name = "OpenDAQClient";
-    this->deviceInfo = deviceInfo;
+
+    if (const auto devInfoPtr = deviceInfo.asPtrOrNull<IDeviceInfoConfig>(); devInfoPtr.assigned())
+        devInfoPtr.setConnectionString("daq.root://default_client");
```
JakaMohorko: This is a semi-large change; it changes the default connection string of the root device. Is there a good reason for this update?

core/opendaq/device/include/opendaq/device_info_impl.h
```diff
@@ -1220,6 +1246,38 @@ void DeviceInfoConfigImpl<TInterface, Interfaces...>::triggerCoreEventMethod(con
...
+    {
+        return OPENDAQ_SUCCESS;
+    }
+
+    **ids = IPropertyObject::Id;
+    (*ids)++;
+
+    Super::InterfaceIds::AddInterfaceIds(*ids);
+    return OPENDAQ_SUCCESS;
+}
```
JakaMohorko, line 1220: I wonder if we could work around this constraint inside the property object implementation if we have well-known core object names now. This is just used to circumvent the `ctObject` check for the main interface.

core/coreobjects/include/coreobjects/property_object_impl.h
```diff
@@ -332,6 +335,7 @@ class GenericPropertyObjectImpl : public ImplementationOfWeak<PropObjInterface,
 
     std::unordered_map<StringPtr, BaseObjectPtr, StringHash, StringEqualTo> propValues;
     PropertyOrderedMap localProperties;
+    std::set<std::string> corePropertyNames;
```
JakaMohorko: Might be worth using `StringPtr` here to avoid additional heap allocations on conversion to openDAQ objects.

core/opendaq/device/include/opendaq/device_info_impl.h
```diff
@@ -1220,6 +1246,38 @@ void DeviceInfoConfigImpl<TInterface, Interfaces...>::triggerCoreEventMethod(con
     }
 }
 
+template <typename TInterface, typename ... Interfaces>
+ErrCode DeviceInfoConfigImpl<TInterface, Interfaces...>::clone(IPropertyObject** cloned)
+{
+    OPENDAQ_PARAM_NOT_NULL(cloned);
+    *cloned = this->template thisInterface<IPropertyObject>();
+    return OPENDAQ_SUCCESS;
+}
```
JakaMohorko, line 1202: Do we really want to return `this` inside of clone? This is against the role of the function. Could we just avoid calling `clone` on device info objects?

core/opendaq/device/include/opendaq/device_info_impl.h
```diff
@@ -1124,7 +1139,14 @@ ErrCode DeviceInfoConfigImpl<TInterface, Interfaces...>::setPropertyValueNoLock(
...
     auto owner = getOwnerOfProperty(propertyName);
     if (owner.assigned())
+    {
+        // Mirrored/config-client devices must not be written via setPropertyValue here:
+        // that always issues an RPC and deadlocks when DeviceInfo is applied during
+        // ComponentUpdateEnd / nested remoteUpdate (server still in SendOutCoreEvents).
+        if (owner.template supportsInterface<IMirroredDevice>())
+            return owner.template asPtr<IPropertyObjectProtected>(true)->setProtectedPropertyValue(propertyName, value);
         return owner->setPropertyValue(propertyName, value);
+    }
```
JakaMohorko: Couldn't this be checked via a flag that's true while remote updating (`deserializationComplete` or similar)? `setProtectedPropertyValue` is also an RPC, or?

This is special handling for `location` and `userName` if I'm not mistaken. I'd just assume that we should skip those when updating.

core/coreobjects/include/coreobjects/property_object_impl.h
```diff
@@ -2347,6 +2372,22 @@ ErrCode GenericPropertyObjectImpl<PropObjInterface, Interfaces...>::addProperty(
...
+    OPENDAQ_RETURN_IF_FAILED(errCode);
+    return errCode;
+}
+
+template <typename PropObjInterface, typename... Interfaces>
+ErrCode GenericPropertyObjectImpl<PropObjInterface, Interfaces...>::addCoreProperty(IProperty* property)
+{
+    const ErrCode errCode = addPropertyInternal(property, true);
+    OPENDAQ_RETURN_IF_FAILED(errCode);
+    return errCode;
```
JakaMohorko, line 2393:
```suggestion
    OPENDAQ_RETURN_IF_FAILED(addPropertyInternal(property, true));
    return OPENDAQ_SUCCESS;
```

core/coreobjects/include/coreobjects/property_object_impl.h
```diff
@@ -2347,6 +2372,22 @@ ErrCode GenericPropertyObjectImpl<PropObjInterface, Interfaces...>::addProperty(
     return errCode;
 }
 
+template <class PropObjInterface, class... Interfaces>
+ErrCode GenericPropertyObjectImpl<PropObjInterface, Interfaces...>::addProperty(IProperty* property)
+{
+    const ErrCode errCode = addPropertyInternal(property);
+    OPENDAQ_RETURN_IF_FAILED(errCode);
+    return errCode;
```
JakaMohorko, line 2385:
```suggestion
    OPENDAQ_RETURN_IF_FAILED(addPropertyInternal(property));
    return OPENDAQ_SUCCESS;
```

JakaMohorko (changes requested):

core/opendaq/device/include/opendaq/device_impl.h
```diff
@@ -251,6 +250,8 @@ class GenericDevice : public SignalContainerImpl<TInterface, IDevicePrivate, IDe
     DevicePtr getParentDevice();
     OperationModeType getOperationMode();
 
+    ErrCode setDeviceInfo(const DeviceInfoPtr& deviceInfo);
```
JakaMohorko: This is a breaking change that breaks any existing openDAQ device implementation. Is there a ways to work around this? Could we have change the `deviceInfo` type to a lambda that calls `setDeviceInfo`? If nothing else, the `deviceInfo` object should be moved to the private scope - right now it will just fail silently; it should be caught at compile time.

shared/libraries/config_protocol/include/config_protocol/config_server_component.h
```diff
@@ -394,17 +374,6 @@ inline void ConfigServerComponent::applyProps(uint16_t protocolVersion, const Pr
     }
 }
 
-inline void ConfigServerComponent::parseAndGetDeviceInfo(PropertyObjectPtr& component, std::string& propName)
```
JakaMohorko, line 397: A version bump is needed. This broke forward compatibility (new server -> old client)

core/opendaq/device/include/opendaq/device_impl.h
```diff
@@ -2036,27 +2069,23 @@ void GenericDevice<TInterface, Interfaces...>::serializeCustomObjectValues(const
...
         serializer.key("deviceInfo");
-        if (forUpdate)
-        {
-            deviceInfo.template asPtr<IUpdatable>(true).serializeForUpdate(serializer);
-        }
-        else
-        {
-            deviceInfo.serialize(serializer);
-        }
+        deviceInfo.serialize(serializer);
```
JakaMohorko: Don't we end up double-serializing `deviceInfo` here? We should be able to remove this serialization, but keep backwards compatibility on update/deserialize?

Actually no, we need to handle the serialization for update as well, as we need to also be able to update 3.40 or older devices. This likely requires a new serializer version.

shared/libraries/config_protocol/include/config_protocol/config_client_device_impl.h
```diff
@@ -405,6 +407,32 @@ ErrCode GenericConfigClientDeviceImpl<TDeviceBase>::setAsRoot()
...
+template <class TDeviceBase>
+ErrCode GenericConfigClientDeviceImpl<TDeviceBase>::getInfo(IDeviceInfo** info)
+{
+    OPENDAQ_PARAM_NOT_NULL(info);
+
+    if (this->isComponentRemoved)
+        return DAQ_MAKE_ERROR_INFO(OPENDAQ_ERR_COMPONENT_REMOVED);
+
+    // Remote config client: DeviceInfo lives in the DaqDeviceInfo property (deserialized / updated from server).
+    this->deviceInfo = this->objPtr.getPropertyValue("DaqDeviceInfo").template asPtrOrNull<IDeviceInfo>(true);
```
JakaMohorko: Is there a chance for the property not to exist? This will throw an exception in this case.

We should at least use the interface `getPropertyValue` method and return an error in that case.

core/coreobjects/include/coreobjects/property_object_impl.h
denise-opendaq: if we return the error, we will be in a muffle state without continue or restore the state

core/coreobjects/include/coreobjects/property_object_impl.h
denise-opendaq: what if for some reason the error Code will be OPENDAQ_IFNORED or another not failed error code?

core/coreobjects/include/coreobjects/property_object_impl.h
denise-opendaq: in the perfect life not. but if somebody would create core property. then remove it. and then create a normal property, the prop obj will classify it as a core property which is wrong I think.

core/opendaq/device/include/opendaq/device_impl.h
denise-opendaq: I don't think we have to deviceInfo.template asPtr<IUpdatable>(true).serializeForUpdate(serializer); 
I left deserialize to make posiblr read config from older version of the device

core/opendaq/device/include/opendaq/device_info_impl.h
denise-opendaq: because Property constructor tries to freeze everything which has IFreezable interface

shared/libraries/config_protocol/include/config_protocol/config_client_device_impl.h
denise-opendaq: we are creating this property always on creating the device

external_modules/CMakeLists.txt
denise-opendaq: its just fixing the tests (the test were failing because another amount of the properties )

JakaMohorko: I think that's valid, I would just include it in a separate PR - merge the OPC UA branch, then open a PR to main.

core/opendaq/device/include/opendaq/device_impl.h
JakaMohorko: I think here it's more that it'll be serialized as a property value as part of the `PropertyObjectImpl` serialization. We should either filter it out there, or remove the serialization here.

My concern is that we need to serialize it under a specific key for older versions, so we can't use the `PropertyObjectImpl` one anyhow (or we need to update the serializer version).

shared/libraries/config_protocol/include/config_protocol/config_client_device_impl.h
JakaMohorko: Will we use the same approach for `ISynchronization`? I think those will behave in the same manner. I suppose the questions are:
- Can this field change more than once? If no, it'd make sense to add the property the moment `DeviceInfo` is  created and not earlier and avoid calling `setPropertyValueProtected`.
- Will `ISynchronization` always be available? Some devices without their own clock will not have the `ClockSyncInterface`. In this case, we either have no interfaces, or the `daqSynchronization` property is not at all added.

external_modules/CMakeLists.txt
denise-opendaq: there is already opened pr https://github.com/openDAQ/OpcUaModules/pull/33

JakaMohorko (approved):

PR #1261, opened 2026-08-03
Track pooled objects only while they are handed out

dejancrn:
# Brief

Track pooled objects only while they are handed out.

# Description

ObjectPool never destroys its objects: at refcount 0 PooledObject::releaseRef() pushes the object onto the free list instead of deleting it. Debug object tracking registers in GenericObjInstance's constructor and unregisters in its destructor, so a pooled object stayed tracked for the lifetime of the process.

Every object a pool had to construct to meet peak demand was therefore reported as a leak by DaqMemCheckListener, blaming whichever test raised the pool's high-water mark past the 100 objects each static pool pre-allocates. The dump gives it away: every entry stringifies as a bare number and values repeat, because free-list residents keep whatever value they were last assigned.

Bind tracking to the period an object is handed out instead, which is what the leak checks mean by "alive": marked live when the pool resets it for a caller, parked again at construction and when its last reference goes. Objects sitting in the free list are no longer tracked, while one that leaks while handed out stays detectable.

The hooks live in PooledObject rather than in ObjectPool because test_object_pool.cpp instantiates the pool with a plain class that is not an openDAQ object and has no tracking to bind.

### DIFF
--- a/core/coretypes/include/coretypes/pooled_object.h
+++ b/core/coretypes/include/coretypes/pooled_object.h
@@ -24,13 +24,13 @@ namespace daq::object_pool
 {
 
 /*!
- * Debug object tracking normally spans construction to destruction, but a pooled object is
- * constructed once and then recycled indefinitely: it is never destroyed while the pool lives, so
- * it would stay tracked for the whole process. Every object the pool has to construct to meet peak
- * demand would then be reported as a leak by the first test that pushed the demand up, even though
- * nothing leaked. Tracking is therefore bound to the period the object is handed out, which is what
- * the leak checks mean by "alive": marked live when the pool resets it for a caller, and parked
- * again when its last reference goes. Objects sitting in the free list are not tracked.
+ * @brief An object owned by an ObjectPool and recycled rather than destroyed.
+ *
+ * Reaching reference count 0 returns the object to the pool's free list instead of deleting it, so a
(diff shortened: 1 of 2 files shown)


core/coretypes/include/coretypes/pooled_object.h
```diff
@@ -18,10 +18,20 @@
...
 
+/*!
+ * Debug object tracking normally spans construction to destruction, but a pooled object is
+ * constructed once and then recycled indefinitely: it is never destroyed while the pool lives, so
+ * it would stay tracked for the whole process. Every object the pool has to construct to meet peak
+ * demand would then be reported as a leak by the first test that pushed the demand up, even though
+ * nothing leaked. Tracking is therefore bound to the period the object is handed out, which is what
+ * the leak checks mean by "alive": marked live when the pool resets it for a caller, and parked
+ * again when its last reference goes. Objects sitting in the free list are not tracked.
+ */
```
JakaMohorko, line 34: These comments are a bit too verbose - it should be a 1-sentence indicator of the issue at hand.

JakaMohorko (approved):

core/coretypes/include/coretypes/pooled_object.h
dejancrn: Agreed, it actually describes the issue, not the intent of the class. I will update.

PR #1262, opened 2026-08-03
New capability API and discovery mechanism

viacheslauK:
# Brief

Adds `ProtocolGroupId` and `ProtocolSecurityLevel` to `IServerCapability`/`IServerCapabilityConfig` and `getProtocolGroupId` to `IStreaming`, so that protocols which are variants of one another (e.g. plain and TLS-secured transports of the same protocol) can be grouped — `StreamingSourceManager` then attaches only the single most preferred streaming source per group instead of one per protocol. Also allows a single server to advertise more than one discovery service.

# Description

**Server capability API**

- `IServerCapability::getProtocolGroupId` / `IServerCapabilityConfig::setProtocolGroupId` — an empty group ID means the protocol belongs to no group. Backed by a new `ProtocolGroupId` string property, default `""`.
- `IServerCapability::getProtocolSecurityLevel` / `IServerCapabilityConfig::setProtocolSecurityLevel` — backed by a new `SecurityLevel` int property, default `-1` (unknown/undefined), `0` for insecure, positive values ordered by increasing security. Nothing in this PR consumes the security level yet.

**Streaming API**

- `IStreaming::getProtocolGroupId` and a new trailing `protocolGroupId` parameter on the `StreamingImpl` constructor (defaults to `nullptr`, stored as `""`).

**Streaming source selection (`StreamingSourceManager::attachStreamingsToDevice`)**

- For each protocol group present on the device, computes the best (lowest) priority in `prioritizedProtocolsMap` and skips every capability in that group that is not the best one; also skips a group that already has an attached — or pending, within the same call — streaming source.

**Multiple discovery services per server**

- New `virtual ServerImpl::getDiscoveryConfigs()`, defaulting to a one-element list wrapping the existing `getDiscoveryConfig()`; `enableDiscovery` now registers every returned config against every discovery server.
- `MDNSDiscoveryServer::services` becomes a `std::multimap`, keyed by server id; `registerService` rejects a duplicate only when the same `serviceName` is already registered under that id, and `unregisterService` sends goodbye for and erases the whole range for an id.

**Mocks and tests**

- `MockStreaming::StrictWithGroup` and a `MockStreamingWithProtocol` factory taking an explicit protocol ID and group ID.
- New tests: server capability get/set, freeze, serialize/deserialize and clone round-trips for both new fields; `IStreaming::getProtocolGroupId` incl. the null-out-param error; a `RecordingDiscoveryServer` fake driving five `ServerDiscoveryTest` cases over multi- and single-config servers.
- `WebsocketModulesChannelTest.ConnectAndDisconnectBackwardCompatibility` now skips in the secure parameterisation instead of connecting over `daq.wss://`.

# API changes

```diff
  IServerCapability : IPropertyObject
+ ErrCode INTERFACE_FUNC getProtocolGroupId(IString** protocolGroupId)
+ ErrCode INTERFACE_FUNC getProtocolSecurityLevel(IInteger** securityLevel)

  IServerCapabilityConfig : IServerCapability
+ ErrCode INTERFACE_FUNC setProtocolGroupId(IString* protocolGroupId)
+ ErrCode INTERFACE_FUNC setProtocolSecurityLevel(IInteger* securityLevel)

  IStreaming : IBaseObject
+ ErrCode INTERFACE_FUNC getProtocolGroupId(IString** protocolGroupId) const
```

None. Both capability properties have defaults (`""` and `-1`), and `getProtocolGroupId` on an existing streaming implementation returns `""`.

# Required module changes

- Any out-of-tree implementation of `IServerCapability` / `IServerCapabilityConfig` / `IStreaming` that does not derive from `ServerCapabilityConfigImpl` / `StreamingImpl` must implement the new methods.
- Modules that want group-based deduplication of streaming sources must call `setProtocolGroupId` on their server capability and pass `protocolGroupId` to the `StreamingImpl` constructor. Without it, behaviour is unchanged.
- Servers advertising several mDNS services under one server id should override `getDiscoveryConfigs()` instead of `getDiscoveryConfig()`.

### DIFF
--- a/core/opendaq/device/include/opendaq/server_capability.h
+++ b/core/opendaq/device/include/opendaq/server_capability.h
@@ -83,6 +83,13 @@ DECLARE_OPENDAQ_INTERFACE(IServerCapability, IPropertyObject)
     virtual ErrCode INTERFACE_FUNC getProtocolName(IString** protocolName) = 0;
 
     /*!
+     * @brief Gets the id of the protocol group supported by the device. Should not contain spaces or special characters except for '_' and '-'.
+     * Could be empty if the protocol is not part of any group.
+     * @param[out] protocolGropuId The id of the protocol group.
+     */
+    virtual ErrCode INTERFACE_FUNC getProtocolGroupId(IString** protocolGropuId) = 0;
+
+    /*!
      * @brief Gets the id of the protocol supported by the device. Should not contain spaces or special characters except for '_' and '-'.
      * @param[out] protocolId The id of the protocol.
      */
@@ -140,6 +147,16 @@ DECLARE_OPENDAQ_INTERFACE(IServerCapability, IPropertyObject)
      * @param[out] version The protocol version.
      */
     virtual ErrCode INTERFACE_FUNC getProtocolVersion(IString** version) = 0;
+
+    /*!
+     * @brief Gets the security level of the protocol.
+     * @param[out] securityLevel The security level of the protocol.
+     *
+     * Could be '-1' if the protocol does not define any security level or if the security level is not known,
+     * '0' if the protocol is not secure, and positive integer values for different security levels defined by the protocol.
+     * The higher the value, the more secure the protocol is considered to be.
+     */
+    virtual ErrCode INTERFACE_FUNC getProtocolSecurityLevel(IInteger** securityLevel) = 0;
 };
 /*!@}*/
 
--- a/core/opendaq/device/include/opendaq/server_capability.h
+++ b/core/opendaq/device/include/opendaq/server_capability.h
@@ -87,7 +87,7 @@ DECLARE_OPENDAQ_INTERFACE(IServerCapability, IPropertyObject)
      * Could be empty if the protocol is not part of any group.
      * @param[out] protocolGropuId The id of the protocol group.
      */
-    virtual ErrCode INTERFACE_FUNC getProtocolGroupId(IString** protocolGropuId) = 0;
+    virtual ErrCode INTERFACE_FUNC getProtocolGroupId(IString** protocolGroupId) = 0;
 
     /*!
      * @brief Gets the id of the protocol supported by the device. Should not contain spaces or special characters except for '_' and '-'.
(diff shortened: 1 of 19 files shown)


core/opendaq/device/include/opendaq/server_capability.h
```diff
@@ -82,6 +82,13 @@ DECLARE_OPENDAQ_INTERFACE(IServerCapability, IPropertyObject)
      */
     virtual ErrCode INTERFACE_FUNC getProtocolName(IString** protocolName) = 0;
 
+    /*!
+     * @brief Gets the id of the protocol group supported by the device. Should not contain spaces or special characters except for '_' and '-'.
+     * Could be empty if the protocol is not part of any group.
+     * @param[out] protocolGropuId The id of the protocol group.
+     */
```
JakaMohorko, line 89: We should probably enforce a certain syntax for this to guarantee that these are unique?

core/opendaq/device/include/opendaq/server_capability.h
```diff
@@ -82,6 +82,13 @@ DECLARE_OPENDAQ_INTERFACE(IServerCapability, IPropertyObject)
...
+    virtual ErrCode INTERFACE_FUNC getProtocolGroupId(IString** protocolGropuId) = 0;
```
JakaMohorko:
```suggestion
    virtual ErrCode INTERFACE_FUNC getProtocolGroupId(IString** protocolGroupId) = 0;
```

JakaMohorko (approved):
Looks good to me. The only thought is that we should standardize how the group IDs are formed. There should be a standard syntax to it that makes it unlikely for there to be ID clashes.

core/opendaq/device/include/opendaq/server_capability.h
viacheslauK: I do not see any changes here

NikolaiShipilov: typo `Gropu` -> `Group`

NikolaiShipilov (approved):

PR #1263, opened 2026-08-05
Fix client root device not being synchronized correctly after reconnect if "RestoreClientConfigOnReconnect" is enabled

ankirn-ds:
# Brief

Current issue is that structure of the client device can be out of sync on reconnect if "RestoreClientConfigOnReconnect" is enabled. 

i.e. If a device have additional subdevices after reconnect this is complety unknown to the client device.

# Description

### Before change

There were 2 execution paths:
1) "RestoreClientConfigOnReconnect" = true: 
serialised state of the local device was applied at reconnect

2) "RestoreClientConfigOnReconnect" = false: 
local device was updated from the one on server

### After change :

Local device is always updated from the one on server.

Only if "RestoreClientConfigOnReconnect" = true, serialized state is stored before client device is updated and setting is applied after that.

### DIFF
--- a/shared/libraries/config_protocol/tests/test_core_events.cpp
+++ b/shared/libraries/config_protocol/tests/test_core_events.cpp
@@ -22,6 +22,7 @@
 #include "config_protocol/config_client_device_impl.h"
 #include <coreobjects/user_factory.h>
 #include <opendaq/mock/mock_streaming_factory.h>
+#include <opendaq/mock/mock_physical_device.h>
 
 using namespace daq;
 using namespace daq::config_protocol;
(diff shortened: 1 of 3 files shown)


shared/libraries/config_protocol/tests/test_core_events.cpp
```diff
@@ -1775,12 +1781,40 @@ TEST_F(ConfigCoreEventTest, ReconnectRestoreClientConfig)
     };
 
     client->reconnect(True);
-    ASSERT_EQ(updateCount, 1);
+    //ASSERT_EQ(updateCount, 2);
```
NikolaiShipilov: Was this check disabled intentionally to avoid test failures  - because smth now does not work as expected, or it is left as the commented-out-line accidentally ?

NikolaiShipilov (commented):
lgtm. just one question was posted to be addressed - regarding changes in the tests

shared/libraries/config_protocol/tests/test_core_events.cpp
ankirn-ds: Sorry accidently commented out since I double check that before and after version. It should be update to 2.  Will change it.

NikolaiShipilov (approved):

PR #1266, opened 2026-08-11
[BUG] Replace modified native client property metadata if changed while updating

JakaMohorko:
# Brief

Native client now:
- Updates `IProperty` metadata that changed on the server during `onUpdateEnded`
- Deserializes properties into their `ConfigClient` variant during `onUpdateEnded`

# Description

On a remote update (config load, reconnect), the native client reconciled properties by name only:
new ones were added, missing ones removed, but existing ones were never touched. A property removed
and re-added on the server with different metadata kept its stale metadata on the client.

Additionally, newly added properties were incorrectly deserialized into `PropertyImpl` instead of
the `ConfigClient` variant, losing the config-client behavior (e.g. server-side listener flags). This fix introduces a small overhead during property deserialization; The equality check is negligible.

# Solution

- The native config client now deserializes properties into `ConfigClientPropertyImpl` when
  applying the `ComponentUpdateEnd` core event.
- It compares each newly deserialized property with the existing one via the new value-based
  `Property::equals` and replaces the existing property when they are not equal.

### DIFF
--- a/core/coreobjects/include/coreobjects/property_impl.h
+++ b/core/coreobjects/include/coreobjects/property_impl.h
@@ -1237,6 +1237,91 @@ public:
         return daqDuplicateCharPtr(stream.str().c_str(), str);
     }
 
+    ErrCode INTERFACE_FUNC equals(IBaseObject* other, Bool* equal) const override
+    {
+        OPENDAQ_PARAM_NOT_NULL(equal);
+
+        *equal = false;
+        if (other == nullptr)
+            return OPENDAQ_SUCCESS;
+
+        const ErrCode errCode = daqTry([&]()
+        {
+            const auto otherProperty = BaseObjectPtr::Borrow(other).asPtrOrNull<IProperty>(true);
+            if (!otherProperty.assigned())
+                return OPENDAQ_SUCCESS;
+
+            if (otherProperty.getObject() == propPtr.getObject())
+            {
+                *equal = true;
+                return OPENDAQ_SUCCESS;
+            }
+
+            const auto otherInternal = otherProperty.asPtrOrNull<IPropertyInternal>(true);
+            if (!otherInternal.assigned())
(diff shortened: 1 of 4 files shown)


denise-opendaq (approved):

core/coreobjects/include/coreobjects/property_impl.h
```diff
@@ -1237,6 +1237,91 @@ class PropertyImpl : public ImplementationOf<IProperty, ISerializable, IProperty
...
+            if (!equalsMetadataField(this->suggestedValues, otherInternal.getSuggestedValuesUnresolved()))
+                return OPENDAQ_SUCCESS;
+
+            const auto thisInternal = propPtr.asPtr<IPropertyInternal>(true);
+            if (thisInternal.getHasOnReadListeners() != otherInternal.getHasOnReadListeners())
+                return OPENDAQ_SUCCESS;
+            if (thisInternal.getHasOnGetSelectionValuesListeners() != otherInternal.getHasOnGetSelectionValuesListeners())
+                return OPENDAQ_SUCCESS;
+            if (thisInternal.getHasOnGetSuggestedValuesListeners() != otherInternal.getHasOnGetSuggestedValuesListeners())
+                return OPENDAQ_SUCCESS;
```
NikolaiShipilov, line 1295: Does it actually make sense to compare listeners' presence per property - and conclude that properties aren't equal if only one of them traced via event listener ?

NikolaiShipilov (approved):

core/coreobjects/include/coreobjects/property_impl.h
JakaMohorko: It's not quite the comparison of listeners; this is a flag for the Native Client. It states whether the server has read listeners - such property getters are forwarded to the server. So technically, these are not equal, as they differ in behaviour.

PR #1268, opened 2026-08-12
Fix IcmpPing leaking its object, socket and reply buffer

dejancrn:
# Brief

Fix `IcmpPing` leaking its object, raw ICMP socket and 64 kB reply buffer on every device discovery, plus the data race and dangling send buffer around it.

# Description

- Stop the ping explicitly in `ModuleManagerImpl::checkNetworkSettings`. The re-armed `async_receive` handler captures `shared_from_this()` and `handleReceive` unconditionally re-arms it, so the ping kept a strong reference to itself and releasing the local `shared_ptr` never destroyed it. Every discovery leaked the object, its reply buffer, an open raw ICMP socket and a work guard on the shared `io_context`, and each leaked instance kept processing every ICMP packet the host received.
- Guard the reply state (`responseAddresses`, `numReplies`) with `mutexReplies`. `waitSendAndReply` returns on a fixed 1 s timeout while the receive loop is still running, so `getReplyAddresses` was copying an `unordered_set` on the calling thread while `handleReceive` inserted into it on an `io_context` thread. Mutating under the lock also closes the lost-wakeup window on `allRepliesReceived`.
- Set up everything the async handlers observe (`numRemotes`, `numSent`, `numReplies`, `sequenceNumber`) in `start()` before the first operation is armed, instead of in `startSend()` after `startReceive()` had already been posted.
- Share the encoded request packet with every send handler. It was a stack `boost::asio::streambuf` in `startSend`, but asio requires the buffer to stay valid until `async_send_to` completes, which can be after `startSend` returns.
- Treat `not_connected` from `shutdown()` as expected, since a raw ICMP socket is never connected and would otherwise log an error on every discovery now that `stop()` is actually called.

# Usage example

No usage change — the fix is internal to device discovery. Existing calls are unaffected:

```cpp
auto instance = Instance();
auto devices = instance.getAvailableDevices();   // no longer leaks a socket per call
```

# API changes

None. `IcmpPing` is a concrete internal helper, not an openDAQ interface — no abstract struct or pure virtual function is touched, so module binary compatibility is unaffected. The header changes are member-level only (`numReplies` becomes `std::atomic`, `mutexReplies` becomes `mutable`), and the class is `final` with a private constructor, always built inside the openDAQ library through `IcmpPing::Create`.

```diff
(no interface changes)
```

None.

# Required module changes

None.

---

## Not addressed here

Two pre-existing behavioural issues in the same code path are left alone, as they are judgement calls rather than defects in this fix:

- `icmp->setMaxHops(1)` caps TTL at 1, so any device more than one hop away can never reply and is classified `Unreachable`.
- A ping timeout maps straight to `AddressReachabilityStatus::Unreachable`. ICMP silence is not evidence of unreachability — firewalls commonly drop echo requests while leaving TCP open — and `setAddressesReachable` skips `Unreachable` addresses when selecting a capability's connection string, so a reachable device can lose its IPv4 connection string.

### DIFF
--- a/changelog/changelog_3.30-3.40.md
+++ b/changelog/changelog_3.30-3.40.md
@@ -39,6 +39,7 @@
 
 ## Bug fixes
 
+- [#1268](https://github.com/openDAQ/openDAQ/pull/1268) Fix IcmpPing leaking its object, socket and reply buffer.
 - [#1263](https://github.com/openDAQ/openDAQ/pull/1263) Fix client root device not being synchronized correctly after reconnect if "RestoreClientConfigOnReconnect" is enabled
 - [#1266](https://github.com/openDAQ/openDAQ/pull/1266) Native client updates `IProperty` metadata that changed on the server during `onUpdateEnded`
 - [#1266](https://github.com/openDAQ/openDAQ/pull/1266) Native client deserializes properties into their `ConfigClient` variant during `onUpdateEnded`
--- a/core/opendaq/modulemanager/include/opendaq/icmp_ping.h
+++ b/core/opendaq/modulemanager/include/opendaq/icmp_ping.h
@@ -15,8 +15,11 @@
  */
 #pragma once
 #include <boost/asio.hpp>
+#include <atomic>
 #include <chrono>
+#include <condition_variable>
 #include <mutex>
+#include <unordered_set>
 #include <opendaq/logger_ptr.h>
 #include <opendaq/logger_component_ptr.h>
 
@@ -72,7 +75,7 @@ private:
     std::atomic<std::size_t> numSent;
     std::mutex mutexRequests;
     std::condition_variable allRequestsSent;
-    std::mutex mutexReplies;
+    mutable std::mutex mutexReplies;
     std::condition_variable allRepliesReceived;
 
     uint16_t identifier;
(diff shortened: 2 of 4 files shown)


dejancrn:
DewesoftX periodically calls getAvailableDevices. After a few hours, memory usage grew, and on some PCs the UI became unresponsive, presumably due to many open sockets. This may also fix the issue we observed with the CAF system, where ping was very slow when DewesoftX was running in the background.

JakaMohorko:
@dejancrn Is this something that should be pushed to 3.40?

dejancrn:
> @dejancrn Is this something that should be pushed to 3.40?

It's a serious issue if someone periodically calls getAvailableDevices. We will for sure include it in our fork based on 3.40 RC, because this is our use case. If an app does not call it often, then I guess it's not so serious.

The final decision is up to you.

JakaMohorko:
> > @dejancrn Is this something that should be pushed to 3.40?
> 
> It's a serious issue if someone periodically calls getAvailableDevices. We will for sure include it in our fork based on 3.40 RC, because this is our use case. If an app does not call it often, then I guess it's not so serious.
> 
> The final decision is up to you.

I think the PR looks good and is a good patch - I'd move it to 3.40. Could it be re-targeted there? And a changelog entry added?

JakaMohorko (approved):
Looks  good, missing a changelog and should be re-targeted to 3.40.

dejancrn:
> Looks good, missing a changelog and should be re-targeted to 3.40.

done

PR #1269, opened 2026-08-12
Install OpenSSL in all CI

viacheslauK:
# Brief

The ws-streaming dependency requires OpenSSL >= 1.1.1 on the host for TLS streaming support and does not fetch it automatically, so every build path now installs it: CI, packaging, regression simulator generation, and the documented build prerequisites.

  # Description

  - `ci.yml`: add `libssl-dev` to the `ubuntu-*-x86_64-*` apt list and `libssl-dev:i386` to the `ubuntu-*-x86-*` one.
  - `ci.yml`: for `windows-*-x86-*` run `vcpkg install openssl:x86-windows`, put its `bin` directory on `PATH`, and set the `OPENSSL_ROOT_DIR` cache variable — the runner image ships only 64-bit OpenSSL.
  - `.github/packages/*.txt`: `libssl-dev` (apt), `libssl-dev:i386` (apt-i386) and `openssl-devel` (yum), which covers the packaging jobs including the manylinux container.
  - `build-sdk` action: new Windows-only step that installs 32-bit OpenSSL via vcpkg and exports `OPENSSL_ROOT_DIR` / `PATH` when `enable-32bit` is set.
  - `regression_simulator_generator.yml`: add `libssl-dev` to the apt install list.
  - `quick_start_building_opendaq.adoc`: list OpenSSL (>= 1.1.1) among the Linux prerequisites, explain that it is not fetched automatically, and add `libssl-dev` to the apt-get command.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,11 @@ jobs:
           },
           {
             "match-jobs": ["windows-*-x86-*"],
-            "cache-variables": {"OPENSSL_ROOT_DIR": "C:/Program Files (x86)/OpenSSL-Win32"}
+            "cache-variables": {"OPENSSL_ROOT_DIR": "C:/vcpkg/installed/x86-windows"}
+          },
+          {
+            "match-jobs": ["windows-*-x86_64-*"],
+            "cache-variables": {"OPENSSL_ROOT_DIR": "C:/vcpkg/installed/x64-windows"}
           },
           {
             "match-jobs": ["macos-*"],
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,11 +45,15 @@ jobs:
         [
           {
             "match-jobs": ["ubuntu-*-x86_64-*"],
-            "apt-install": ["lld", "libx11-dev", "libxi-dev", "libxcursor-dev", "libxrandr-dev", "libgl-dev", "libudev-dev", "libfreetype6-dev"]
+            "apt-install": ["lld", "libssl-dev", "libx11-dev", "libxi-dev", "libxcursor-dev", "libxrandr-dev", "libgl-dev", "libudev-dev", "libfreetype6-dev"]
           },
           {
             "match-jobs": ["ubuntu-*-x86-*"],
-            "apt-install": ["lld", "libx11-dev:i386", "libxi-dev:i386", "libxcursor-dev:i386", "libxrandr-dev:i386", "libgl-dev:i386", "libudev-dev:i386", "libfreetype6-dev:i386"]
+            "apt-install": ["lld", "libssl-dev:i386", "libx11-dev:i386", "libxi-dev:i386", "libxcursor-dev:i386", "libxrandr-dev:i386", "libgl-dev:i386", "libudev-dev:i386", "libfreetype6-dev:i386"]
+          },
+          {
+            "match-jobs": ["windows-*-x86-*"],
+            "run": "vcpkg install openssl:x86-windows; Add-Content -Path $env:GITHUB_PATH -Value C:\\vcpkg\\installed\\x86-windows\\bin -Encoding utf8"
           },
           {
             "match-jobs": ["windows-*", "*-x86_64-*", "macos-*"],
@@ -71,6 +75,10 @@ jobs:
           {
             "match-jobs": ["windows-*-msvs-*"],
             "cache-variables": {"OPENDAQ_MSVC_SINGLE_PROCESS_BUILD": true}
+          },
+          {
+            "match-jobs": ["windows-*-x86-*"],
+            "cache-variables": {"OPENSSL_ROOT_DIR": "C:/vcpkg/installed/x86-windows"}
           },
           {
             "match-jobs": ["macos-*"],
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,10 @@ jobs:
             "run": "vcpkg install openssl:x86-windows; Add-Content -Path $env:GITHUB_PATH -Value C:\\vcpkg\\installed\\x86-windows\\bin -Encoding utf8"
           },
           {
+            "match-jobs": ["windows-*-x86_64-ninja-clang-*"],
+            "run": "vcpkg install openssl:x64-windows; Add-Content -Path $env:GITHUB_PATH -Value C:\\vcpkg\\installed\\x64-windows\\bin -Encoding utf8"
+          },
+          {
             "match-jobs": ["windows-*", "*-x86_64-*", "macos-*"],
             "pip-install": ["numpy"],
             "use-python-version": "3.14"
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -45,11 +45,19 @@ jobs:
         [
           {
             "match-jobs": ["ubuntu-*-x86_64-*"],
-            "apt-install": ["lld", "libx11-dev", "libxi-dev", "libxcursor-dev", "libxrandr-dev", "libgl-dev", "libudev-dev", "libfreetype6-dev"]
+            "apt-install": ["lld", "libssl-dev", "libx11-dev", "libxi-dev", "libxcursor-dev", "libxrandr-dev", "libgl-dev", "libudev-dev", "libfreetype6-dev"]
```
denise-opendaq, line 48: can we just in lt add cmake which will fetch libressl. so the module itself will have all necessary deps?

.github/workflows/ci.yml
```diff
@@ -45,11 +45,19 @@ jobs:
...
+            "apt-install": ["lld", "libssl-dev:i386", "libx11-dev:i386", "libxi-dev:i386", "libxcursor-dev:i386", "libxrandr-dev:i386", "libgl-dev:i386", "libudev-dev:i386", "libfreetype6-dev:i386"]
+          },
+          {
+            "match-jobs": ["windows-*-x86-*"],
+            "run": "vcpkg install openssl:x86-windows; Add-Content -Path $env:GITHUB_PATH -Value C:\\vcpkg\\installed\\x86-windows\\bin -Encoding utf8"
+          },
+          {
+            "match-jobs": ["windows-*-x86_64-ninja-clang-*"],
+            "run": "vcpkg install openssl:x64-windows; Add-Content -Path $env:GITHUB_PATH -Value C:\\vcpkg\\installed\\x64-windows\\bin -Encoding utf8"
           },
```
alexadereyko, line 61: Windows runners are shipped with OpenSSL — nothing needs to be installed for x64. `windows-2022-x86_64-msvs-v143-debug` in this same run found it with no configuration at all:

```
Found OpenSSL 3.6.3 at C:/Program Files/OpenSSL/include
```

That is the same version vcpkg builds from source, and `FindOpenSSL` picks up `C:/Program Files/OpenSSL` on its own regardless of generator.

Only x86 needs anything: install the 32-bit package and point CMake at it via the preset (see the `OPENSSL_ROOT_DIR` block below). vcpkg builds it from source on every run — `Restored 0 package(s)`, `Elapsed time to handle openssl:x86-windows: 7 min` — while the choco package ships a prebuilt Win32 installer and fits the declarative `choco-install` key instead of a raw `run`.

```suggestion
          {
            "match-jobs": ["windows-*-x86-*"],
            "choco-install": ["openssl --x86"]
          },
```

.github/workflows/ci.yml
```diff
@@ -72,6 +80,14 @@ jobs:
...
             "cache-variables": {"OPENDAQ_MSVC_SINGLE_PROCESS_BUILD": true}
           },
+          {
+            "match-jobs": ["windows-*-x86-*"],
+            "cache-variables": {"OPENSSL_ROOT_DIR": "C:/vcpkg/installed/x86-windows"}
+          },
+          {
+            "match-jobs": ["windows-*-x86_64-ninja-clang-*"],
+            "cache-variables": {"OPENSSL_ROOT_DIR": "C:/vcpkg/installed/x64-windows"}
+          },
```
alexadereyko, line 90: This is the "point CMake at it" half — x64 needs nothing, x86 points at the 32-bit install.

```suggestion
          {
            "match-jobs": ["windows-*-x86-*"],
            "cache-variables": {"OPENSSL_ROOT_DIR": "C:/Program Files (x86)/OpenSSL-Win32"}
          },
```

.github/workflows/ci.yml
viacheslauK: It is a good question. I would even say that then it is worth doing this not in this module, but in the ws-streaming library. But I chose this route because it's faster. I figured we'd add the build later. Now's the time to figure it out. @JakaMohorko

JakaMohorko: I would keep it in here. We'll need it for Native in either case. This should, however, be a dependency in the module repo, if it is not yet.

The integration tests will be extracted form the main repository once we finish up the bundle repo creation. At that point, only the module binaries should be consumed; the module should not be rebuilt and the dependency will not be needed.

viacheslauK: The dependency is inside ws-streaming.

.github/workflows/ci.yml
```diff
@@ -45,11 +45,15 @@ jobs:
...
           },
           {
             "match-jobs": ["ubuntu-*-x86-*"],
-            "apt-install": ["lld", "libx11-dev:i386", "libxi-dev:i386", "libxcursor-dev:i386", "libxrandr-dev:i386", "libgl-dev:i386", "libudev-dev:i386", "libfreetype6-dev:i386"]
+            "apt-install": ["lld", "libssl-dev:i386", "libx11-dev:i386", "libxi-dev:i386", "libxcursor-dev:i386", "libxrandr-dev:i386", "libgl-dev:i386", "libudev-dev:i386", "libfreetype6-dev:i386"]
+          },
+          {
+            "match-jobs": ["windows-*-x86-*"],
+            "choco-install": ["openssl --x86"]
           },
```
alexadereyko, line 61: Retracting my earlier suggestions — both were wrong, and this restores vcpkg.

`choco install openssl --x86` fails on the runner with `This package does not support 32 bit architecture`, and dropping the x64 entry makes `windows-2022-x86_64-ninja-clang-release` fail at configure with `OpenSSL >= 1.1.1 was not found`. msvs and intel-cc do find `C:/Program Files/OpenSSL`, clang does not — so the image install is not usable everywhere, and the original code was right.

One change on top of restoring it: `windows-*-x86_64-*` instead of `windows-*-x86_64-ninja-clang-*`, so msvs and intel-cc stop silently depending on whatever OpenSSL the runner image happens to ship. The two patterns are disjoint — `-x86_64-` does not match `windows-*-x86-*`.

That costs runtime, though: vcpkg builds OpenSSL from source on every run (`Restored 0 package(s)`, `Elapsed time to handle openssl:x86-windows: 7 min`), so widening the filter adds ~7 min to two more jobs. Worth doing together with a vcpkg binary cache, or keeping the narrow clang-only filter if you would rather not pay it.

```suggestion
          {
            "match-jobs": ["windows-*-x86-*"],
            "run": "vcpkg install openssl:x86-windows; Add-Content -Path $env:GITHUB_PATH -Value C:\\vcpkg\\installed\\x86-windows\\bin -Encoding utf8"
          },
          {
            "match-jobs": ["windows-*-x86_64-*"],
            "run": "vcpkg install openssl:x64-windows; Add-Content -Path $env:GITHUB_PATH -Value C:\\vcpkg\\installed\\x64-windows\\bin -Encoding utf8"
          },
```

.github/workflows/ci.yml
```diff
@@ -72,6 +76,10 @@ jobs:
             "match-jobs": ["windows-*-msvs-*"],
             "cache-variables": {"OPENDAQ_MSVC_SINGLE_PROCESS_BUILD": true}
           },
+          {
+            "match-jobs": ["windows-*-x86-*"],
+            "cache-variables": {"OPENSSL_ROOT_DIR": "C:/Program Files (x86)/OpenSSL-Win32"}
+          },
```
alexadereyko, line 90: Pairs with the above — same two filters, x86 and x64 root dirs.

```suggestion
          {
            "match-jobs": ["windows-*-x86-*"],
            "cache-variables": {"OPENSSL_ROOT_DIR": "C:/vcpkg/installed/x86-windows"}
          },
          {
            "match-jobs": ["windows-*-x86_64-*"],
            "cache-variables": {"OPENSSL_ROOT_DIR": "C:/vcpkg/installed/x64-windows"}
          },
```

alexadereyko (approved):

PR #1270, opened 2026-08-13
fix: Awaitable::hasCompleted no longer requires getResult (TBBAS-3270)

viacheslauK:
# Brief

  `IAwaitable::hasCompleted` reported `false` for work that had already finished until the caller
  consumed the result via `getResult`.

  # Description

  - `AwaitableImpl::hasCompleted` now polls the pending future with
  `future.wait_for(std::chrono::seconds(0)) == std::future_status::ready` instead of `!future.valid()`.
  - Added a file-local `CompletionGuard` RAII helper that sets `completed` when its scope is left,
  including during stack unwinding. A task that finished by throwing still reports itself as completed after  `future.get()` rethrows.
  - Applied the guard to the `is_void_v<TReturn>` branch of `getResult` as well, which previously never
  set `completed` at all. `AwaitableImpl<void>` reported `false` from `hasCompleted` after its result had been taken.
  - Tests.

  # Usage example

  Polling an awaitable now works without consuming its result:

  ```cpp
  auto awaitable = scheduler.scheduleFunction([] { return 1; });

  while (!awaitable.hasCompleted())
      std::this_thread::sleep_for(std::chrono::milliseconds(50));

  Int result = awaitable.getResult();
  ```

# Note

  One behavioural change worth knowing about: because the void branch of `getResult` now sets
  `completed`, a repeated `getResult` on an awaitable from `scheduleGraph` returns `nullptr` with
  `OPENDAQ_SUCCESS` instead of `OPENDAQ_ERR_EMPTY_AWAITABLE`. Repeated `getResult` on an awaitable from
  `scheduleFunction` is unchanged and still surfaces `GeneralErrorException`.

### DIFF
--- a/core/opendaq/scheduler/src/awaitable_impl.cpp
+++ b/core/opendaq/scheduler/src/awaitable_impl.cpp
@@ -7,25 +7,6 @@
 
 BEGIN_NAMESPACE_OPENDAQ
 
-namespace
-{
-    // Sets the flag once the guarded scope is left, including during stack unwinding.
-    struct CompletionGuard
-    {
-        explicit CompletionGuard(std::atomic<bool>& completed)
-            : completed(completed)
-        {
-        }
-
-        ~CompletionGuard()
-        {
-            completed = true;
-        }
-
-        std::atomic<bool>& completed;
-    };
-}
-
 template <typename TReturn>
 AwaitableImpl<TReturn>::AwaitableImpl(Future future)
     : future(std::move(future))
@@ -65,17 +46,12 @@ template <typename TReturn>
 ErrCode AwaitableImpl<TReturn>::getResult(daq::IBaseObject** result)
 {
     OPENDAQ_PARAM_NOT_NULL(result);
-
-    if (!completed && !future.valid())
-    {
-        return DAQ_MAKE_ERROR_INFO(OPENDAQ_ERR_EMPTY_AWAITABLE);
-    }
+    OPENDAQ_RETURN_IF_FAILED(this->wait());
 
     if constexpr (std::is_void_v<TReturn>)
     {
         try
         {
-            CompletionGuard guard(completed);
             future.get();
         }
         catch (...)
@@ -89,10 +65,7 @@ ErrCode AwaitableImpl<TReturn>::getResult(daq::IBaseObject** result)
     else
     {
         std::optional<BaseObjectPtr> optional;
-        OPENDAQ_TRY(
-            CompletionGuard guard(completed);
-            optional = future.get();
-        )
+        OPENDAQ_TRY(optional = future.get();)
 
         if (!optional.has_value())
         {
--- a/core/opendaq/scheduler/src/awaitable_impl.cpp
+++ b/core/opendaq/scheduler/src/awaitable_impl.cpp
@@ -2,7 +2,29 @@
 #include <opendaq/scheduler_errors.h>
 #include <coretypes/objectptr.h>
 
+#include <chrono>
+#include <future>
+
 BEGIN_NAMESPACE_OPENDAQ
+
+namespace
+{
+    // Sets the flag once the guarded scope is left, including during stack unwinding.
+    struct CompletionGuard
+    {
+        explicit CompletionGuard(std::atomic<bool>& completed)
+            : completed(completed)
+        {
+        }
+
+        ~CompletionGuard()
+        {
+            completed = true;
+        }
+
+        std::atomic<bool>& completed;
+    };
+}
 
 template <typename TReturn>
 AwaitableImpl<TReturn>::AwaitableImpl(Future future)
@@ -53,6 +75,7 @@ ErrCode AwaitableImpl<TReturn>::getResult(daq::IBaseObject** result)
     {
         try
         {
+            CompletionGuard guard(completed);
             future.get();
         }
         catch (...)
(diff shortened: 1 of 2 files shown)


JakaMohorko (approved):

core/opendaq/scheduler/src/awaitable_impl.cpp
```diff
@@ -53,21 +75,24 @@ ErrCode AwaitableImpl<TReturn>::getResult(daq::IBaseObject** result)
...
+
         *result = nullptr;
     }
     else
     {
         std::optional<BaseObjectPtr> optional;
-        OPENDAQ_TRY(optional = future.get();)
-
-        completed = true;
+        OPENDAQ_TRY(
```
denise-opendaq: do we need CompletionGuard if we can just set completed= true before return ?

core/opendaq/scheduler/src/awaitable_impl.cpp
denise-opendaq: !future.valid() means that future get was called. how it can happened with !completed

core/opendaq/scheduler/src/awaitable_impl.cpp
viacheslauK: future.get() throws an exception -> stack unwinding -> OPENDAQ_TRY() catches the exception and returns -> completed stays == false;
With guard it will be set to true during stack unwinding.

core/opendaq/scheduler/src/awaitable_impl.cpp
viacheslauK: Default constructed future will return false. This check handles the case when a user create an Awaitable object with an "empty" future object

denise-opendaq: I would probably just refuse creating AwaitableImpl with not valid feature as there is not way to set a an another future. plus   as far as I can see its only called in scheduleGraph 
```cpp
 auto awaitable_ = createWithImplementation<IAwaitable, AwaitableImpl<void>>(executor->run(*flow));
 ```
 
 so AwaitableImpl creates only with a valid future

File: core/opendaq/scheduler/src/awaitable_impl.cpp (hunk 1 above)
Comment (denise-opendaq) [reply core/opendaq/scheduler/src/awaitable_impl.cpp]: ErrCode AwaitableImpl<TReturn>::getResult(daq::IBaseObject** result) can call AwaitableImpl<TReturn>::wait() at the begging (which will set completed = true)

and in the method AwaitableImpl<TReturn>::wait()  we can remove daq_try as the method future.wait(); throws an exception only if its not valid (there is check before already) 

with this you can remove 

```cpp
if (!completed && !future.valid())
    {
        return DAQ_MAKE_ERROR_INFO(OPENDAQ_ERR_EMPTY_AWAITABLE);
    }
```

in AwaitableImpl<TReturn>::getResult(daq::IBaseObject** result)

the only question is left is that AwaitableImpl is not thread safe, but I personally don't have an idea who will use this object in multiple threads

core/opendaq/scheduler/src/awaitable_impl.cpp
viacheslauK: In general, I agree that creating an object without a valid 'future' is pointless. However, I haven't touched that logic. The current implementation handles this scenario correctly: calling 'getResult' will return an error, and calling `hasCompleted` will return an `false` status (probably we should change `false` to `true`, not sure). A fix for this behavior could be to add an exception to the constructor to prevent the creation of such an object, but that change is beyond the scope of the current PR (bug fix).

core/opendaq/scheduler/src/awaitable_impl.cpp
viacheslauK: AwaitableImpl::wait() will set `completed` to `true` only if `future` is valid and `future.wait()` will not throw an exception (https://en.cppreference.com/cpp/thread/future/wait says that it may). So in this case I can't remove 
`if (!completed && !future.valid())
    {
        return DAQ_MAKE_ERROR_INFO(OPENDAQ_ERR_EMPTY_AWAITABLE);
    }`
because of invalid future and I can't remove the guard because wait can throw. So, to use this approach I have to guarantee that the future is always not default constructed. The only way to do it is to throw an exception inside the constructor (but now there is not way to throw an exception from it). And even in this case this class will have the same behavior that it has right now. And it is not the case of this PR. So what is the point?

denise-opendaq: | AwaitableImpl::wait() will set completed to true only if future is valid and future.wait() will not throw an exception (https://en.cppreference.com/cpp/thread/future/wait says that it may). So in this case I can't remove

| The implementations are encouraged to detect the case when valid() == false before the call and throw a [std::future_error](https://en.cppreference.com/cpp/thread/future_error) with an error condition of [std::future_errc::no_state](https://en.cppreference.com/cpp/thread/future_errc).

you are already checking for valid so it wont throw 

about second part 
```cpp 
if (!completed && !future.valid()) { return DAQ_MAKE_ERROR_INFO(OPENDAQ_ERR_EMPTY_AWAITABLE); }
```
can be replaced with wait. so you don't need check anymore (it will happen inside wait. plus you don't need to set complete in getResult because wait() will set it)

in another words 
AwaitableImpl::wait()
{
1. check feature
2. future wait
3. complete true
}

AwaitableImpl::getResult(out)
{
 OPENDAQ_RETURN_IF_FAILED(this->wait())
out = future.get();
}

denise-opendaq (approved):

PR #1273, opened 2026-08-17
Fix native streaming server crash on access to non-existed packet server

NikolaiShipilov:
# Brief

Fixes a SIGABRT in the native streaming server caused by an unchecked `std::unordered_map::at()` in `StreamingManager::processPackets()`. Adds safer lookups via `find()` instead of `at()`.

Fixes wrong packet-streaming optimization parameters order

### DIFF
--- a/shared/libraries/native_streaming_protocol/src/streaming_manager.cpp
+++ b/shared/libraries/native_streaming_protocol/src/streaming_manager.cpp
@@ -47,11 +47,13 @@ void StreamingManager::sendPacketToSubscribers(const std::string& signalStringId
         {
             while (std::next(it) != registeredSignal.subscribedClientsIds.end())
             {
-                sendDaqPacket(sendPacketBufferCb, packetStreamingServers.at(*it), PacketPtr(packet), *it, registeredSignal.numericId);  // copy packet ptr
+                if (auto serverIt = packetStreamingServers.find(*it); serverIt != packetStreamingServers.end())
+                    sendDaqPacket(sendPacketBufferCb, serverIt->second, PacketPtr(packet), *it, registeredSignal.numericId);  // copy packet ptr
                 ++it;
             }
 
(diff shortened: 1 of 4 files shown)


shared/libraries/native_streaming_protocol/src/streaming_manager.cpp
```diff
@@ -47,11 +47,13 @@ void StreamingManager::sendPacketToSubscribers(const std::string& signalStringId
         {
             while (std::next(it) != registeredSignal.subscribedClientsIds.end())
             {
-                sendDaqPacket(sendPacketBufferCb, packetStreamingServers.at(*it), PacketPtr(packet), *it, registeredSignal.numericId);  // copy packet ptr
+                if (auto serverIt = packetStreamingServers.find(*it); serverIt != packetStreamingServers.end())
```
JakaMohorko: We could extract this check into a common, explicitly named function.

shared/libraries/native_streaming_protocol/src/streaming_manager.cpp
```diff
@@ -113,8 +117,9 @@ PacketStreamingServerPtr StreamingManager::getPacketServerIfRegistered(const std
 {
     std::scoped_lock lock(sync);
 
-    if (const auto it = streamingClientsIds.find(clientId); it != streamingClientsIds.end())
-        return packetStreamingServers.at(clientId);
+    if (const auto itStreamingClientsIds = streamingClientsIds.find(clientId); itStreamingClientsIds != streamingClientsIds.end())
```
JakaMohorko, line 119: Similar to the server check. Could be in a separate function.

JakaMohorko (approved):
Minor refactoring comments. Lgtm in general.

Might be worth checking if we could somehow catch this issue earlier - to prevent us from getting into a unsynchronized state where we try to send to disconnected servers/clients. This looks like a race condition. But in general, this fix is adequate for now.

dejancrn (approved):

PR #1277, opened 2026-08-20
Treat missing domain tick resolution as 1/1 in readers

JakaMohorko:
# Brief

Multi reader now handles missing tick resolutions as a 1/1 ratio.

### DIFF
--- a/core/opendaq/reader/include/opendaq/reader_domain_info.h
+++ b/core/opendaq/reader/include/opendaq/reader_domain_info.h
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 #pragma once
+#include <coretypes/ratio_factory.h>
 #include <opendaq/custom_log.h>
 #include <opendaq/logger_component_ptr.h>
 #include <opendaq/reader_utils.h>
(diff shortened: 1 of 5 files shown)


core/opendaq/reader/include/opendaq/reader_domain_info.h
```diff
@@ -71,7 +72,8 @@ struct ReaderDomainInfo
         setMaxResolution(maxResolution);
     }
 
-    RatioPtr resolution{};
+    // a signal without a domain tick resolution is treated as having a resolution of 1/1
+    RatioPtr resolution = Ratio(1, 1);
```
denise-opendaq, line 76: I would add warning if the developer has not called setResolution

denise-opendaq (commented):
looks good. but people should be informed that reader uses default resolution. as it mistake on developer side

core/opendaq/reader/include/opendaq/reader_domain_info.h
JakaMohorko: That's the caveat though - I don't think it's mandatory. I don't know if we have it documented, but a missing tick resolution simply implies we don't scale by it. So such a configuration is actually valid.

denise-opendaq (approved):

PR #1279, opened 2026-08-21
Bindings and Python examples for the TLStreamingModuleModern with a TLS channel

viacheslauK:
# Brief

Expose `ProtocolGroupId` and `ProtocolSecurityLevel` in the C and Python bindings, bump generated-file copyright to 2026, and realign the RTGen source templates with the `bin/Templates` copies they had drifted from. 
The Python integration examples demonstrate how the TLStreamingModuleModern module's TLS channel works.

  # Description

  - Regenerate C and Python bindings for `IServerCapability::getProtocolGroupId`/`getProtocolSecurityLevel`, `IServerCapabilityConfig::setProtocolGroupId`/`setProtocolSecurityLevel`, and  `IStreaming::getProtocolGroupId`.
  - Bump copyright to 2026 in the RTGen header templates, regenerating the 144 Python binding files that carry the header.
  - Sync `RTGen/src/.../Templates` with `RTGen/bin/Templates`, which is what `rtgen.exe` actually loads — a rebuild would otherwise have silently changed generated output. Drops `on $DateAndTime$` from `c.header.template`/`c.template` (removed in `bin/` by #1134), drops the redundant self-assignment guards from `cpp.ptr.template`, and adds `c.method.definition.template` plus its `RTGen.C.csproj` entry.
  - `lt_tls_client.py`,  `lt_tls_server.py` are python examples that demonstrate communication between a client and a server using a TLS channel.
  - `generate_certificates.sh` is a helper to create TLS secrets for testing purposes.

  # Usage example

  ```python
  print(cap.protocol_group_id, cap.protocol_security_level.value)
  ```

  ```c
  daqServerCapability_getProtocolGroupId(capability, &groupId);
  daqServerCapability_getProtocolSecurityLevel(capability, &securityLevel);
  ```

```bash
python3 lt_tls_server.py --cert secrets/server.crt --key secrets/server.key --ca secrets/ca.crt
```

```bash
python lt_tls_client.py --ca secrets/ca.crt --cert secrets/client.crt --key secrets/client.key
```

  # API changes

  No core interface changes — the interfaces were already extended upstream; this only exposes them. 

  None.

  # Required module changes

  None.

### DIFF
--- a/bindings/c/include/copendaq/device/server_capability_config.h
+++ b/bindings/c/include/copendaq/device/server_capability_config.h
@@ -44,6 +44,7 @@ extern "C"
 
     daqErrCode EXPORTED daqServerCapabilityConfig_setConnectionString(daqServerCapabilityConfig* self, daqString* connectionString);
     daqErrCode EXPORTED daqServerCapabilityConfig_addConnectionString(daqServerCapabilityConfig* self, daqString* connectionString);
+    daqErrCode EXPORTED daqServerCapabilityConfig_setProtocolGroupId(daqServerCapabilityConfig* self, daqString* protocolGroupId);
     daqErrCode EXPORTED daqServerCapabilityConfig_setProtocolId(daqServerCapabilityConfig* self, daqString* protocolId);
     daqErrCode EXPORTED daqServerCapabilityConfig_setProtocolName(daqServerCapabilityConfig* self, daqString* protocolName);
     daqErrCode EXPORTED daqServerCapabilityConfig_setProtocolType(daqServerCapabilityConfig* self, daqProtocolType type);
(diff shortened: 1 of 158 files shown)


bindings/c/include/copendaq/device/server_capability_config.h
```diff
@@ -44,6 +44,7 @@ extern "C"
 
     daqErrCode EXPORTED daqServerCapabilityConfig_setConnectionString(daqServerCapabilityConfig* self, daqString* connectionString);
     daqErrCode EXPORTED daqServerCapabilityConfig_addConnectionString(daqServerCapabilityConfig* self, daqString* connectionString);
+    daqErrCode EXPORTED daqServerCapabilityConfig_setProtocolGroupId(daqServerCapabilityConfig* self, daqString* protocolGroupId);
```
denise-opendaq, line 47: @JakaMohorko do we need to have a binding for ServerCapabilityConfig?

denise-opendaq (approved):

bindings/c/include/copendaq/device/server_capability_config.h
JakaMohorko: No, but if it's already there, I would keep it up-to-date. Pretty much every config interface should not be in bindings, but we have not yet pruned these.

PR #1280, opened 2026-08-24
Python GUI: Property View uses PropertyType

Maksim3l:
# Brief

Switches the properties view of the Python GUI demo example from `CoreType` to `IProperty::getPropertyType` for deciding how properties are displayed and how they can be edited, makes ratio properties editable, and hands mouse wheel events over the overlay widgets to the tree behind them.

# Description

## Property type

- Replace `value_type`, `item_type` and `selection_values` with `property_type` in the properties view, the attributes and data descriptor views, the metadata dialog and the instance config dialog
- Selection types are now checked by property type: `Selection` for a plain value list, `IndexSelection` and `SparseSelection`
- Read if a method property is a func/proc once, when the button is built, instead of off the property on every invocation
- Added IDict to the exception to keep empty row instead of displaying `daq::IDict`
- Stop folders collapsing when a component is added to or removed from an object
- Pass the property type to the config dialog printer instead of `item_type`

## Ratio properties

- Make ratio properties editable
- Convert `n/d` and `n` to `daq.Ratio` in `value_to_coretype`
- Accept a non-string value in `value_to_coretype`

## Mouse wheel over overlay widgets

- Fix the properties tree not scrolling while the pointer sits on an overlay combo box or a method button
- Add `utils.bind_mousewheel_to`, binding the Windows `<MouseWheel>` delta and the X11 `<Button-4>` and `<Button-5>` on a widget to scroll the tree behind it

# Implementation details

<details>

### Property type

Previously there was no deterministic way to check a property's type. `PropertyType` was added in the new release, so the task was to replace `CoreType` and the rest of the fields with it.

The method button and the selection combo boxes used to read the type off the property every time they were used. `execute` did it twice per press, once to check if it was an `IFunction` or an `IProcedure` and again to see if it returned a result. When the view refreshes, the overlay widgets are destroyed and rebuilt, so the type is now read only at building and carried in a default argument, `_is_function` on the button and `_is_keyed` on the combo box.

A minor error was found during this, `daq::IDict` still showed up in cells despite the daq I types being done away with in a previous update. `List` and `Dict` shared an elif arm that put `str(node.get_property_value(name))` in the value cell, and a dict stringifies to `daq::IDict`. `Dict` has its own arm now and leaves the cell empty.

```diff
- LoadInstanceConfigDialog::_printed_value(value_type, value)
+ LoadInstanceConfigDialog::_printed_value(property_type, value)                                             # removed artifact under checkbox (text printed)
- PropertiesTreeview::_place_method_button.execute(_prop, _iid)
+ PropertiesTreeview::_place_method_button.execute(_prop, _iid, _is_function)                                # added to not need to double read the type
- PropertiesTreeview::_place_selection_combobox.on_change(event, _prop, _labels, _indices, _cb)
+ PropertiesTreeview::_place_selection_combobox.on_change(event, _prop, _labels, _indices, _cb, _is_keyed)    # writes the index or key, not the label, for IndexSelection and SparseSelection
```

Edited, mostly switching out `value_type`, `item_type` and the `CoreType` members for `property_type`:

```diff
  PropertiesTreeview::fill_properties(parent_iid, node, hidden)                    # value cell elif chain, the duplicate Proc and Func arms collapsed into one, Struct and Object likewise, List and Dict split apart
  PropertiesTreeview::_collect_overlay_items(parent_iid)                           # selection_values only read once the type says selection
  PropertiesTreeview::_create_overlay_for_item(iid, prop)
  PropertiesTreeview::_sync_overlays()
  PropertiesTreeview::handle_paste()
  PropertiesTreeview::handle_clear_property_value()
  PropertiesTreeview::show_menu(event)
  PropertiesTreeview::edit_value(event)
  LoadInstanceConfigDialog::display_config_options(parent_node, prop_object)       # Object test added ahead of the cast
  LoadInstanceConfigDialog::edit_value(event)
  DataDescriptorTreeview::fill_tree(key, value, parent='')
  AttributesTreeview::fill_tree(key, value, parent='', display_attributes=False)
  MetadataDialog::update.fill_tree(key, value, parent='')
  utils::update_properties(target, source)
```

### Ratio properties

Ratio properties were not editable. `value_to_coretype` had no `ctRatio` case, so nothing turned entry text into a `daq.Ratio`.

The `ctRatio` case takes `n/d` or `n`, pulls the numbers out with `re.match(r'\s*(-?\d+)\s*(?:/\s*(-?\d+))?')` and defaults the denominator to 1. No match raises `ValueError`.

`update_property` hands `new_value` straight to `set_property_value`, so the conversion sits in `save_simple_value` ahead of that call. `Ratio` also joined `Int`, `Float` and `String` in the `edit_value` arm that opens a plain entry.

The unit symbol strip for bool, int and float runs on `str(value)` now. It called `str.strip(value)` directly before, which raises `TypeError` on anything that is not a string.

Edited:

```diff
  utils::value_to_coretype(value, coretype)             # ctRatio branch, and the strip runs on str(value)
  PropertiesTreeview::save_simple_value(entry, path)    # converts a ratio entry before update_property
  PropertiesTreeview::edit_value(event)                 # Ratio added to the plain entry arm
```

### Mouse wheel over overlay widgets

A widget placed on top of a scrollable one takes the wheel event and the tree behind it never sees it. The bool checkboxes bound their own closure that scrolled the tree and re-synced the overlays, on `<MouseWheel>` only. The combo boxes bound `lambda e: 'break'` on `<MouseWheel>`, `<Button-4>` and `<Button-5>`, which swallowed the wheel and left the tree still. The method button bound nothing.

`bind_mousewheel_to` binds all three sequences on one widget, scrolls the widget passed as `scrollable`, calls `after_scroll` if it was given, and returns `'break'`. `mousewheel_steps` returns -1 for `<Button-4>` and 1 for `<Button-5>`, which carry no delta, and otherwise `-event.delta / 120`, falling back to a single step when that rounds to zero.

The overlays are placed by pixel over tree rows, so both callers pass `_sync_overlays` as `after_scroll` to move them once the view has shifted.

```diff
+ utils::mousewheel_steps(event)                                      # Button-4/5 carry no delta, a small Windows delta still moves one step
+ utils::bind_mousewheel_to(widget, scrollable, after_scroll=None)    # one binding set for all three sequences
- PropertiesTreeview::_place_bool_checkbox._on_mousewheel(e)          # scrolled, Windows sequence only
- LoadInstanceConfigDialog::_place_bool_checkbox._on_mousewheel(e)    # the same closure, second copy
```

Edited, local wheel bindings replaced with `utils.bind_mousewheel_to`:

```diff
  PropertiesTreeview::_make_combobox(iid, values, current_value, editable=False)
  LoadInstanceConfigDialog::_make_combobox(iid, values, current_value, column, editable=False)
  PropertiesTreeview::_place_method_button(iid, prop)                                             # had none before, the wheel over the button did nothing
```

</details>

# Usage example

In terminal use:

```shell
python gui_demo.py
```

### DIFF
--- a/examples/applications/python/GUI Application/gui_demo/components/generic_properties_treeview.py
+++ b/examples/applications/python/GUI Application/gui_demo/components/generic_properties_treeview.py
@@ -616,9 +616,7 @@ class PropertiesTreeview(ttk.Treeview):
             cb.bind('<FocusOut>', _clear_active_if_needed)
             cb.bind('<KeyPress>', lambda e : 'break')
 
-        cb.bind('<MouseWheel>', lambda e: 'break')
-        cb.bind('<Button-4>', lambda e: 'break')
-        cb.bind('<Button-5>', lambda e: 'break')
+        utils.bind_mousewheel_to(cb, self, self._sync_overlays)
         return cb
 
     def _place_bool_checkbox(self, iid, prop):
@@ -639,12 +637,7 @@ class PropertiesTreeview(ttk.Treeview):
             self.refresh()
 
         cb.configure(command=on_change)
-        def _on_mousewheel(e):
-            self.yview_scroll(int(-1 * (e.delta / 120)), 'units')
-            self._sync_overlays()
-            return 'break'
-
-        cb.bind('<MouseWheel>', _on_mousewheel)
+        utils.bind_mousewheel_to(cb, self, self._sync_overlays)
         self._overlay_comboboxes[iid] = cb
 
     def _place_method_button(self, iid, prop):
@@ -692,6 +685,7 @@ class PropertiesTreeview(ttk.Treeview):
                 
         btn = ttk.Button(self, text=prop.name, command=execute)
         btn.place(x=x, y=y, width=width, height=height)
+        utils.bind_mousewheel_to(btn, self, self._sync_overlays)
         self._overlay_comboboxes[iid] = btn
         
     def _tree_indent(self):
--- a/examples/applications/python/GUI Application/gui_demo/components/generic_properties_treeview.py
+++ b/examples/applications/python/GUI Application/gui_demo/components/generic_properties_treeview.py
@@ -163,21 +163,19 @@ class PropertiesTreeview(ttk.Treeview):
                 # This property was marked as hidden
                 continue
 
-            if property_info.selection_values is not None and property_info.item_type != daq.CoreType.ctUndefined:
+            property_type = property_info.property_type
+
+            if property_type in (daq.PropertyType.IndexSelection, daq.PropertyType.SparseSelection):
                 if len(property_info.selection_values) > 0:
                     property_value = printed_value(
                         property_info.item_type, node.get_property_selection_value(property_info.name))
                 else:
                     property_value = 'Selection list is empty'
-            elif property_info.value_type == daq.CoreType.ctProc:
+            elif property_type in (daq.PropertyType.Procedure, daq.PropertyType.Function):
                 property_value = self._last_method_results.get(property_info.name, '')
-            elif property_info.value_type == daq.CoreType.ctFunc:
-                property_value = self._last_method_results.get(property_info.name, '')
-            elif property_info.value_type == daq.CoreType.ctStruct:
+            elif property_type in (daq.PropertyType.Struct, daq.PropertyType.Object):
                 property_value = ''
-            elif property_info.value_type == daq.CoreType.ctObject:
-                property_value = ''
-            elif property_info.value_type in (daq.CoreType.ctList, daq.CoreType.ctDict):
+            elif property_type in (daq.PropertyType.List, daq.PropertyType.Dict):
                 property_value = str(node.get_property_value(property_info.name))
             else:
                 property_value = printed_value(
@@ -198,7 +196,7 @@ class PropertiesTreeview(ttk.Treeview):
                 print(e)
 
             # Insert a treeview entry widget for the property
-            if property_info.value_type in (daq.CoreType.ctFunc, daq.CoreType.ctProc):
+            if property_type in (daq.PropertyType.Procedure, daq.PropertyType.Function):
                 display_name = '          ' + property_info.name
             else:
                 display_name = property_info.name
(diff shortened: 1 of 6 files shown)


examples/applications/python/GUI Application/gui_demo/components/generic_properties_treeview.py
```diff
@@ -492,27 +497,35 @@ def _collect_overlay_items(self, parent_iid):
...
-        if prop.value_type in (daq.CoreType.ctFunc, daq.CoreType.ctProc):
+        property_type = prop.property_type
+        if property_type in (daq.PropertyType.Procedure, daq.PropertyType.Function):
             self._place_method_button(iid, prop)
-        elif prop.value_type == daq.CoreType.ctBool:
+        elif property_type == daq.PropertyType.Bool:
             self._place_bool_checkbox(iid, prop)
-        elif prop.selection_values is not None and len(prop.selection_values) > 0:
+        elif (property_type in (daq.PropertyType.Selection, daq.PropertyType.IndexSelection,
+                                daq.PropertyType.SparseSelection)
```
JakaMohorko, line 523: Nitpick, but this check is often repeated. Could be a helper method.

examples/applications/python/GUI Application/gui_demo/components/generic_properties_treeview.py
```diff
@@ -492,27 +497,35 @@ def _collect_overlay_items(self, parent_iid):
...
+              and len(prop.selection_values) > 0):
             self._place_selection_combobox(iid, prop)
-        elif prop.value_type == daq.CoreType.ctEnumeration:
+        elif property_type == daq.PropertyType.Enumeration:
             self._place_enum_combobox(iid, prop)
-        elif (prop.value_type in (daq.CoreType.ctString, daq.CoreType.ctFloat, daq.CoreType.ctInt)
+        elif (property_type in (daq.PropertyType.Int, daq.PropertyType.Float, daq.PropertyType.String)
```
JakaMohorko, line 528: Same for this one.  This check is repeated a few times.

JakaMohorko (approved):

PR #1283, opened 2026-08-25
Replay the last value of a constant signal without a domain signal

dejancrn:
# Brief

A constant rule signal with no domain signal now hands its current value to an input port on connect, constant values may be vectors, and a reference function block demonstrates both.

# Description

- Replay the last value of a constant signal that has no domain signal, immediately after an input port connects. Changes to such a signal are rare, so a port connecting between changes would otherwise wait indefinitely without knowing the current value.
- Replay the same value to native streaming subscribers after the first. Only the first subscriber creates a reader, so the later ones never saw the value handed over locally on connect.
- Fix constant rule signals with dimensions. The rule calculator was built from the rule and sample type alone, so it treated one sample as one value and left the rest of each buffer uninitialised. It now receives the element count from the descriptor that owns it.
- Add a `ConstantValue` reference function block with a single `Value` property, holding semicolon separated values whose count is the vector size and whose notation decides the sample type. Connecting a signal to its input port makes the output follow that domain instead.
- Render signals that have no domain signal. The renderer discarded them entirely: their context stayed invalid because a domain descriptor was required, and every packet without a domain packet was dropped.

# Usage example

A constant that a consumer can read at any time, without waiting for the next change:

```cpp
const auto descriptor = DataDescriptorBuilder().setSampleType(SampleType::Int64).setRule(ConstantDataRule()).build();
signal.setDescriptor(descriptor);
signal.sendPacket(ConstantDataPacket(descriptor, int64_t{7}));

// Connecting at any later point yields the descriptor event and then the value 7.
port.connect(signal);
```

Vectors, with the element count taken from the descriptor's dimensions:

```cpp
const auto descriptor = DataDescriptorBuilder()
                            .setSampleType(SampleType::Int32)
                            .setDimensions(List<IDimension>(Dimension(LinearDimensionRule(1, 0, 3))))
                            .setRule(ConstantDataRule())
                            .build();

signal.sendPacket(ConstantDataPacket(descriptor, std::vector<int32_t>{4, 5, 6}));
```

The reference function block, where the text decides both the vector size and the sample type:

```cpp
auto fb = instance.addFunctionBlock("RefFBModuleConstantValue");

fb.setPropertyValue("Value", "2.5");     // one Float64 element
fb.setPropertyValue("Value", "1;2;3");   // three Int32 elements

// With nothing connected to its input port, a consumer connecting here is handed the current value.
port.connect(fb.getSignals()[0]);

// Connect a domain and the output follows it instead, one constant packet per domain packet.
fb.getInputPorts()[0].connect(device.getSignalsRecursive()[0]);
```

# API changes

Binary compatibility holds. `IDataRuleCalcPrivate` is unchanged; the element count reaches the calculator from the descriptor that owns it. The additions below are header-only factories and internal types.

```diff
+ DataPacketPtr ConstantDataPacket(const DataDescriptorPtr& descriptor, T value)
+ DataPacketPtr ConstantDataPacket(const DataDescriptorPtr& descriptor, const std::vector<T>& values)
+ DataPacketPtr ConstantDataPacketWithDomain(const DataPacketPtr& domainPacket, const DataDescriptorPtr& descriptor, uint64_t sampleCount, const std::vector<T>& values)
+ DataPacketPtr ConstantDataPacketWithRawValue(const DataDescriptorPtr& descriptor, void* rawValue)
+ static DataRuleCalc* createDataRuleCalcTyped(const DataRulePtr& outputRule, SampleType outputType, SizeT elementCount = 1)
+ DataRuleCalcTyped<T>::DataRuleCalcTyped(const DataRulePtr& rule, SizeT elementCount = 1)
```

None.

# Required module changes

None.

---

Two related defects were found while doing this and are deliberately **not** addressed here:

- A linear rule combined with dimensions has the same defect the constant rule had, writing one value per sample into a buffer sized for a whole vector. Rejecting the combination in `validate()` was implemented and withdrawn, because `DataDescriptorTest.DeserializeBackwardsCompat` deserializes a descriptor written by an older version that pairs three dimensions with a linear rule. That makes it a data compatibility decision rather than an API one.
- `OrdinalObjectImpl::compareTo` returns `OPENDAQ_EQUAL` when neither operand is greater or lower, which is every NaN comparison, so `ObjectPtr::operator==` reports NaN as equal to everything. The property system then treats a NaN write as a no-op: no listener runs, the value is unchanged, and `setPropertyValue` still reports success.

### DIFF
--- a/core/opendaq/signal/include/opendaq/signal_impl.h
+++ b/core/opendaq/signal/include/opendaq/signal_impl.h
@@ -166,6 +166,8 @@ private:
     void disconnectInputPort(const ConnectionPtr& connection);
     void clearConnections(std::vector<ConnectionPtr>& connections);
     void setKeepLastPacket();
+    bool isConstantWithoutDomain();
+    DataPacketPtr createLastValuePacket();
     TypePtr addToTypeManagerRecursively(const TypeManagerPtr& typeManager,
                                         const DataDescriptorPtr& descriptor) const;
     void buildTempConnections(TempConnections& tempConnections);
@@ -409,6 +411,7 @@ ErrCode SignalBase<TInterface, Interfaces...>::setDescriptor(IDataDescriptor* de
         auto lock = this->getRecursiveConfigLock2();
 
         dataDescriptor = descriptorPtr;
+        setKeepLastPacket();
         const auto packet = DataDescriptorChangedEventPacket(descriptorToEventPacketParam(dataDescriptor), nullptr);
 
         // Should this return a failure error code or execute all sendPacket calls and return one of the errors?
@@ -507,6 +510,8 @@ ErrCode SignalBase<TInterface, Interfaces...>::setDomainSignal(ISignal* signal)
 
         if (domainSignal.assigned())
             domainSignal.asPtr<ISignalEvents>().domainSignalReferenceSet(this->template borrowPtr<SignalPtr>());
+
+        setKeepLastPacket();
     }
 
     if (!this->coreEventMuted && this->coreEvent.assigned())
@@ -887,6 +892,9 @@ ErrCode SignalBase<TInterface, Interfaces...>::listenerConnectedInternal(IConnec
     
     const auto packet = createDataDescriptorChangedEventPacket();
 
+    // A constant signal without a domain signal changes rarely, so hand the new listener the current value.
+    const auto lastValuePacket = createLastValuePacket();
+
     if (connections.empty())
     {
         const ErrCode errCode = wrapHandler(this, &Self::onListenedStatusChanged, true);
@@ -896,9 +904,17 @@ ErrCode SignalBase<TInterface, Interfaces...>::listenerConnectedInternal(IConnec
     connections.push_back(connectionPtr);
 
     if (!schedule)
+    {
         connectionPtr.enqueueOnThisThread(packet);
+        if (lastValuePacket.assigned())
+            connectionPtr.enqueueOnThisThread(lastValuePacket);
+    }
     else
+    {
         connectionPtr.enqueueWithScheduler(packet);
+        if (lastValuePacket.assigned())
+            connectionPtr.enqueueWithScheduler(lastValuePacket);
+    }
 
     return OPENDAQ_SUCCESS;
 }
@@ -1074,6 +1090,7 @@ ErrCode SignalBase<TInterface, Interfaces...>::clearDomainSignalWithoutNotificat
     auto lock = this->getRecursiveConfigLock2();
 
     domainSignal = nullptr;
+    setKeepLastPacket();
 
     return OPENDAQ_SUCCESS;
 }
@@ -1266,9 +1283,40 @@ ErrCode SignalBase<TInterface, Interfaces...>::enableKeepLastValue(Bool enabled)
 }
 
 template <typename TInterface, typename... Interfaces>
+bool SignalBase<TInterface, Interfaces...>::isConstantWithoutDomain()
+{
+    if (onGetDomainSignal().assigned())
+        return false;
+
+    const auto descriptor = onGetDescriptor();
+    if (!descriptor.assigned())
+        return false;
+
+    const auto rule = descriptor.getRule();
+    return rule.assigned() && rule.getType() == DataRuleType::Constant;
+}
+
+template <typename TInterface, typename... Interfaces>
+DataPacketPtr SignalBase<TInterface, Interfaces...>::createLastValuePacket()
+{
+    if (!this->active || !isConstantWithoutDomain())
+        return nullptr;
+
+    if (!lastValueCache.valueDescriptorCached())
+        return nullptr;
+
+    // A value captured under an older descriptor would contradict the descriptor announced on connect.
+    const auto& cachedDescriptor = lastValueCache.getValueDataDescriptor();
+    if (!BaseObjectPtr::Equals(cachedDescriptor, onGetDescriptor()))
+        return nullptr;
+
+    return ConstantDataPacketWithRawValue(cachedDescriptor, lastValueCache.getRawValueData());
+}
+
+template <typename TInterface, typename... Interfaces>
 void SignalBase<TInterface, Interfaces...>::setKeepLastPacket()
 {
-    keepLastPacket = keepLastValue && isPublic;
+    keepLastPacket = keepLastValue && (isPublic || isConstantWithoutDomain());
 
     if (!keepLastPacket)
     {
--- a/examples/modules/ref_fb_module/modules/ref_fb_module/src/constant_value_fb_impl.cpp
+++ b/examples/modules/ref_fb_module/modules/ref_fb_module/src/constant_value_fb_impl.cpp
@@ -1,0 +1,350 @@
+/*
+ * Copyright 2022-2025 openDAQ d.o.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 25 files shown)


core/opendaq/signal/include/opendaq/signal_impl.h
```diff
@@ -887,6 +892,9 @@ ErrCode SignalBase<TInterface, Interfaces...>::listenerConnectedInternal(IConnec
     
     const auto packet = createDataDescriptorChangedEventPacket();
 
+    // A constant signal without a domain signal changes rarely, so hand the new listener the current value.
+    const auto lastValuePacket = createLastValuePacket();
```
tomaz-cvetko: I think we should make this more explicit in terms of when we want to send the last packet on connect. Call the `createLastValuePacket` here with conditions `!this->active || !isConstantWithoutDomain()` to make it explicit that this should happen only under certain conditions (active and constant rule). The failure of `createLastValuePacket` to produce a valid packet should be tied to the creation not being possible (i. e. cache empty or outdated descriptor).

I know the result is the same, but I think it is important to convey "when listener connects, send the last packet IF it is constant rule signal" (and not unconditionally with failure).

core/opendaq/signal/include/opendaq/signal_impl.h
```diff
@@ -1265,10 +1282,41 @@ ErrCode SignalBase<TInterface, Interfaces...>::enableKeepLastValue(Bool enabled)
...
+        return false;
+
+    const auto rule = descriptor.getRule();
+    return rule.assigned() && rule.getType() == DataRuleType::Constant;
+}
+
+template <typename TInterface, typename... Interfaces>
+DataPacketPtr SignalBase<TInterface, Interfaces...>::createLastValuePacket()
+{
+    if (!this->active || !isConstantWithoutDomain())
```
tomaz-cvetko: Is `isConstantWithoutDomain` the best condition here? I imagine someone might implement the same constant value with infrequent updates with either `no domain signal` or `explicit domain signal`. I think both are reasonable choices.

Both have the same problem on the receiving side after connecting. Therefore I think it would be sensible to replay the last packet to the async constant rule signals as well. Otherwise, we are forcing someone to switch to no domain if they want information immediately after connecting.

examples/modules/ref_fb_module/modules/ref_fb_module/src/constant_value_fb_impl.cpp
```diff
@@ -0,0 +1,350 @@
...
+void ConstantValueFbImpl::onDisconnected(const InputPortPtr& port)
+{
+    auto lock = this->getAcquisitionLock();
+
+    domainConnected = false;
+    domainDescriptor = nullptr;
+
+    // Clearing the domain descriptor first tells consumers the domain is gone while they still reference it.
+    outputDomainSignal.setDescriptor(nullptr);
+    outputSignal.setDomainSignal(nullptr);
```
tomaz-cvetko, line 285: `signal.setDomainSignal` doesn't trigger a `DATA_DESCRIPTOR_CHANGED` event by itself, so changing the domain signal is not something the reader/receiver can observe directly by inspecting the data queue. In this case, information is sent by clearing descriptor on the domain signal. Is that a canonical way to signal the loss of domain signal?

examples/modules/ref_fb_module/modules/ref_fb_module/src/constant_value_fb_impl.cpp
```diff
@@ -0,0 +1,350 @@
...
+        packet = connection.dequeue();
+    }
+}
+
+void ConstantValueFbImpl::processEventPacket(const EventPacketPtr& packet)
+{
+    if (packet.getEventId() != event_packet_id::DATA_DESCRIPTOR_CHANGED)
+        return;
+
+    const auto [valueChanged, domainChanged, newValueDescriptor, newDomainDescriptor] = parseDataDescriptorEventPacket(packet);
```
tomaz-cvetko, line 324: If the input signal clears its descriptors by sending NullDataDescriptor this will now fold into "no change happened" and the FB will advertise outdated descirptors. When using `parseDataDescriptorEventPacket`, the `valueChanged` and `domainChanged` flags have to be taken into account to handle descriptor clearing as a real change.

examples/modules/ref_fb_module/modules/ref_fb_module/src/constant_value_fb_impl.cpp
```diff
@@ -0,0 +1,350 @@
...
+
+    // A connected value signal describes its domain separately; a domain signal is one itself.
+    const DataDescriptorPtr descriptor = newDomainDescriptor.assigned() ? newDomainDescriptor : newValueDescriptor;
```
tomaz-cvetko: In the case when the input is (value signal with domain signal) and we are "forwarding the domain", we can't just blindly take the descriptor that did change - if input value descriptor changed (but not input domain descriptor), we have to ignore this and return early.

The correct implementation would go along the lines of knowing what we are forwarding then unconditionally setting the descriptor (nullptr may mean descriptor was cleared) `if (forwardedChanged) { outputDomainSignal.setDescriptor(newForwardedDescriptor); }`

tomaz-cvetko (commented):
The changes implement the proposed functionality in the core. There are some questions about the reference constant value FB implementation (handling descriptors and removing domain signal seem a bit suspicious).

However, I think that we should implement the replay for all constant signals that don't have a linear rule domain. I think someone might try to implement such signals with constant value + explicit domain signal and would also expect the same system of replaying the last value to simplify getting the "current value". Does that not make sense?

Another question is about whether removing domain signal is a valid strategy. How is a signal consumer supposed to react to receiving NullDataDescriptor? Is it clear enough that that may mean the loss of the domain signal alltogether?

examples/modules/ref_fb_module/modules/ref_fb_module/src/constant_value_fb_impl.cpp
dejancrn: This is how it works right now. However, I agree that `outputSignal.setDomainSignal(nullptr)` should trigger DATA_DESCRIPTOR_CHANGED, but that fix probably needs its own PR.

examples/modules/ref_fb_module/modules/ref_fb_module/src/constant_value_fb_impl.cpp
dejancrn: I agree, will update

examples/modules/ref_fb_module/modules/ref_fb_module/src/constant_value_fb_impl.cpp
dejancrn: I agree, will check

core/opendaq/signal/include/opendaq/signal_impl.h
dejancrn: I agree, will update

core/opendaq/signal/include/opendaq/signal_impl.h
dejancrn: I deliberately left out explicit constant signals because we did not define how they should behave. Should both the domain and value signals be defined as constant, and therefore both produce the last value on connect? Or only the value signal, but then the number of packets is not the same? What happens if just the domain is constant? If we allow it only when both are constant, what happens when one signal changes its descriptor to non-constant? To me, it seems like a complex topic, so I simply skipped explicit constant signals for now.

For now, we only need domainless constant signals.

dejancrn: Forgot to mention shared domains. It gets even more complicated there.

dejancrn:
> The changes implement the proposed functionality in the core. There are some questions about the reference constant value FB implementation (handling descriptors and removing domain signal seem a bit suspicious).
> 
> However, I think that we should implement the replay for all constant signals that don't have a linear rule domain. I think someone might try to implement such signals with constant value + explicit domain signal and would also expect the same system of replaying the last value to simplify getting the "current value". Does that not make sense?
> 
> Another question is about whether removing domain signal is a valid strategy. How is a signal consumer supposed to react to receiving NullDataDescriptor? Is it clear enough that that may mean the loss of the domain signal alltogether?

I have addressed the descriptor issues in Constant FB. Constant signals with an explicit domain are still open and require careful thought and design, as they may have several consequences.

PR #1284, opened 2026-08-27
fix(gui_demo): skip modules that don't implement available_streaming_types

viacheslauK:
# Brief

Guard the streaming-protocol enumeration in the GUI demo's connection-string dialog so a module without `available_streaming_types` doesn't break the whole dialog.

# Description

- In `AddConfigDialog.from_connection_string`, wrap `_module.available_streaming_types` in `try`/`except RuntimeError` and skip the module when it raises; other modules are collected as before.


Maksim3l (approved):
LGTM

PR #1285, opened 2026-08-27
Fix a renderer casting bug when sample type changes

tomaz-cvetko:
# Brief

Fix the issue where renderer FB displayed garbled points when sample type changed (float32 <-> float64) during operation.

# Description

Renderer used to assume the latest descriptor described all data points being drawn. This is false when descriptors change. Especially for the case where the width of the single sample value changed, the result was a garbled graph (random lines due to incorrect casting).

To fix that, use descriptor of each packet to draw the line by casting to the provided per-packet type.

Similar fix was applied for the domain. Although with domain it is expected to be significantly fewer occurences of this phenomenon (changing width of the domain sample affects the duration from origin that can be covered by the ticks). Additionally, this issue may have arisen only for explicit data rules, otherwise the ticks were computed from a rule in a correct manner.


denise-opendaq (approved):
I have a feeling that we just have to use stream reader per port and don't duplicate this per type read. but it needs bigger rework :(

PR #1286, opened 2026-08-31
Python GUI: List editing update

Maksim3l:
# Brief

Replaces the container property dialog with in place editing of list and dict items in the properties view, and adds add and remove actions for them to the row context menu.

# Description

## List and dict items

- Edit items in place by double click, the value column for values, the name column for dict keys
- Add and remove items from the row context menu, above or below on a list item
- Default a new item from the item type
- Generate a new dict key from the key type: the next integer, `key 2` for strings, the unused boolean
- Offer the editing only for bool, int, float and string, dict keys included
- Open the new item for editing after the refresh
- Refuse duplicate dict keys on rename
- A renamed dict key moves its entry to the end
- Drop the container dialog from the properties view, the instance config dialog still uses it
- Rename the metadata field selector's OK button to Save

# Implementation details

<details>

### Which containers can be edited

The item type says whether a container can be edited: bool, int, float and string have a default value, anything else does not. For a dict the key type has to be one of those as well. An item row has no property of its own, so its path resolves to the parent list or dict.

```diff
+ PropertiesTreeview::_default_value_for(core_type)  # bool, int, float, string; anything else is not editable
+ PropertiesTreeview::_can_edit_container(prop)      # item type, and for a dict the key type, need a default
+ PropertiesTreeview::container_item_owner(path)     # the list or dict a row is an item of
```

### Adding

A new item uses the item type to know what to make, and takes its default value. A dict key comes from the key type: the next integer, `key 2` for strings, the unused boolean. The list is rebuilt as a new `daq.List` and written to the property. The refresh that follows rebuilds the rows, so the added row is looked up again, by index for a list and by key text for a dict, and the editor opens on it.

```diff
+ PropertiesTreeview::handle_container_add(container_path, index)  # index for above and below on a list item
+ PropertiesTreeview::_new_dict_key(container, key_type)           # next integer, key/key 2, or the unused boolean
+ PropertiesTreeview::_rebuilt_list(items)
+ PropertiesTreeview::item_for_path(path)
+ PropertiesTreeview::added_item_id(container_path, added)         # by index for a list, by key text for a dict
+ PropertiesTreeview::edit_added_item(container_path, added)       # selects and scrolls to the new row
+ PropertiesTreeview::edit_resolved_item(container_path, added)    # resolved again: the refresh rebuilt the row
```

### Removing

The row index gives the list item to delete, the row text gives the dict key. The container is then written back to the property.

```diff
+ PropertiesTreeview::handle_container_remove(container_path, item_iid)
```

### Editing

A double click puts an entry over the value cell, or over the name cell for a dict key. The typed text is converted to the item type, or to the key type for a key. A key that is already in the dict is refused: `daq.Dict` has no `hasKey`, so the key is read, and a `RuntimeError` from `__getitem__` means it is free.

```diff
+ PropertiesTreeview::edit_container_item(item_iid, owner_path, edit_key)  # entry over the name or the value cell
+ PropertiesTreeview::edit_container_item.commit(_event)
+ PropertiesTreeview::save_container_item(entry, owner_path, locator, edit_key)
+ PropertiesTreeview::_renamed_key(container, old_key, new_key)            # daq.Dict has no hasKey binding
- PropertiesTreeview::edit_value()
+ PropertiesTreeview::edit_value(event)                                    # identify_column tells a key edit from a value edit
```

```diff
  PropertiesTreeview::show_menu(event)  # add, remove and edit entries for a container and its items
```

</details>

# Usage example

In terminal use:

```shell
python gui_demo.py
```

### DIFF
--- a/examples/applications/python/GUI Application/gui_demo/components/generic_properties_treeview.py
+++ b/examples/applications/python/GUI Application/gui_demo/components/generic_properties_treeview.py
@@ -7,7 +7,6 @@ import opendaq as daq
 from .. import utils
 from ..app_context import AppContext
 from .function_dialog import FunctionDialog
-from .edit_container_property import EditContainerPropertyDialog
 from .metadata_dialog import MetadataDialog
 from .metadata_fields_selector_dialog import MetadataFieldsSelectorDialog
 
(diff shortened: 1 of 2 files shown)


examples/applications/python/GUI Application/gui_demo/components/generic_properties_treeview.py
```diff
@@ -819,6 +990,89 @@ def edit_struct_property(self, selected_item_id, name, parent):
...
+            return
+
+        try:
+            value = prop.value
+
+            if prop.property_type == daq.PropertyType.List:
+                if not 0 <= locator < len(value):
+                    return
+                value[locator] = utils.value_to_coretype(text, prop.item_type)
+                prop.value = value
```
JakaMohorko, line 1044: The only issue I see here is that property writes can be expensive - a write to a device might take a few seconds. Here, to replace a list of items with new ones, we must make at least one write per field. I think this is fine for now. 

We should think about an option to edit the list, cache the changes, and apply them as one property write.

JakaMohorko (approved):

PR #1288, opened 2026-08-31
Python GUI: Added a search to tree view

Maksim3l:
# Brief

Adds a filter box above the component tree of the Python GUI demo example, matching a row's text, its component name, local id and tags.

# Description

## Tree filter

- Add a filter box above the tree, matching row text, name, local id and tags
- Keep the whole subtree of a match and stop searching below it
- Lift matches to the top level, open them, and drop what holds no match
- Hold the hint as gray text inside the box while it is empty, since ttk has no placeholder
- Add a clear cross inside the box, with the entry padded on the right for it
- Clear on Escape
- Hand the focus to the tree on clear
- Stop the filter box taking focus on start
- Refilter only when the query text changed, rebuilding the tree from the instance each time
- Filter the Modules tab from the same query

## Tree

- Stop folders collapsing when a component is added to or removed from an object
- Ignore a right click on the tree while the Modules tab is open

# Implementation details

<details>

### Filtering the tree

The row text is matched first, then the component's name, local id and tags, each read behind a `try`, since a component that has gone away raises rather than answering. Deleting the non matching rows is safe because every query change rebuilds the tree from the instance. A match inside another match is not lifted out of it.

Algorithem:
- Strip and lowercase the query, and test each field as a case insensitive substring
- Test the row's text first, then the component's `name`, its `local_id`, then each entry of `tags.list`, stopping at the first field that holds the query
- Test the row's text alone where `context.nodes` holds no component for the row
- Leave the global id out of the tested fields

`tree_restore_selection` leaves the selection empty for a default folder row (`Sig`, `FB`, `Dev`, `IP`, `IO`, `Srv`).

```diff
+ App::handle_tree_search_focus_in(event)        # clears the hint before typing
+ App::handle_tree_search_focus_out(event)       # puts the hint back when left empty
+ App::handle_tree_search_changed(event)         # rebuilds only when the query text changed
+ App::handle_tree_search_clear(event)           # also the Escape binding and the cross
+ App::tree_item_matches_search(iid, query)      # row text, name, local id, tags
+ App::tree_apply_search_filter()
+ App::tree_apply_search_filter.collect(parent)  # keeps a match's subtree, does not look deeper
```

Other:

```diff
  App::tree_widget_create(parent_frame)         # the search row, its style and the clear cross
  App::tree_update(new_selected_node)           # applies the filter, on the Modules tab too
  App::tree_restore_selection(old_node)         # a default folder row is left unselected, so it is not collapsed
  App::handle_tree_right_button_release(event)  # no context menu on the Modules tab
```

</details>

# Usage example

In terminal use:

```shell
python gui_demo.py
```

### DIFF
--- a/examples/applications/python/GUI Application/gui_demo.py
+++ b/examples/applications/python/GUI Application/gui_demo.py
@@ -582,7 +582,9 @@ class App(tk.Tk):
 
         # if component is alive and in treeview
         if node and self.tree.exists(desired_iid):
-            if desired_iid != current_iid:  # if component is not already selected
+            if self._is_default_folder(desired_iid):
+                self.tree.selection_set('')
+            elif desired_iid != current_iid:  # if component is not already selected
                 self.tree.selection_set(desired_iid)
                 self.tree.focus(desired_iid)
                 self.tree.see(desired_iid)
--- a/examples/applications/python/GUI Application/gui_demo.py
+++ b/examples/applications/python/GUI Application/gui_demo.py
@@ -97,6 +97,7 @@ class App(tk.Tk):
 
         self.context = AppContext(context_params)
         self.event_port = EventPort(self, event_callback=self.on_refresh_event)
+        self.tree_search_query = ''
 
         self.context.ui_scaling_factor = int(args.scale)
         self.context.include_reference_devices = bool(args.demo)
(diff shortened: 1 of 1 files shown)


JakaMohorko (approved):
Lgtm, but the filter algorithm should be described in the PR description - what are we actually matching against? 

In the future, it'd be a nice addition to be able to select what to match (name, tag, local ID, global ID...)

PR #1289, opened 2026-08-31
Python GUI: Revamped way of adding components to tree

Maksim3l:
# Brief

Replaces the top add button row of the Python GUI demo example with a plus and a three dots button on the instance row, rebuilds every tree context menu from one builder, and drops the parent selection tree from the add device and add function block dialogs.

# Description

## Tree row buttons

- Remove the top button row: Add device, Add function block, Add server
- Add a plus and a three dots button on the instance row, laid out from the right edge
- The plus shows what the instance can add, the dots show what a right click looks like on the row
- Buttons on the instance row only, always visible, every other row is served by its context menu
- Colour a button with its row's background, darkened under the pointer, since a `tk.Label` cannot be transparent
- Keep them placed and coloured through scrolling, selection and tree rebuilds
- Stop the button click propagating to the row underneath

## Context menus

- Build every tree menu from one builder, separators between non empty groups
- Order the groups as add, then state, then removal: lock and unlock, begin and end update, clear property values, remove
- Decide what a component accepts in one place: a function block takes nested function blocks, a device also takes devices
- Only the instance takes servers, `IDevice::onAddServer` refuses the rest
- Leave Remove off a channel, which belongs to its device and cannot be removed on its own
- Show no menu at all where the groups come out empty, instead of an empty popup
- Take the component or its global id in the menu commands, not the tree selection at run time

## Add dialogs

- Drop the parent selection tree from the add device and add function block dialogs
- Add to the component the dialog was opened on, and name it in the title
- Fall back to the instance when the add device dialog is opened on a non device
- Narrow both dialogs

# Implementation details

<details>

### Tree row buttons

The three add buttons sat in a frame above the tree and called `add_device_dialog_show`, `add_function_block_dialog_show` and `add_server_dialog_show` with no component, so they always targeted the instance whatever the tree had selected. The buttons now sit on the instance row itself and the menu they open carries the component.

A `tk.Label` cannot be transparent, so each button paints the row's own background. `tree_widget_create` looks up `Treeview` background and its selected variant from the style, and `darken_color` blends both towards black through `winfo_rgb` for the hover state. `tree_row_button_color` picks between the four depending on whether the row is selected and whether the pointer is on the button.

Placement is by pixel out of `bbox(iid)`, so anything that moves the row has to re-place them. `yscroll`, `<Configure>` and `<MouseWheel>` all queue `tree_root_buttons_sync` on `after_idle`, and `tree_update` hides them before the delete and syncs after the rebuild.

`<Leave>` also fires when a popup grabs under the pointer, which left the hover shade stuck on. `tree_row_buttons_sync` re-reads `winfo_containing(*winfo_pointerxy())` against `tree_row_button_info` instead of trusting the event, `tree_row_menu_open` freezes the colour while a menu opened from a button is up, and the sync runs 200 ms after that menu closes.

`tree_root_row_iid` returns nothing on the Modules tab, so the buttons stay hidden there.

```diff
+ App::darken_color(color, factor=0.85)        # hover shade for a button, tk.Label has no alpha
+ App::tree_row_buttons_create(tree)           # the plus and the dots, with their enter, leave and click bindings
+ App::tree_row_button_color(button, iid)      # the row's own colour, selected and hovered included
+ App::tree_row_buttons_place(buttons, iid)    # right to left from the right edge of the row's bbox
+ App::tree_row_buttons_hide(buttons)
+ App::tree_row_buttons_recolor()
+ App::tree_row_buttons_sync()                 # pointer re-read, so a popup grab leaves no hover stuck on
+ App::tree_root_row_iid()                     # the instance row, nothing on the Modules tab
+ App::tree_root_buttons_sync()                # re-places after a scroll, a resize or a rebuild
+ App::tree_root_buttons_hide()
+ App::handle_tree_mousewheel(event)
+ App::handle_tree_row_button_enter(event)
+ App::handle_tree_row_button_leave(event)
+ App::handle_tree_add_button_clicked(event)   # add entries only
+ App::handle_tree_more_button_clicked(event)  # the whole row menu
+ App::tree_root_menu_popup(event, add_only)   # returns break, the row underneath stays unselected
```

Edited:

```diff
  App::__init__(args)                        # the Add device, Add function block and Add server frame above the tree is gone
  App::tree_widget_create(parent_frame)      # style colours, the buttons, and the scroll, configure and select bindings that re-place them
  App::tree_update(new_selected_node=None)   # hides the buttons before the delete, syncs after the rebuild
```

### Context menus

Five builders each returned a `tk.Menu`, and `create_function_block_menu`, `create_device_menu` and `create_server_menu` extended `create_property_object_menu` by appending to what it had already added. Entry order followed that inheritance rather than what the entries do, so a device menu ran Begin update, End update, Clear property values, Lock, Unlock, Add function block, Remove.

`menu_build` takes groups of `(label, icon, command)` and draws a separator between the non-empty ones, so nothing dangles when a group comes out empty. `menu_groups` dispatches on `IFunctionBlock`, `IDevice` and `IServer` and falls back to the update and property groups, in the order add, state, property, removal.

`menu_add_items` holds what a component accepts. It casts to `IDevice` or `IFunctionBlock`, reads `available_function_block_types` behind a `try` for `RuntimeError`, and offers Add server only when the node's global id matches the instance, since `IDevice::onAddServer` refuses the rest. Remove is dropped for an `IChannel` and for the instance.

`handle_tree_right_button_release` used to fall back to `create_property_object_menu(None)`, which always carried three commands, so a row with nothing applicable still opened a menu. Both it and `tree_root_menu_popup` now test `popup.index(tk.END) is None` and return without popping up.

The commands close over the component or its global id at build time. `handle_begin_update`, `handle_end_update`, `handle_lock` and `handle_unlock` take an iid and fall back to the tree selection when it is not given, so the same entries serve the row menu and the buttons. The casts moved from the call site in `handle_tree_right_button_release` into the group builders.

```diff
- App::create_property_object_menu(node)
- App::create_component_menu(node)
- App::create_function_block_menu(node)
- App::create_device_menu(node)
- App::create_server_menu(node)
+ App::menu_build(groups)                           # separator between non-empty groups only
+ App::menu_add_items(node)                         # what the component accepts, in one place
+ App::menu_update_items(node)
+ App::menu_property_items(node)
+ App::menu_function_block_groups(node, add_items)  # no Remove on a channel
+ App::menu_device_groups(node, add_items)          # no Remove on the instance
+ App::menu_server_groups(node)
+ App::menu_groups(node, include_add=True)          # add dropped where a plus button already offers it
+ App::create_node_menu(node, include_add=True)
- App::add_device_dialog_show()
+ App::add_device_dialog_show(component=None)       # adds to the component it was opened on
- App::handle_begin_update()
+ App::handle_begin_update(iid=None)                # the row the menu was opened on, selection as the fallback
- App::handle_end_update()
+ App::handle_end_update(iid=None)
- App::handle_lock()
+ App::handle_lock(iid=None)
- App::handle_unlock()
+ App::handle_unlock(iid=None)
```

Edited:

```diff
  App::handle_tree_right_button_release(event)   # one builder, and nothing opens when the groups come out empty
```

### Add dialogs

Both dialogs carried a parent tree with its own scrollbar in grid column 0, and the target was whatever was selected in it. `AddDeviceDialog.initial_update` filled that tree and selected the instance, `AddFunctionBlockDialog.update_dialog` filled it and selected `parent_component` or the instance, and `update_parent_devices` walked the component tree recursively in both.

The target is now the node the dialog was opened on. `resolve_parent_device` casts it to `IDevice` and falls back to the instance for anything else, `AddFunctionBlockDialog.update_dialog` falls back to the instance before listing types, and both set their title from the target's name. The grid drops to one column, and the widths go from 1000 to 700 and from 1400 to 1000.

```diff
- AddDeviceDialog::select_parent_device(device_id)
- AddDeviceDialog::handle_parent_device_selected(event)
- AddDeviceDialog::update_parent_devices(tree, parent_id, component)
- AddDeviceDialog::update_parent_devices.traverse_devices_recursive(tree, parent_id, component)
+ AddDeviceDialog::resolve_parent_device(node)  # the node it was opened on, the instance for a non-device
- AddFunctionBlockDialog::select_parent_device(device_id)
- AddFunctionBlockDialog::handle_parent_device_selected(event)
- AddFunctionBlockDialog::update_parent_devices(tree, parent_id, component)
- AddFunctionBlockDialog::update_parent_devices.traverse_devices_recursive(tree, parent_id, component)
```

Edited:

```diff
  AddDeviceDialog::__init__(parent, context, node)              # no parent tree, one grid column, title from the target
  AddDeviceDialog::initial_update()                             # fills the child device list and focuses the connection string box
  AddFunctionBlockDialog::__init__(parent, context, selected_component)   # no parent tree, no uniform columns, title from the target
  AddFunctionBlockDialog::update_dialog()                       # falls back to the instance, then lists the types
```

</details>

# Usage example

In terminal use:

```shell
python gui_demo.py
```


JakaMohorko (approved):
Lgtm.

PR #1290, opened 2026-08-31
Python GUI: Added a logger display

Maksim3l:
# Brief

Adds a logs window to the Python GUI demo example, tailing a file sink the instance is built with, and a toolbar button beside refresh that opens it.

# Description

## Logs window

- Add a non modal logs window tailing the instance's log file
- Follow tail, a clear button, and the line count and file path in the status line
- Colour a line by its severity token: error and critical, warning, debug and trace
- Read at most 256 KiB from the end of the file on open
- Keep at most 5000 lines, dropping from the top
- Poll every 500 ms from the last read offset, and stop the poll on close
- Report no file sink, a missing file and a read error in the status line
- Reuse the window: a second click on the button deiconifies and raises the open one

## Instance log file

- Add a `BasicFileLoggerSink` writing `opendaq_gui_demo_<pid>.log` in the temp directory, removed first if it exists
- Leave the log file path unset when the sink cannot be added

## Toolbar

- Add a flat toolbar button, the frame's background darkened on hover and further while held
- Use it for refresh and logs
- Run the command only when the release lands inside the button
- Add `Dialog::show_modeless`, the same startup as `show` without the `grab_set` and `wait_window`

# Implementation details

<details>

### Logs window

The sink is added on the builder, since `add_logger_sink` is bound there and nowhere else, so the file exists before the instance does. It is named after the process id, and the window seeks to the end of it on open rather than reading what is already there.

```diff
+ LogsDialog::__init__(parent, context, **kwargs)  # follow, clear, status line, text view
+ LogsDialog::initial_update()                     # starts one tail window before the end of the file
+ LogsDialog::poll()
+ LogsDialog::read_new_lines()                     # from the kept offset to the end
+ LogsDialog::append_lines(lines)                  # appends and drops from the top past MAX_LINES
+ LogsDialog::tag_for_line(line)                   # colour from the severity token in the line
+ LogsDialog::set_status(text)
+ LogsDialog::handle_clear_clicked()               # clears the view, never the file
+ LogsDialog::close()                              # cancels the pending poll
+ Dialog::show_modeless()                          # show without grab_set and wait_window
+ App::logs_dialog_show()                          # raises the open window instead of a second one
```

### Instance log file and toolbar button

```diff
+ App::toolbar_button_create(parent, icon, command)  # flat label button, background tracks the frame
+ App::toolbar_button_create.on_release(event)       # only fires when the release is inside the button
```

Other:

```diff
  AppContext::__init__(params)  # the file sink and the log file path
  App::__init__(args)           # refresh and logs, both as toolbar buttons
```

</details>

# Usage example

In terminal use:

```shell
python gui_demo.py
```

### DIFF
--- a/examples/applications/python/GUI Application/gui_demo/app_context.py
+++ b/examples/applications/python/GUI Application/gui_demo/app_context.py
@@ -1,5 +1,6 @@
 import os
 import platform
+import tempfile
 
 import opendaq as daq
 
@@ -50,6 +51,15 @@ class AppContext(object):
 
         for protocol in getattr(params, 'discovery_servers', []):
             builder.add_discovery_server(protocol)
+
+        self.log_file_path = os.path.join(
+            tempfile.gettempdir(), 'opendaq_gui_demo_{}.log'.format(os.getpid()))
+        try:
+            if os.path.exists(self.log_file_path):
+                os.remove(self.log_file_path)
+            builder.add_logger_sink(daq.BasicFileLoggerSink(self.log_file_path))
+        except Exception:
+            self.log_file_path = None
 
         self.instance = daq.InstanceFromBuilder(builder)
         self.instance.context.on_core_event + daq.QueuedEventHandler(self.on_core_event)
--- a/examples/applications/python/GUI Application/gui_demo.py
+++ b/examples/applications/python/GUI Application/gui_demo.py
@@ -28,6 +28,7 @@ try:
     from gui_demo.components.add_server_dialog import AddServerDialog
     from gui_demo.components.add_function_block_dialog import AddFunctionBlockDialog
     from gui_demo.components.load_instance_config_dialog import LoadInstanceConfigDialog
+    from gui_demo.components.logs_dialog import LogsDialog
     from gui_demo.app_context import AppContext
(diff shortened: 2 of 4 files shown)


examples/applications/python/GUI Application/gui_demo/app_context.py
```diff
@@ -51,6 +52,15 @@ def __init__(self, params):
...
             builder.add_discovery_server(protocol)
 
+        self.log_file_path = os.path.join(
+            tempfile.gettempdir(), 'opendaq_gui_demo_{}.log'.format(os.getpid()))
+        try:
+            if os.path.exists(self.log_file_path):
+                os.remove(self.log_file_path)
+            builder.add_logger_sink(daq.BasicFileLoggerSink(self.log_file_path))
+        except Exception:
+            self.log_file_path = None
```
JakaMohorko, line 62: Would we potentially want a more rigorous log cleanup procedure? This for now just cleans up in case of duplicate names. These logs are meant to be more temporary, so it'd probably make sense to just delete all matches on bootup.

JakaMohorko (approved):

PR #1299, opened 2026-09-11
Fix scheduler shutdown from one of its own worker threads

dejancrn:
# Brief

Nothing in openDAQ stops the scheduler explicitly, so it is destroyed when the last reference to the context is released — which can happen on one of the scheduler's own worker threads if objects are released while work is still queued. 

The worker then waits for itself to finish and hangs forever. If the process exits instead, it can crash because the thread pool's shared memory is torn down while workers are still using it. This is the cause of the intermittent `test_audio_device_module` failures on CI.

# Changes:

- `SchedulerImpl::stop()` now detects when it runs on one of its own workers and hands the executor's destruction off to a separate thread instead of blocking on it.
- The wav writer test now stops the scheduler before teardown, so the process no longer exits with work in flight.
- Added a regression test for releasing the last scheduler reference from a worker.

### DIFF
--- a/core/opendaq/scheduler/src/scheduler_impl.cpp
+++ b/core/opendaq/scheduler/src/scheduler_impl.cpp
@@ -10,6 +10,7 @@
 #include <opendaq/work_factory.h>
 #include <coretypes/function_ptr.h>
 #include <coretypes/validation.h>
+#include <thread>
 #include <utility>
 #include <opendaq/thread_name.h>
 
@@ -206,7 +207,16 @@ ErrCode SchedulerImpl::stop()
     stopped = true;
 
     LOGP_T("Stopping scheduler")
-    executor.reset();
+    if (executor && executor->this_worker_id() != -1)
+    {
+        // Destroyed by one of its own workers (a scheduled callback released the last reference to
+        // the context): the executor cannot join the calling thread, so let another thread do it.
+        std::thread([executor = std::move(executor)]() mutable { executor.reset(); }).detach();
+    }
+    else
+    {
+        executor.reset();
+    }
 
     LOGP_T("Stopping main thread worker")
     mainThreadWorker.reset();
--- a/core/opendaq/scheduler/src/scheduler_impl.cpp
+++ b/core/opendaq/scheduler/src/scheduler_impl.cpp
@@ -209,8 +209,7 @@ ErrCode SchedulerImpl::stop()
     LOGP_T("Stopping scheduler")
     if (executor && executor->this_worker_id() != -1)
     {
-        // Destroyed by one of its own workers (a scheduled callback released the last reference to
-        // the context): the executor cannot join the calling thread, so let another thread do it.
+        // Running on one of the executor's own workers, which it cannot join: another thread destroys it.
         std::thread([executor = std::move(executor)]() mutable { executor.reset(); }).detach();
     }
     else
--- a/changelog/changelog_3.40-3.50.md
+++ b/changelog/changelog_3.40-3.50.md
@@ -13,6 +13,7 @@
 
 ## Bug fixes
 
+- [#1299](https://github.com/openDAQ/openDAQ/pull/1299) Fix a deadlock when a scheduler worker releases the last reference to the context: the scheduler no longer waits on its own thread while destroying the executor. Such a shutdown could also crash at process exit, with Taskflow's node pool destroyed under live workers, which is what made `test_audio_device_module` fail intermittently on the gcc-7 32-bit CI lane.
 - [#1295](https://github.com/openDAQ/openDAQ/pull/1295) Fix an intermittent deadlock between `setOperationModeRecursive` and a sub-device's acquisition thread. The device tree lock taken while the operation mode changes no longer locks signals and input ports.
 
 ## Misc
--- a/core/opendaq/scheduler/tests/test_scheduler.cpp
+++ b/core/opendaq/scheduler/tests/test_scheduler.cpp
@@ -4,6 +4,7 @@
 #include <chrono>
 #include <thread>
 #include <future>
+#include <memory>
 #include <opendaq/work_factory.h>
 
 #include <opendaq/logger_factory.h>
@@ -195,6 +196,45 @@ TEST_F(SchedulerTestCommon, StartsAndStopsFromWork)
     scheduler.scheduleWorkOnMainLoop(work);
     scheduler.runMainLoop();
     ASSERT_TRUE(called);
+}
+
+// A callback can drop the last reference to the scheduler, which is then destroyed on one of its
+// own workers. That release has to return: an executor waiting for the task that is destroying it
+// never finishes.
+TEST_F(SchedulerTestCommon, LastReferenceReleasedFromWorker)
+{
+    using namespace std::chrono_literals;
+
+    const auto logger = Logger();
+
+    std::promise<void> mainReleased;
+    auto mainReleasedFuture = mainReleased.get_future().share();
+    std::promise<void> workerReleased;
+    auto workerReleasedFuture = workerReleased.get_future();
+
+    // dies with the callback, i.e. once the worker has finished the task
+    auto token = std::make_shared<int>(0);
+    std::weak_ptr<int> callbackAlive = token;
+
+    {
+        auto scheduler = Scheduler(logger, 2);
+        auto work = Work([scheduler, token = std::move(token), mainReleasedFuture, &workerReleased]() mutable
+        {
+            mainReleasedFuture.wait();
+            scheduler.release();  // the last reference: ~SchedulerImpl runs on this worker
+            workerReleased.set_value();
+        });
+        scheduler.scheduleWork(work);
+    }
+    mainReleased.set_value();
+
+    ASSERT_EQ(workerReleasedFuture.wait_for(5s), std::future_status::ready)
+        << "destroying the scheduler from its own worker deadlocked";
+
+    const auto deadline = std::chrono::steady_clock::now() + 5s;
+    while (!callbackAlive.expired() && std::chrono::steady_clock::now() < deadline)
+        std::this_thread::sleep_for(1ms);
+    ASSERT_TRUE(callbackAlive.expired());
 }
 
 TEST_F(SchedulerTestCommon, ExecutesOneTimeWorkWithTimeLoop)
--- a/core/opendaq/scheduler/tests/test_scheduler.cpp
+++ b/core/opendaq/scheduler/tests/test_scheduler.cpp
@@ -198,9 +198,7 @@ TEST_F(SchedulerTestCommon, StartsAndStopsFromWork)
     ASSERT_TRUE(called);
 }
 
-// A callback can drop the last reference to the scheduler, which is then destroyed on one of its
-// own workers. That release has to return: an executor waiting for the task that is destroying it
-// never finishes.
+// Releasing the last reference from a work callback destroys the scheduler on its own worker.
 TEST_F(SchedulerTestCommon, LastReferenceReleasedFromWorker)
 {
     using namespace std::chrono_literals;
@@ -212,7 +210,7 @@ TEST_F(SchedulerTestCommon, LastReferenceReleasedFromWorker)
     std::promise<void> workerReleased;
     auto workerReleasedFuture = workerReleased.get_future();
 
-    // dies with the callback, i.e. once the worker has finished the task
+    // destroyed with the callback, once the worker is done with the task
     auto token = std::make_shared<int>(0);
     std::weak_ptr<int> callbackAlive = token;
 
--- a/examples/modules/audio_device_module/tests/test_fb_wav_writer.cpp
+++ b/examples/modules/audio_device_module/tests/test_fb_wav_writer.cpp
@@ -64,6 +64,12 @@ protected:
         writerFb = module.createFunctionBlock("AudioDeviceModuleWavWriter", nullptr, "writer");
         writerRe = writerFb;
     }
+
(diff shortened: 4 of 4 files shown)


dejancrn:
# Scheduler shutdown: deadlock and exit-time crash

Status: analysis behind PR [#1299](https://github.com/openDAQ/openDAQ/pull/1299), which fixes the
deadlock and the failing test. The general solution in the last section is a proposal, not
implemented.

## Summary

openDAQ has no deterministic scheduler shutdown. `SchedulerImpl` and its `tf::Executor` are
destroyed when the last reference to the `Context` is released, on whichever thread releases it.
Work queued on the scheduler pins the objects it targets while it runs, so when an owner lets go
of its objects with work still in flight, the last reference is released by a scheduler worker.
Two things then go wrong:

1. The worker destroys its own executor. `~tf::Executor` waits for every task to finish, including
   the one running the destructor, and the worker blocks forever with the whole pool leaked.
2. If the process exits before the queue has drained, `exit()` destroys Taskflow's global node pool
   while workers are still recycling nodes into it, and the process dies with `SIGSEGV` in
   `tf::Executor::_tear_down_async` or `free(): invalid pointer`.

The second one is the intermittent failure of `test_audio_device_module` on
`ubuntu-20.04-x86-ninja-gcc-7-release`. Both are reachable from any application that releases its
objects while data is flowing.

## Symptom

`test_audio_device_module` prints `[  PASSED  ] 20 tests.` and then crashes on the way out:

```
Thread 276 "Scheduler16" received signal SIGSEGV, Segmentation fault.
#0  tf::Executor::_tear_down_async(tf::Node*)
#1  tf::Executor::_invoke(tf::Worker&, tf::Node*)
#2  tf::Executor::_spawn(unsigned int)::{lambda ...}
```

Main thread at the same moment:

```
#0  tf::ObjectPool<tf::Node, 65536u>::~ObjectPool()
#1  __cxa_finalize
#2  __do_global_dtors_aux            <- libopendaq-32-3.so static destructors
#3  _dl_fini
#4  exit
#5  __libc_start_main
```

or, on the allocator path, `free(): invalid pointer` from the same place.

## Lifetime model

- `Context` owns the `Scheduler`; the `Scheduler` owns the `tf::Executor` and its worker threads.
- Every component, input port, signal and reader holds a `ContextPtr`.
- `GenericInputPortImpl::notifyPacketEnqueuedScheduler` posts a `Work` per enqueued packet
  (`input_port_impl.h`). The `Work` lambda holds weak references and resolves them while it runs:
  `StreamReader` (the listener) and `InputPort`. The function block's own callback does the same
  for the function block (`thisWeakRef.getRef()`).
- Taskflow 3.5 keeps every task node in a process-global pool: `inline ObjectPool<Node> node_pool;`
  (`taskflow/core/graph.hpp`). It is a static with a destructor that frees all blocks.

Nothing in `Instance`, `Context` or the test utilities calls `IScheduler::stop()`. The executor's
lifetime is therefore the lifetime of the last object that happens to reference the context.

## Failure mode 1: the last reference is released by a worker

Captured with gdb by sleeping 8 s at `exit()` in a run whose queue had drained:

```
#0  __kernel_vsyscall
#3  __pthread_cond_wait
#4  std::condition_variable::wait
#5  daq::SchedulerImpl::stop()                       <- inlined ~tf::Executor / wait_for_all()
#6  daq::SchedulerImpl::~SchedulerImpl()
#8  daq::ContextImpl::~ContextImpl()
#10 daq::ComponentImpl<IInputPortConfig, IInputPortPrivate>::~ComponentImpl()
#11 daq::GenericInputPortImpl<IInputPortConfig>::~GenericInputPortImpl()
#13 daq::StreamReaderImpl::~StreamReaderImpl()
#15 daq::WorkImpl<GenericInputPortImpl::setListener(...)::lambda>::execute()
#17 tf::Executor::named_silent_async<SchedulerImpl::scheduleWork(IWork*)::lambda>
#18 tf::Executor::_invoke
```

The `Work` released its strong reference to the `StreamReader`, which was the last one; the chain
runs through the input port and the context into `SchedulerImpl::stop()`, whose `executor.reset()`
calls `~tf::Executor`. That destructor does `wait_for_all()` first, which waits for
`_num_topologies == 0`. The async node executing this very destructor only decrements the count in
`_tear_down_async`, after `_invoke` returns, so the wait never ends. `_done` is never set, the
other 31 workers idle forever, and the executor, the context and everything below stay allocated
until the process exits. In a test this is invisible; in an application it is a hang at shutdown.

## Failure mode 2: process exit with live workers

Same setup, but the process reaches `exit()` while the queue is still draining. At the
`~ObjectPool` breakpoint during `exit()`:

- 35 threads alive: main, two spdlog threads, and all 32 `Scheduler0`..`Scheduler31`.
- 28 workers blocked on the writer function block's mutex inside
  `WAVWriterFbImpl::processInputData()`, one inside `StreamReaderImpl::readPackets()`, the rest
  finishing tasks in `_tear_down_async` / `~tf::Node`.

ASan (gcc 7, `-m32`, run under `taskset -c 0-3` so the first report is not cut short by the next
faulting thread):

```
heap-use-after-free, READ of size 4, thread Scheduler23
  #0 tf::Executor::_tear_down_async                 executor.hpp:1559   node_pool.recycle(node)
  #1 tf::Executor::_invoke                          executor.hpp:1443

freed by thread T0 (main):
  #0 tf::ObjectPool<tf::Node>::Block::operator delete
  #1 tf::ObjectPool<tf::Node>::~ObjectPool          object_pool.hpp:247
  #2 __cxa_finalize

previously allocated by thread T292 (miniaudio worker):
  tf::ObjectPool<tf::Node>::Block::operator new
  daq::GenericInputPortImpl<IInputPortConfig>::notifyPacketEnqueuedScheduler   input_port_impl.h:384
  daq::ConnectionPtr::enqueue
  daq::SignalConfigPtr::sendPacket
  WAVReaderFbImpl::miniaudioDataCallback
  ma_worker_thread
```

`exit()` runs the static destructors of `libopendaq` while its own worker threads are still
executing tasks; the node pool goes first and the workers touch freed blocks.

## Why it is intermittent

`WavWriterTest.ValidWriteToFile` connects a wav reader to a wav writer whose input port uses
`PacketReadyNotification::Scheduler`. Every packet becomes one work item, and every work item takes
the function block's `sync` mutex in `processInputData()`. With `hardware_concurrency()` workers
(32 on the machine used, 2-4 on a GitHub runner) they convoy on that mutex, and the main thread,
which needs the same mutex in `stopRecording()`, is starved:

```
#0  __lll_lock_wait
#2  __GI___pthread_mutex_lock
#3  daq::MutexImpl::lock()
#5  daq::PropertyObjectCoreImpl::getRecursiveLockGuard
#7  WAVWriterFbImpl::stopRecording()
#9  WavWriterTest_ValidWriteToFile_Test::TestBody()
```

The test normally takes 1.0 s; in roughly one run in three it takes 4.8-17 s because of this
stall, during which the reader keeps producing packets and the queue keeps growing. The fixture is
then destroyed with that backlog still queued, and the next test plus process exit take about 30
ms. Without `MALLOC_PERTURB_` the freed pool memory usually still reads as valid, which is why 206
local runs passed before the crash was reproduced; with it, 3 of 30 runs crash.

## Existing workarounds in the tests

Twelve test sites already stop the scheduler by hand before letting their objects go:

| where | how |
| --- | --- |
| `core/opendaq/reader/tests/reader_common.h`, `test_packet_reader.cpp` | `TearDown()`: `scheduler.stop()` followed by a 10 ms sleep |
| `core/opendaq/signal/tests/test_signal.cpp`, `test_packet_e2e.cpp` | `scheduler.stop()` at the end of the test body |
| `examples/modules/ref_fb_module/.../test_ref_fb_module.cpp`, `test_fb_statistics.cpp`, `test_fb_power_reader.cpp` (6x) | `ctx.getScheduler().stop()` at the end of the test body |

`test_fb_statistics.cpp` carries the comment `// Fix so PacketReadyNotification::Scheduler works`.
The problem has been met before and worked around per test. The other ~30 fixtures that build a
bare `Context(Scheduler(...))` are exposed whenever scheduler-notified work is in flight at
teardown.

## Fix in PR #1299

- `SchedulerImpl::stop()` checks `executor->this_worker_id() != -1`. When the call comes from one
  of the executor's own workers, the executor is moved into a detached `std::thread` that waits for
  the remaining work, joins the workers and destroys it. The calling worker returns, finishes its
  task and is joined like the others. This removes failure mode 1.
- `WavWriterTest::TearDown()` calls `context.getScheduler().stop()`, following the reader tests.
  This removes failure mode 2 for this test by draining and joining on the owning thread before
  any object is released.

Measured on Ubuntu 20.04 / gcc 7.5 / `-m32`, 32 cores, `MALLOC_PERTURB_=165`, 30 runs each:
`main` 3 crashes, `stop()` change only 2, both changes 0. With only the `stop()` change, a stalled
run leaves no `Scheduler*` thread alive 8 s after `exit()`, where before all 32 were still parked.

Limits: the detached thread still races `exit()` for a few milliseconds if the process ends right
after a worker-side release; and the fixture change is per test.

## Proposed general solution

The intent of the refcounted design is that no shutdown code is needed. Failure mode 1 can be met
entirely inside the SDK. Failure mode 2 cannot be met by refcounting alone: an object graph
legitimately outlives `main` while a worker holds it, and C++ offers no safe way to destroy statics
under threads running arbitrary callbacks, so *something* has to join the pool before static
destruction. It can be the SDK rather than the user, in three layers:

1. **Cancel semantics for `stop()`.** Each posted task checks an atomic flag shared with the
   scheduler (`if (!cancelled) work->execute()`); `stop()` sets it, so `wait_for_all()` completes in
   microseconds instead of draining the backlog, then joins. Shutdown becomes "drop the queue", which
   is the right policy at teardown. Keeping the executor alive until the flag is set also closes the
   existing race between `scheduleWork` (reads `stopped`, dereferences `executor`) and a concurrent
   `stop()`.
2. **`Instance` owns its scheduler's lifetime.** `~InstanceImpl` cancels and joins the scheduler
   after `rootDevice.remove()`, when the builder created it. Applications built on `Instance` get
   deterministic teardown with no shutdown code. Schedulers supplied through
   `InstanceBuilder::setContext`/`setScheduler` are left to their owner.
3. **Exit hook, best effort.** An `atexit` handler registered when the first executor is created
   cancels and joins every executor still alive, with a bounded wait. On glibc a handler registered
   during `main` runs before `_dl_fini`, which is where the DSO static destructors (and
   `~ObjectPool`) run. Windows terminates the other threads at `ExitProcess` before any DLL is
   detached, so the hook is mostly moot there. macOS ordering has not been checked. The residual
   hole is statics created *after* the hook was registered: they are destroyed before it runs, and a
   callback still in flight could touch one. Hence "best effort".

With these, the explicit `stop()` calls in the tests become unnecessary, and a bare `Context` that
simply goes out of scope gets: queue dropped, threads joined, clean exit. Work still queued at that
point is not completed; that is a policy choice and has to be stated.

**Alternative that removes the problem by construction:** one process-wide executor, created
lazily, owned by a static and torn down by the exit hook, with `Context` holding a handle instead of
the pool. No destructor can then run on the pool's own worker, and there is exactly one thing to
join at exit. It costs the per-`Instance` worker count and isolation, so it is a semantic change.

Related, independent of the above:

- The convoy is a latency problem of its own. `PacketReadyNotification::SchedulerQueueWasEmpty`
  coalesces to one work item per port and would have kept the backlog at one, making the window
  nearly unreachable without changing the fact that it exists.
- Taskflow's global `node_pool` is what makes even an idle leaked executor unsafe at exit. Never
  destroying that pool would remove this particular crash but not the class.

## Reproduction

Ubuntu 20.04 container, `gcc-7`/`g++-7` with `-m32` (`OPENDAQ_FORCE_COMPILE_32BIT=ON`), Release,
32 cores. Plain runs pass; the crash needs the freed memory to be visibly poisoned:

```sh
MALLOC_PERTURB_=165 ./test_audio_device_module        # ~3 crashes in 30 runs
```

Useful gdb one-liners against the release build:

```sh
# who is alive when the node pool dies
gdb -batch -ex 'set breakpoint pending on' \
    -ex 'break tf::ObjectPool<tf::Node, 65536u>::~ObjectPool' \
    -ex run -ex 'info threads' -ex 'thread apply all bt' --args ./test_audio_device_module

# let the queue drain past exit() and see the worker parked in stop()
gdb -batch -ex 'break exit' -ex run -ex 'call (int)sleep(8)' -ex 'info threads' \
    -ex 'thread apply all bt' --args ./test_audio_device_module --gtest_filter=WavWriterTest.*
```

ASan with gcc 7 needs `-U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=0` (always-inline conflict in gtest) and
`taskset -c 0-3`, otherwise libasan 4 aborts the first report when a second thread faults.

core/opendaq/scheduler/src/scheduler_impl.cpp
```diff
@@ -206,7 +207,15 @@ ErrCode SchedulerImpl::stop()
     stopped = true;
 
     LOGP_T("Stopping scheduler")
-    executor.reset();
+    if (executor && executor->this_worker_id() != -1)
+    {
+        // Running on one of the executor's own workers, which it cannot join: another thread destroys it.
+        std::thread([executor = std::move(executor)]() mutable { executor.reset(); }).detach();
```
denise-opendaq, line 213: ~looks like a bug on the task flow side. maybe this lib already have a fix, so we can just bump the version of it~ Never mind, I just read the brief

denise-opendaq (approved):
overall looks good to me

JakaMohorko (approved):

PR #1303, opened 2026-09-15
Stop the mDNS discovery server before an instance releases its root device

dejancrn:
# Brief

Stop the mDNS discovery server's service thread before an instance releases its root device, so a request the thread is answering cannot outlive the device tree.

# Description

The service thread answers for the root device: the IP configuration callbacks upgrade the weak device reference to a strong one for the duration of a request, and the registered service's device info forwards user name and location reads to the device. The instance destructor released the root device, and through it the context owning the discovery servers, without stopping that thread. With a request in flight the device tree outlived the instance, reported as a leak of several thousand tracked objects in debug runs, and the thread then destroyed the context and the server on itself, whose destructor joined its own thread and terminated the process.

- `InstanceImpl::~InstanceImpl` clears the root device on every discovery server after the servers are removed, before releasing the device.
- `MdnsDiscoveryServerImpl::setRootDevice(nullptr)` stops the service thread before unregistering the IP configuration service; registering a service starts it again. `MDNSDiscoveryServer::stop()` moved to public for this.
- The loop wakes every 100 ms, so the join only moves from the context destructor to the instance destructor.
- `daqPrintTrackedObjects` prints a placeholder for an object whose string conversion throws, so a leak report cannot end a run.

`ModulesDeviceDiscoveryTest.ChangeIpConfigError` leaked once in about 22 runs and aborted the next one; after the fix it passed 60 of 60 repeats, and the discovery tests 10 of 10.

# Usage example

No behaviour change for applications. The service thread is now joined at instance destruction rather than at context destruction:

```cpp
{
    auto instance = InstanceBuilder().addDiscoveryServer("mdns").build();
    instance.addServer("OpenDAQOPCUA", nullptr);
}   // the mDNS service thread is joined here, before the root device goes
```

# API changes

None. `MDNSDiscoveryServer::stop()` is on an internal helper class in `shared/libraries/discovery_server`, not on an openDAQ interface, so module binary compatibility is unaffected.

None.

# Required module changes

None.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


JakaMohorko (approved):
Lgtm, just needs a PR tag edit in the changelog.

PR #1305, opened 2026-09-15
Fix native streaming server abort on shutdown

tomaz-cvetko:
# Brief

Server should first stop and join its threads - there may be tasks pending in those threads that affect the root device. Additionally, exceptions should be caught to prevent and at least log a server shutdown failure.

# Description

During investigations of a flaky test I encountered an abort when the test concluded while the disconnect handler was still running (due to a investigative sleep). The cause of this crash was the fact that stopServerInternal removed and cleared connected clients while the disconnected handler was doing the same - the double removal caused a crash.

To avoid this, server should be stopped and the threads should be joined. Only then should we proceed with cleaning up the state - with threads joined we know we have the ownership over the state.

The change is mainly a change in the order of operations. Now the first step is to stop the server and join the threads. The second one is to update the state to be in line with the stopped server. The try catch block is there to ensure that we get a log line in case something goes wrong (instead of the silent abort) and to ensure that registeredClientIds and disconnectedClientIds are cleared.


modules/native_streaming_server_module/src/native_streaming_server_impl.cpp
```diff
@@ -279,27 +279,39 @@ void NativeStreamingServerImpl::stopServerInternal()
...
+            }
+        }
+        catch (const std::exception& e)
+        {
+            LOG_W("Native server - exception occurred on shutdown: {}", e.what());
+        }
+        catch (...)
+        {
+            LOG_W("Native server - unknown exception on shutdown");
         }
```
NikolaiShipilov, line 311: These worth to be `error` level log `LOG_E` messages

NikolaiShipilov (approved):
lgtm. 
Does this change alone allow to exclude any tests from "flaky" category ?

modules/native_streaming_server_module/src/native_streaming_server_impl.cpp
tomaz-cvetko: Will change.

tomaz-cvetko:
> Does this change alone allow to exclude any tests from "flaky" category ?

The fix for the flaky test was done in a separate PR. This one is more of a "issue waiting to happen". It could cause some unexplained crashes of tests/application and it would be hard to figure out/exactly reproduce. During normal operation the window for the race is small, I accidentally made it bigger (added a sleep in a sensitive place) when looking at the other issue. So I think we should add this because the change itself is minor but it could save us headaches down the line.

I'm not sure if I can (reliably) prove this with a test though.

PR #1307, opened 2026-09-18
Precompiled headers & Unity builds

JakaMohorko:
# Brief
Precompiled headers and unity builds through opendaq-cmake-utils v1.1.0, on by default.

# Overview

Requires opendaq-cmake-utils v1.1.0: https://github.com/openDAQ/opendaq-cmake-utils/pull/7

- PCH for the SDK libraries, bindings, modules, examples and tests.
- Unity builds for the test, binding and selected library targets. Test sources get file-local namespaces so they can share a translation unit.


dejancrn (approved):

NikolaiShipilov (approved):
👍 
Some of the external dependencies (e.g. spdlog, sfml; ) also have their own PCH acceleration options. So, maybe it makes sense to enable such options recursively when `OPENDAQ_ENABLE_PCH` is `ON`.

PR #1309, opened 2026-09-18
Publish native connection status without throwing 

JakaMohorko:
# Brief

No longer throws `NotFound` on transport status change before connection statuses are published.

#  Overview

A transport status callback can reach the native device before its connection statuses exist. The `NotFound` exception escaped the asio handler and terminated the process without a test summary.

The status updates now check return codes and ignore a missing status.


tomaz-cvetko (approved):
Probably needs a changelog entry. Otherwise lgtm.

PR #1310, opened 2026-09-18
Do not rebuild package version on commit + Enable `/MP` flag on CI

JakaMohorko:
# Brief
Provide `OPENDAQ_PACKAGE_VERSION` through a generated header. Use `/MP` for VS CI jobs.

## Overview
Rebuild less after a commit and use all cores on the Visual Studio lanes

- Provide `OPENDAQ_PACKAGE_VERSION` through a generated header instead of a global compile definition, so a new revision rebuilds only the files that include it.
- Let the Visual Studio CI lanes compile with `/MP`.


alexadereyko (approved):
Looks good to me

PR #1311, opened 2026-09-18
Speed up and stabilize the device module tests

JakaMohorko:
# Brief

Add the `ScanOnAdd` option to skip the network scan on `addDevice`; tests wait for state instead of sleeping; declare `RESOURCE_LOCK` for parallel ctest; write a minidump on access violations.

# Overview

- Replace fixed sleeps and packet counts with waits for the expected state.
- Add the `ScanOnAdd` option: system-wide through `ModuleManager.ScanOnAdd`, and per call through the `General` add-device config property, which defaults to the option. When off, `addDevice` does not scan the network; a smart `daq://` connection string then fails unless an earlier scan already listed the device. The device module tests turn it off.
- Declare the ports and mDNS each test binary occupies through `RESOURCE_LOCK`, including fetched module tests, and add the `run_tests_parallel` preset.
- Write a minidump when a test process hits an access violation on Windows, before gtest reports the failure.

> **Note:** `ScanOnAdd` is a temporary home on the current `General` config object. The add-device config rework will move it into the new `IAddDeviceConfig` interface under the same name and semantics.


tests/integration/test_opendaq_device_modules/test_helpers/test_helpers.h
```diff
@@ -346,6 +336,98 @@ namespace test_helpers
...
+    {
+        const auto deadline = std::chrono::steady_clock::now() + timeout;
+        while (!condition())
+        {
+            if (std::chrono::steady_clock::now() >= deadline)
+                return false;
+            std::this_thread::sleep_for(std::chrono::milliseconds(20));
+        }
+        return true;
+    }
```
tomaz-cvetko: I introduced the same concept (slightly different signature) in my flakiness fix. It is located higher up in this file. I think we should just keep one of those. The sleeping period is probably not that important/variable so I suggest that we delete the template <typename Predicate> waitFor and keep this one.

tomaz-cvetko (approved):
It (especially ScanOnAdd) looks good to me.

PR #1315, opened 2026-09-21
Fix MacOS packaging + install missing header

JakaMohorko:
# Brief

Install the crash dump header and resolve the macOS package resources from the openDAQ tree.


NikolaiShipilov (approved):

alexadereyko (approved):
Looks good to me

